PR #14, opened 2021-04-19
fix scikit-image version skipping in the test suite

grlee77:
Due to a typo in test skip conditions (using `skimage>=1.18`) rather than the intended case of `scikit-image>=0.18` a handful of test-cases were always being skipped even when new enough scikit-image was available.

Fixing the test skip condition so that these cases get run revealed a couple of errors in the test cases themselves, but fortunately did not reveal any underlying errors in the functions being tested.


quasiben (approved):

quasiben:
@gpucibot merge

Issue #24, opened 2021-05-09
[DOC] document how to do an editable install (for Python developers)


grlee77:
## Report needed documentation

**Report needed documentation**

It would be useful to expand the contributor documentation to explain how the python package can be installed as an "editable install". 

When I tried a naive `pip install -e . -v` call from within the `python/cucim` folder where `setup.py` is contained, this runs fine, but I am unable to import the package. I think the issue is that the `cucim` module folder is within a `src` subfolder rather than directly in the folder containing `setup.py`.

**Describe the documentation you'd like**
A workaround that fixes it for me, it to add the following to the `setup.cfg`:

```
[egg_info]
egg_base = src
```
which causes the `cucim.egg-info` folder to get placed into the `src` subfolder rather than in the same folder as `setup.py` itself.

**Steps taken to search for needed documentation**
read `CONTRIBUTING.md`

PR #26, opened 2021-05-09
add egg_info section to setup.py for editable installs

grlee77:
See #24

This PR adds an `egg_info` section to our `setup.cfg` so that editable installs via `pip install -e . ` work.


quasiben:
@gpucibot merge

PR #29, opened 2021-05-12
MAINT: avoid use of deprecated cp.float, cp.int, cp.bool

grlee77:
This PR silences a number of warnings when running the tests with the current CuPy development branch. As in NumPy, these types are deprecated and `float`, `int` or `bool` should be used instead.


grlee77:
not sure if we should create a new "maintenance" label? I marked it as "bug" for now, but it is really preventing a future bug if these dtypes actually get removed.

jakirkham:
Yeah the labels are configured by RAPIDS OPS, but we can ask

jakirkham (approved):

jakirkham:
@gpucibot merge

PR #46, opened 2021-06-02
Fix-up install issues encountered on CI part 2

quasiben:
A fresh attempt at fixing up CI (xref https://github.com/rapidsai/cucim/pull/45)

PR #45, opened 2021-05-27
Fix-up install issues encountered on CI

jakirkham:
~~Includes PR ( https://github.com/rapidsai/cucim/pull/41 ) and PR ( https://github.com/rapidsai/cucim/pull/43 )~~ Dropped as these are now in `branch-21.06`

Also makes some fixes to hopefully address the install issues we are seeing on CI.

Issue #44, opened 2021-05-23
cuda 10.x support


chrisroat:
Is cuda 10.x support feasible or planned?

quasiben:
Unfortunately, no.    cuCIM depends on 11.X features -- @gigony is the best person to explain where those features are used.

gigony:
Hi @chrisroat,

To support GDS feature in cuCIM, CUDA version >= 11.x is needed (though we currently disabled GDS feature for now and would enable it again in the future) and we currently release conda packages with CUDA 11.x only

However, since cuCIM's IO package (cucim.CuImage/ cucim.clara.xxx) works independently with cuCIM's CuPy-based scikit-image API (cucim.skimage), I think you can install and use only `cucim.skimage` (previously https://github.com/mritools/cupyimg ) package through the following command using pip:

```bash
# Replace cupy-cuda10X with corresponding version name such as `cupy-cuda102` available from https://cupy.dev/
pip install --upgrade scipy scikit-image cupy-cuda10X cucim
```

Please help let us know your use cases with cuCIM!

chrisroat:
Thanks!  The github README.md explicitly indicates cuda 11 is needed.  I have another package stuck on cuda 10 for the near future, so want to stick with 10 for a while.   I did a small test with your cuda 10 suggestion with cucim, and things seem fine.  I will try some larger tests in our full stack before we make the full switch from cupyimg (which we currently use) to cucim.

Our use case is GPU-accelerated skimage rigid registration of neuroscience image processing.   The rest of our cupyimg use is in scipy, which seems to be absorbed into mainline cupy.  We also have some other existing cupy based routines, as well.

I will close this because you've provided a lot of helpful information and I should be good to go.

chrisroat:
I am finding in my cuda 10.1 environment, something seems to still want to pull in 11.0, which isn't present:

```
    from cucim.skimage import registration
../../.local/share/virtualenvs/starmap-T47byR32/lib/python3.8/site-packages/cucim/__init__.py:34: in <module>
    from .clara import CuImage
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

    import os
    
    from . import cli
    from . import converter
    # import hidden methods
[...]

PR #37, opened 2021-05-19
Fix memory leak in deflate decoder

gigony:
I did a silly mistake in deflate decoder which caused a memory leak when loading the Deflate-compressed TIFF image.

This patch will free allocated memory in decode_deflate() method.

For the next steps to prevent this kind of errors:
- Make use of a safe pointer (smart pointer) or scoped deleter for 3rdparty library/internal memory allocation.
- Run unit/integration tests with memory sanitizer/tracker (e.g., valgrind/asan)

PR #48, opened 2021-06-05
Implement Cache and __cuda_array_interface__ for v21.06.00

gigony:
I'm sorry for this big change.
This includes CI build failure fixes as well as new features/fixes/documents that couldn't be merged due to CI build failures.

@quasiben on top of your change (https://github.com/rapidsai/cucim/pull/46), the only missing piece was, adding `pytest` (https://github.com/rapidsai/cucim/pull/48/commits/8334f1f2d46738c31108ebc4303812ab9d93da0c) (which was not revealed until I added `set -x` to the build script (https://github.com/rapidsai/cucim/pull/48/commits/fae529d1589b5db98f55cc3fb941552d25d76ff9).

@jakirkham This change doesn't include a doc builder change in your PR (
https://github.com/rapidsai/cucim/pull/45/files#diff-5ed0e3039eb400d13dc55579a034925ea8d08e9f2054085b1c79fdc46a7c09a8L29) so the script may need to further updated in the next change.

- Fix memory leak in deflate decoder (#37)
- Update build script to handle failures (#45, #46)
- Fix typo in meta.yaml (#45, #46)
- Add pytest to gpu build script (#45, #46)
- Implement cache mechanism (#34)
- Add `__cuda_array_interface__`
- Support CUDA 10.x for cucim.skimage package (#44)
- Expose preferred_memory_capacity() for cache
- Fix failures due to GIL and nullptr access
- Add notebooks for cache and cuda array interface (#34)
- Update change logs for v21.06.00


gigony:
Hi @ajschmidt8, could this change be merged into 21.06.00 branch before release?

python/cucim/setup.py
```diff
@@ -49,14 +44,13 @@ def read(*names, **kwargs):
     url='https://github.com/rapidsai/cucim',
     packages=find_packages('src'),
     package_dir={'cucim': 'src/cucim'},
-    py_modules=[splitext(basename(path))[0] for path in glob('src/*.py')],
+    # # https://docs.python.org/3/distutils/setupscript.html
```
quasiben, line 47: Any reason to leave these comments here instead of removing ?  

Same question for the `package_data=` commented code below

quasiben:
I found it easier to look at the individual commits.  I left a few small comments on the `setup.py` file.  I think there is probably better to do than add the `pytest`  conda package but i'm ok pushing that off for another PR

cpp/src/io/device_type.cpp
```diff
@@ -0,0 +1,106 @@
...
+#include <algorithm>
+
+
+namespace cucim::io
+{
+
+// https://en.cppreference.com/w/cpp/algorithm/find
+#if __cplusplus < 202002L
+template <class InputIt, class UnaryPredicate>
+constexpr InputIt myfind_if(InputIt first, InputIt last, UnaryPredicate p)
```
jakirkham, line 28: Seems like we had this earlier. Should we refactor this code if it is needed? Or if `constexpr` is not needed, we can use from STL. If we do need `constexpr`, maybe Boost also has this and we can use from there?

cpp/src/io/format/image_format.cpp
```diff
@@ -162,6 +162,13 @@ ImageMetadata& ImageMetadata::level_downsamples(const std::pmr::vector<float>& l
     return *this;
 }
 
+ImageMetadata& ImageMetadata::level_tile_sizes(const std::pmr::vector<uint32_t>& level_tile_sizes)
+{
+    level_tile_sizes_ = std::move(level_tile_sizes);
```
jakirkham, line 167: Can we use `std::move` on `const &`? Or should the input be `&&`?

cpp/src/memory/memory_manager.cu
```diff
@@ -74,4 +68,53 @@ void get_pointer_attributes(PointerAttributes& attr, const void* ptr)
     }
 }
 
+bool move_raster_from_host(void** target, size_t size, cucim::io::Device& dst_device)
```
jakirkham, line 71: Can device be `const`?

```suggestion
bool move_raster_from_host(void** target, size_t size, const cucim::io::Device& dst_device)
```

cpp/src/memory/memory_manager.cu
```diff
@@ -74,4 +68,53 @@ void get_pointer_attributes(PointerAttributes& attr, const void* ptr)
...
+        {
+            throw std::bad_alloc();
+        }
+        cucim_free(host_mem);
+        *target = cuda_mem;
+    }
+    return true;
+}
+
+bool move_raster_from_device(void** target, size_t size, cucim::io::Device& dst_device)
```
jakirkham, line 97:
```suggestion
bool move_raster_from_device(void** target, size_t size, const cucim::io::Device& dst_device)
```

python/cucim/setup.py
```diff
@@ -31,13 +28,11 @@ def read(*names, **kwargs):
...
         return fh.read()
 
-# lib_paths = list(map(lambda x: basename(x), glob("src/cucim/clara/*.so*")))
-
 
 opts = dict(
     name='cucim',
     version=read('VERSION').strip(),  # versioneer.get_version(),
-    # cmdclass=versioneer.get_cmdclass()
+    cmdclass=versioneer.get_cmdclass(),
```
jakirkham, line 35: If we are using versioneer for `cmdclass`, should we use for `version` as well?

python/cucim/src/cucim/__init__.py
```diff
@@ -31,4 +31,19 @@
...
 
-from .clara import CuImage, __version__, cli
+# Try to import cupy first.
+# If cucim.clara package is imported first, you may see the following error when running on CUDA 10.x (#44)
+#   python3: Relink `/usr/lib/x86_64-linux-gnu/libnccl.so.2.8.3' with `/lib/x86_64-linux-gnu/librt.so.1' for IFUNC symbol `clock_gettime'
+#   Segmentation fault
+try:
+    import cupy
+except ImportError:
+    pass
```
jakirkham, line 41: Interesting. Do we know why that is?

python/cucim/src/localtest.py
```diff
@@ -109,34 +110,34 @@ def load_tile_cucim(slide, start_loc, tile_size):
 
 openslide_tot_time = 0
 cucim_tot_time = 0
-for num_workers in range(1, num_threads + 1):
+for num_workers in (num_threads,):  # range(1, num_threads + 1):
```
jakirkham, line 113: Why is this commented?

jakirkham (commented):
Thanks Gigon! 😄

Had some questions below

python/cucim/setup.py
gigony: Oh, no specific reason. Just for future reference. Will remove it. Thank you for pointing out that!

python/cucim/src/localtest.py
```diff
@@ -109,34 +110,34 @@ def load_tile_cucim(slide, start_loc, tile_size):
...
+    #         with concurrent.futures.ThreadPoolExecutor(
+    #             max_workers=num_workers
+    #         ) as executor:
+    #             executor.map(
+    #                 lambda start_loc: load_tile_openslide(
+    #                     slide, start_loc, tile_size),
+    #                 start_loc_iter,
+    #             )
+    #         openslide_time = timer.elapsed_time()
+    #         openslide_tot_time += openslide_time
```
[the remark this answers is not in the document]
gigony: my bad. changed the code during the local testing. Will revert it. Thanks!

python/cucim/src/localtest.py
```diff
@@ -147,8 +148,8 @@ def load_tile_cucim(slide, start_loc, tile_size):
             )
         cucim_time = timer.elapsed_time()
         cucim_tot_time += cucim_time
-    print("  Performance gain (OpenSlide/cuCIM): {}".format(
-        openslide_time / cucim_time))
+    # print("  Performance gain (OpenSlide/cuCIM): {}".format(
+    #     openslide_time / cucim_time))
```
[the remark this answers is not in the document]
gigony: Will revert it.

gigony:
Thanks @jakirkham and @quasiben for the review! I would address your comments tonight.

python/cucim/setup.cfg
```diff
@@ -2,7 +2,7 @@
 VCS = git
 style = pep440
 versionfile_source = src/cucim/_version.py
```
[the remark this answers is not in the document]
gigony: I don't think so.
`_version.py` exists in `cucim/python/cucim/src/cucim/_version.py` and when it is packaged, it would be rewritten & placed in `dist/cucim/_version.py` so only need to change `versionfile_build`

ci/gpu/build.sh
```diff
@@ -77,13 +79,13 @@ gpuci_conda_retry build -c ${LIBCUCIM_BLD_PATH} -c conda-forge -c rapidsai-night
 
 # Install cuCIM and its dependencies
 gpuci_logger "Installing cuCIM and its dependencies"
-gpuci_conda_retry install -y -c ${LIBCUCIM_BLD_PATH} -c ${CUCIM_BLD_PATH} \
-    "rapids-build-env=$MINOR_VERSION.*" \
```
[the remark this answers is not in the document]
gigony: I found that there is no rapids-build-env package whose version is `$MINOR_VERSION.*` as we switched to CalVer.

https://anaconda.org/rapidsai/rapids-build-env/files

cpp/include/cucim/cuimage.h
```diff
@@ -17,19 +17,22 @@
...
 
-#include <array>
-#include <memory>
-#include <vector>
-#include <string>
-#include <mutex>
-#include <set>
-
 #include "cucim/core/framework.h"
+
```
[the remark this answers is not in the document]
gigony: Thanks! Will remove it.

ci/gpu/build.sh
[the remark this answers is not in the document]
ajschmidt8: @gigony, this will need to be re-added. We use the `rapidsai-nightly` conda channel for pre-release packaging. You can find the cal-ver packages here: https://anaconda.org/rapidsai-nightly/rapids-build-env/files

CHANGELOG.md
```diff
@@ -1,4 +1,4 @@
-# cuCIM 21.06.00 (Date TBD)
+# cuCIM [21.06.00](https://github.com/rapidsai/cucim/wiki/release_notes_v21.06.00) (Date TBD)
```
ajschmidt8, line 1: For consistency with the changelogs of other RAPIDS libraries, this change should probably be removed. Also, we have a script that updates the changelog during each release, so it would overwrite this change.

cpp/src/memory/memory_manager.cu
gigony: Yes, it should be `const`. Will update it.

ajschmidt8 (changes requested):
Few changes requested here. Also, `branch-21.06` is in code-freeze, so the base branch for this PR will need to be updated to `branch-21.08`.

python/cucim/setup.cfg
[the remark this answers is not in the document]
jakirkham: Interesting. Ok let's leave it then :)

quasiben:
> Few changes requested here. Also, branch-21.06 is in code-freeze, so the base branch for this PR will need to be updated to branch-21.08.

For clarity, we'd like for this to be released with 21.06 as other libraries would like to pick this up -- but not part of the larger RAPIDS integration meta-package.  Is that something we could do ?

ci/gpu/build.sh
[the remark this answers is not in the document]
jakirkham: Somewhat related. I think we are missing an `activate` line (this is probably my fault). Was trying to fix that here ( https://github.com/rapidsai/cucim/pull/45/commits/495aa7316661b2efc95957d39c49f8831fb73f91 ). Think we may need to add that as well

ajschmidt8:
> > Few changes requested here. Also, branch-21.06 is in code-freeze, so the base branch for this PR will need to be updated to branch-21.08.
> 
> For clarity, we'd like for this to be released with 21.06 as other libraries would like to pick this up -- but not part of the larger RAPIDS integration meta-package. Is that something we could do ?

@quasiben, gotcha. This PR has a _lot_ of changes, so I'm not sure it'd be a good idea to merge all of this a few days before the release. Couple questions here though.

1. Which libraries require these changes?
2. Are the changes here considered "required" for `21.06` to work? Or are they just "nice-to-haves"?

jakirkham:
Several of these changes were in previous PRs that were blocked due to CI issues. So Gigon has aggregated them together. IOW they are not as new as they might seem

The changes are really limited to cuCIM and are mainly in response to needs Clara has in their downstream libraries or outside user requests. So they shouldn't effect other RAPIDS libraries

quasiben:
@gigony would be better positioned to answer these questions:

1. Which libraries require these changes?
a. MONAI is a higher lever product depending on 
2. Are the changes here considered "required" for 21.06 to work? Or are they just "nice-to-haves"?
a. The majority of these changes are in the required for 21.06.  I would hope these large code changes before releases decreases as the library matures. 

Again, @gigony can you also comment here ?

gigony:
1.

MONAI and Clara Train SDK is using cuCIM.
since they are using PyPI package and I would release new version of PyPI
package(21.06.00) separately tomorrow, conda package doesn't affect to
those projects. However, not merging thid would cause discrepancy between
the two packages (conda vs PyPI) and that's what I am worrying about.

2.

I am sorry for the last minute changes. I will try to push all the needed
change before code freeze next time.

raydouglass:
I think this is fine to merge this into `21.06` after addressing some of the request changed.

quasiben:
Thanks @raydouglass

python/cucim/setup.py
gigony: In case of versioneer, version comes from git tags, but C++ code's build system currently requires version as `VERSION` file.
I can try to refactor C++ code builder to get the version from git tags later (https://ipenguin.ws/2012/11/cmake-automatically-use-git-tags-as.html) but the current VERSION file seems okay to me.

Currently, version number from versioneer is used only when CuPy with scikit-image is available (https://github.com/rapidsai/cucim/issues/44, https://github.com/rapidsai/cucim/pull/48/commits/818dc2f00416aa7ec07aabc27c460948661ce510#diff-80a85bd6c52ce89c0dc59a606ca1205a2673c20ff860c65a3964db51ffa64ec8R47)

python/cucim/src/cucim/__init__.py
gigony: Don't know the exact root cause, but I guess,

This change is related to https://github.com/rapidsai/cucim/issues/44

If the system has only CUDA 10.x library, since cucim.clara part of the library depends on CUDA 11 and different version of runtime, it looks that it can conflict with symbols CUDA-based library that depends on CUDA 10.X

You can see version-related issues from other Github issue:
https://github.com/tensorflow/tensorflow/issues/19375

python/cucim/src/localtest.py
gigony: It is modified during the local test (since I didn't want to execute it for all available thread # configurations.)
Reverted it.

CHANGELOG.md
gigony: Oh, I see. Will remove it from here.

cpp/include/cucim/cuimage.h
```diff
@@ -162,12 +172,13 @@ class EXPORT_VISIBLE CuImage : public std::enable_shared_from_this<CuImage>
     explicit CuImage();
 
     void ensure_init();
-    bool crop_image(io::format::ImageMetadataDesc* metadata,
-                    io::format::ImageReaderRegionRequestDesc* request,
-                    io::format::ImageDataDesc* out_image_data) const;
+    bool crop_image(io::format::ImageReaderRegionRequestDesc* request, io::format::ImageDataDesc* out_image_data) const;
```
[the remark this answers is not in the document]
gigony: Good points! Thanks. It makes sense. Updated it.

cpp/include/cucim/filesystem/file_handle.h
```diff
@@ -37,11 +37,27 @@ enum class FileHandleType: uint16_t
 #if CUCIM_PLATFORM_LINUX
 struct CuCIMFileHandle
 {
+#    ifdef __cplusplus
+    EXPORT_VISIBLE CuCIMFileHandle();
+    EXPORT_VISIBLE CuCIMFileHandle(int fd, CUfileHandle_t cufile, FileHandleType type, char* path, void* client_data);
+    EXPORT_VISIBLE CuCIMFileHandle(int fd,
+                                   CUfileHandle_t cufile,
+                                   FileHandleType type,
+                                   char* path,
```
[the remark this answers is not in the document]
gigony: `CuCIMFileHandle` is a C struct and used for C interface (https://github.com/gigony/cucim/blob/465c3ef4eef8367dd047c8a9492fed2076e0d717/cpp/include/cucim/io/format/image_format.h#L185) so wouldn't want to introduce C++ type there.
And currently value for `path` is dynamically allocated so cannot be `const` (https://github.com/gigony/cucim/blob/465c3ef4eef8367dd047c8a9492fed2076e0d717/cpp/plugins/cucim.kit.cuslide/src/cuslide/tiff/tiff.cpp#L247).


Will refactor pointer-related code as part of https://github.com/rapidsai/cucim/issues/38 in the future.

cpp/plugins/cucim.kit.cuslide/benchmarks/main.cpp
```diff
@@ -86,7 +86,6 @@ static void test_basic(benchmark::State& state)
         auto handle = image_format->formats[0].image_parser.open(g_config.input_file.c_str());
 
         cucim::io::format::ImageMetadata metadata{};
-        metadata.level_count(1).level_downsamples({ 1.0 }).level_ndim(3);
```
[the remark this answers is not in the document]
gigony: Data in `metadata` would be overwritten by the following method (`image_format->formats[0].image_parser.parse(&handle, &metadata.desc());`) so no need to initialize.

```c
bool(CUCIM_ABI* parse)(CuCIMFileHandle* handle, ImageMetadataDesc* out_metadata);
```

```c++
        cucim::io::format::ImageMetadata metadata{};
        image_format->formats[0].image_parser.parse(&handle, &metadata.desc());
```
https://github.com/gigony/cucim/blob/465c3ef4eef8367dd047c8a9492fed2076e0d717/cpp/include/cucim/io/format/image_format.h#L193

cpp/plugins/cucim.kit.cuslide/src/cuslide/tiff/ifd.cpp
```diff
@@ -446,25 +488,72 @@ bool IFD::read_region_tiles(const TIFF* tiff,
...
-                                                  jpegtable_count, &tile_raster, out_device);
-                }
-                else
-                {
-                    cuslide::deflate::decode_deflate(tiff_file, nullptr, tiledata_offset, tiledata_size, &tile_raster,
-                                                     tile_raster_nbytes, out_device);
-                }
+                // clang-format off
+                // cucim::logger::Timer tt("decode_deflate: {}\n");
+                //fmt::print(stderr, "# {}: {} {} - before key\n", std::chrono::high_resolution_clock::now().time_since_epoch().count(), getpid(), index); // [print]
```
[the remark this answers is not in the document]
gigony: Yes, no need for now, and can be refactored out for as benchmarking code.
Removed them.

cpp/plugins/cucim.kit.cuslide/src/cuslide/tiff/tiff.h
```diff
@@ -17,21 +17,21 @@
...
-//#include <tiffio.h>
-
-#include <cucim/macros/api_header.h>
 #include <cucim/filesystem/file_path.h>
 #include <cucim/filesystem/file_handle.h>
-#include <cucim/memory/memory_manager.h>
 #include <cucim/io/format/image_format.h>
+#include <cucim/macros/api_header.h>
+#include <cucim/memory/memory_manager.h>
+//#include <tiffio.h>
```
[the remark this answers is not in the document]
gigony: Yes, we can. Removed.

cpp/src/cache/cache_type.cpp
```diff
@@ -0,0 +1,101 @@
...
+#include <algorithm>
+
+
+namespace cucim::cache
+{
+
+// https://en.cppreference.com/w/cpp/algorithm/find
+#if __cplusplus < 202002L
+template <class InputIt, class UnaryPredicate>
+constexpr InputIt myfind_if(InputIt first, InputIt last, UnaryPredicate p)
```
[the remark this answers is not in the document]
gigony: Yes, constexpr is needed to implement constexpr map (Inspired by https://www.youtube.com/watch?v=INn3xa4pMfg).

The method is called at both compile time and runtime.

cpp/src/io/device_type.cpp
gigony: Refactored out as `cucim/cpp/include/cucim/cpp20/find_if.h` file. and reuse the header now.
Thank you for pointing out!

cpp/src/cache/image_cache_config.cpp
```diff
@@ -0,0 +1,63 @@
...
+
+#include <fmt/format.h>
+#include <nlohmann/json.hpp>
+
+using json = nlohmann::json;
+
+namespace cucim::cache
+{
+
+void ImageCacheConfig::load_config(void* json_obj)
```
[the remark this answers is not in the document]
gigony: Made it to `const`.
However, doesn't want to declare it with `json&` because we don't want to expose `json` type to the public header (https://github.com/gigony/cucim/blob/566f130425d2c4783788776c832d15783b4f665d/cpp/include/cucim/cache/image_cache_config.h#L43) which requires either cucim library package `nlohmann::json` header or application code depends on it.

cpp/src/io/format/image_format.cpp
gigony: Thanks a lot @jakirkham for pointing out that !!!
Without your review, it would end up with inefficient code (copy instead of move).

Thank you so much!

gigony:
Thank you @quasiben @jakirkham @ajschmidt8 for the review!

I have addressed your comments except https://github.com/rapidsai/cucim/pull/48#discussion_r646838902 to make sure that this change is buildable without errors.

Could we merge this PR first and then apply @jakirkham 's patches shown below, to use rapids conda environment?

- https://github.com/rapidsai/cucim/commit/495aa7316661b2efc95957d39c49f8831fb73f91
- https://github.com/rapidsai/cucim/pull/45/files#diff-5ed0e3039eb400d13dc55579a034925ea8d08e9f2054085b1c79fdc46a7c09a8L29

PR #51, opened 2021-06-08
Update CI scripts

gigony:
Update CI scripts to use rapids environment.

This is a following up PR for #48.

Closes https://github.com/rapidsai/cucim/pull/45
Closes https://github.com/rapidsai/cucim/pull/46

### DIFF
--- a/ci/gpu/build.sh
+++ b/ci/gpu/build.sh
@@ -27,6 +27,7 @@ export HOME=$WORKSPACE
 cd $WORKSPACE
 export GIT_DESCRIBE_TAG=`git describe --abbrev=0 --tags`
 export MINOR_VERSION=`echo $GIT_DESCRIBE_TAG | grep -o -E '([0-9]+\.[0-9]+)'`
+echo "MINOR_VERSION: ${MINOR_VERSION}"
 
 # Get CUDA and Python version
 export CUDA_VERSION=${CUDA_VERSION:-$(cat /usr/local/cuda/version.txt | egrep -o "[[:digit:]]+.[[:digit:]]+.[[:digit:]]+")}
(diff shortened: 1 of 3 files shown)


quasiben:
Thanks @gigony !

jakirkham (approved):

ci/gpu/build.sh
```diff
@@ -48,14 +49,19 @@ nvidia-smi
...
 gpuci_logger "Activate conda env"
 . /opt/conda/etc/profile.d/conda.sh
+conda activate rapids
 
-conda install -c conda-forge conda-build -y
+gpuci_logger "Install dependencies"
+gpuci_conda_retry install -y -c rapidsai-nightly \
+    sysroot_linux-64=2.17 \
+    "cudatoolkit=${CUDA_VER}.*" \
+    "python=${PYTHON_VER}.*" \
```
ajschmidt8, line 58: Is there any reason why `python` is included here? our CI images should have the correct versions of `python` installed by default.

gigony: Thanks @ajschmidt8! 

Not a good reason.
I think it can be removed.

jakirkham: Do we include `cudatoolkit` as well? If so, maybe we can drop that too?

Curious if we can drop `sysroot_linux-64` (was trying that as a workaround in my older PR). Maybe we don't need that either

ajschmidt8: That's a good question. I will check if `cudatoolkit` is already installed in our images. cudf [seems to include it](https://github.com/rapidsai/cudf/blob/branch-21.06/ci/gpu/build.sh#L77-L83). I will verify regardless. Not sure about `sysroot_linux-64`. I'm not familiar with that package.

ajschmidt8 (approved):

quasiben:
@ajschmidt8 I think we need you to merge in.  I see the failed check:
>  Base branch is not under active development

quasiben:
Hmm,  looks like I was commenting as you were merging.

jakirkham:
Thanks AJ! 😄

PR #65, opened 2021-07-10
Fix missing array interface for associated_image()

gigony:
`__array_interface__` for the object returned by `CuImage.associated_image()` was missing so cannot be converted numpy.

This bug was introduced during the implementation of Cache and support for `__cuda_array_interface__`: https://github.com/rapidsai/cucim/pull/48 .

`associated_image()` API is not frequently used by users so it's fine.

We will need to add a test case for this once the artificial test image with the associated image (such as Philips TIFF) is available.

### DIFF
--- a/python/pybind11/cucim_py.cpp
+++ b/python/pybind11/cucim_py.cpp
@@ -148,7 +148,7 @@ PYBIND11_MODULE(_cucim, m)
              py::arg("shm_name") = "") //
         .def_property("associated_images", &CuImage::associated_images, nullptr, doc::CuImage::doc_associated_images,
                       py::call_guard<py::gil_scoped_release>()) //
-        .def("associated_image", &CuImage::associated_image, doc::CuImage::doc_associated_image,
+        .def("associated_image", &py_associated_image, doc::CuImage::doc_associated_image,
              py::call_guard<py::gil_scoped_release>(), //
              py::arg("name") = "", //
              py::arg("device") = io::Device()) //
@@ -392,6 +392,22 @@ py::object py_read_region(const CuImage& cuimg,
         _set_array_interface(region);
 
         return region;
+    }
+}
+
+py::object py_associated_image(const CuImage& cuimg, const std::string& name, const io::Device& device)
+{
+    auto image_ptr = std::make_shared<cucim::CuImage>(cuimg.associated_image(name, device));
+
+    {
+        py::gil_scoped_acquire scope_guard;
+
+        py::object image = py::cast(image_ptr);
+
+        // Add `__array_interace__` or `__cuda_array_interface__` in runtime.
+        _set_array_interface(image);
+
+        return image;
     }
 }
 
--- a/python/pybind11/cucim_py.h
+++ b/python/pybind11/cucim_py.h
@@ -68,6 +68,8 @@ py::object py_read_region(const CuImage& cuimg,
                           const py::object& buf,
                           const std::string& shm_name,
                           const py::kwargs& kwargs);
+py::object py_associated_image(const CuImage& cuimg, const std::string& name, const io::Device& device);
+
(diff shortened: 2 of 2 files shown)


grlee77 (commented):
This looks okay to me, but I have only rarely looked at pybind11 code, so it would be nice to have a test case to be 100% sure. Can you confirm that you tested the fix locally with a TIFF file?

Is there not some small public domain TIFF we could add as test data?

gigony:
Thanks @grlee77!

I have tested it locally with a Philips Tiff file having associated image
so I can confirm it is working properly with the fix.

I couldn't find some very small size of such test data so I would like to
add test case for that later with generating an artificial test data with
associated image during test time by using PyTest's fixture in the separate
PR.

python/pybind11/cucim_py.cpp
```diff
@@ -395,6 +395,22 @@ py::object py_read_region(const CuImage& cuimg,
...
+{
+    auto image_ptr = std::make_shared<cucim::CuImage>(cuimg.associated_image(name, device));
+
+    {
+        py::gil_scoped_acquire scope_guard;
+
+        py::object image = py::cast(image_ptr);
+
+        // Add `__array_interace__` or `__cuda_array_interface__` in runtime.
+        _set_array_interface(image);
```
jakirkham, line 408: Wondering if we should acquire the GIL within `_set_array_interface`. WDYT?

jakirkham (commented):
Thanks Gigon! 😄

Had a question about how we handle acquiring the GIL here

python/pybind11/cucim_py.cpp
gigony: The line 403 (https://github.com/rapidsai/cucim/pull/65/files#diff-48de51ebe61a16c8d3878cba6d31c628f024287fc2671440f9bae39ab6bcbd64R403 ) is for acquiring GIL :)

jakirkham: Was thinking about this ( https://github.com/rapidsai/cucim/blob/372b39d4d196145f7c7cbc76b27c1abe9cb89183/python/pybind11/cucim_py.cpp#L104-L106 ), but I guess we already don't release the GIL (so should have it). So agree it's a non-issue

jakirkham (approved):

jakirkham:
@gpucibot merge

jakirkham:
Thanks Gigon! 😄

gigony:
> Thanks Gigon! 😄
> 
> Had a question about how we handle acquiring the GIL here


The line 403 (https://github.com/rapidsai/cucim/pull/65/files#diff-48de51ebe61a16c8d3878cba6d31c628f024287fc2671440f9bae39ab6bcbd64R403 ) is for acquiring GIL :) 

Before 20.06.00, CuImage object always had `__array_interface__` attributes, and depending on the availability of the array data, it exposed correct `__array_interface__` content or None/{} value. GIL could be released with the mechanism, to speed up in multi-thread use cases.

However, from  20.06.00, we introduced supporting [`__cuda_array_interface__`](https://nbviewer.jupyter.org/github/gigony/cucim/blob/v21.06.00/notebooks/Basic_Usage.ipynb#__cuda_array_interface__-support). Due to how existing frameworks (CuPy, PyTorch, NumPy) worked, I should be able to remove one of those two attributes (`__array_interface__` and `__cuda_array_interface__`) when creating CuImage object (by read_region() or associated_image()). However, there is no native C++ way to create or delete attributes dynamically (I saw you raised issues https://github.com/pybind/pybind11/issues/21 but it is close now).

For this reason, it needed to acquire GIL (for using Python objects in C++).
It could affect performance when used in multi-threading environment, but I did a quick benchmark (with v0.19.0) and couldn't find a significant performance degradation. 
However, we can revisit this issue when the performance issue happens.

Thanks!

PR #72, opened 2021-07-20
Remove int-type bug on Windows in skimage.measure.label

grlee77:
A windows user reported an issue with `skimage.measure.label` in the older `cupyimg` repository: https://github.com/mritools/cupyimg/issues/21.

It seems the same issue is present here, but should be resolved by this PR. I don't think we can easily add a test case on our current linux-based CI as the issue has to do with the NumPy dtype character used to represent the `numpy.int32` dtype. We may want to wait on merging until the fix is confirmed in https://github.com/mritools/cupyimg/issues/21

I think I originally introduced this `int_types` dict with the intention of also supporting >32-bit integers, but it was incomplete and still used `int32` in the actual implementation. This PR just removes this dict altogether and we can revisit the function if we get a request for int64 support.

### DIFF
--- a/python/cucim/src/cucim/skimage/measure/_label_kernels.py
+++ b/python/cucim/src/cucim/skimage/measure/_label_kernels.py
@@ -21,22 +21,10 @@ def _label(x, structure, y, greyscale_mode=False):
     y_shape = cupy.array(y.shape, dtype=numpy.int32)
     count = cupy.zeros(2, dtype=numpy.int32)
     _kernel_init()(x, y)
-    try:
-        int_t = int_types[y.dtype.char]
-    except KeyError:
-        raise ValueError("y must have int32, uint16, uint32 or uint64 dtype")
-    if int_t != "int":
-        raise NotImplementedError(
-            "Currently only 32-bit integer case is implemented"
-        )
     if greyscale_mode:
-        _kernel_connect(True, int_t)(
-            x, y_shape, dirs, ndirs, x.ndim, y, size=y.size
-        )
+        _kernel_connect(True)(x, y_shape, dirs, ndirs, x.ndim, y, size=y.size)
     else:
-        _kernel_connect(False, int_t)(
-            y_shape, dirs, ndirs, x.ndim, y, size=y.size
-        )
+        _kernel_connect(False)(y_shape, dirs, ndirs, x.ndim, y, size=y.size)
     _kernel_count()(y, count, size=y.size)
     maxlabel = int(count[0])  # synchronize
     labels = cupy.empty(maxlabel, dtype=numpy.int32)
@@ -184,11 +172,3 @@ def _kernel_finalize():
         """,
         "cucim_nd_label_finalize",
     )
-
-
-int_types = {
-    "i": "int",
-    "H": "unsigned short",
-    "I": "unsigned int",
-    "L": "unsigned long long",
-}


python/cucim/src/cucim/skimage/measure/_label_kernels.py
```diff
@@ -21,22 +21,10 @@ def _label(x, structure, y, greyscale_mode=False):
     y_shape = cupy.array(y.shape, dtype=numpy.int32)
     count = cupy.zeros(2, dtype=numpy.int32)
     _kernel_init()(x, y)
-    try:
-        int_t = int_types[y.dtype.char]
-    except KeyError:
-        raise ValueError("y must have int32, uint16, uint32 or uint64 dtype")
```
jakirkham, line 27: Am curious why this exception is no longer needed. What happens if we don't raise this?

python/cucim/src/cucim/skimage/measure/_label_kernels.py
```diff
@@ -21,22 +21,10 @@ def _label(x, structure, y, greyscale_mode=False):
...
-    if int_t != "int":
-        raise NotImplementedError(
-            "Currently only 32-bit integer case is implemented"
-        )
```
jakirkham, line 31: Same question here

jakirkham (commented):
Thanks Greg! 😄 

Had a couple quick questions below

python/cucim/src/cucim/skimage/measure/_label_kernels.py
grlee77: I is a private function that is only called here with an input that is hard-coded to `np.int32`.

https://github.com/rapidsai/cucim/blob/1aa5802dd247c277f0ea82affadf99469a2752a7/python/cucim/src/cucim/skimage/measure/_label.py#L120-L121

The kernels in this file originate from CuPy, but have minor modifications here to match `scipy.skimage.measure.label` behavior rather than `scipy.ndimage.label`. The int checking stuff is not in upstream CuPy, I  think I had originally added these so we could switch to int64 if needed, but I'm not sure there is a real use case for > 2 billion labels?

jakirkham: So what happens if a user does pass in an `int64` array?

Should add this most likely would happen just because a user requested a `dtype` of `int`, which would be handled as `int64`. Not so much because they need more labels.

grlee77: > So what happens if a user does pass in an int64 array?

You will get the following compilation error:
```
E   /tmp/tmp0iyn3i4v/3ac4bcb15a8927c9397d795e29cb7ccd_2.cubin.cu(43): error: no instance of overloaded function "atomicCAS" matches the argument list
E               argument types are: (long long *, Y, Y)
```
I can add a check in `_label` that `np.dtype(input) == np.float32` with a message like the old one if you like. That can only happen if users are calling that (private) `_label` function directly, though!

jakirkham: Ah ok. In that case would it make sense to raise in `label`? Also do we want to update this todo?

https://github.com/rapidsai/cucim/blob/00cd5d6c71cda6d1073daed5523923c243cd9c09/python/cucim/src/cucim/skimage/measure/_label.py#L16-L17

grlee77: There is no restriction on the type of `input` here which is the only parameter passed on to `_label` that comes from the user. The `structure` and `labels` array inputs to `_label` are generated internally, not provided by the user. 

The only way the user can encounter the error cited here https://github.com/rapidsai/cucim/pull/72#discussion_r675142312 would be if they were to import the private `_label` function from `cucim.skimage.measure._label_kernels`.

jakirkham (approved):

jakirkham:
@gpucibot merge

jakirkham:
Thanks Greg! 😄

PR #73, opened 2021-07-21
Pin `isort` hook to 5.6.4

charlesbluca:
Pins the `isort` pre-commit hook to 5.6.4 to correspond with the version bump in gpuCI (see rapidsai/integration#286). Also moves the pre-commit configuration to the root directory so it can be detected when running `pre-commit`.


jakirkham:
@gpucibot merge

jakirkham (approved):

PR #106, opened 2021-09-28
Enable GDS and Support Runtime Context (__enter__, __exit__) for CuFileDriver and CuImage

### LINKED ISSUE #3: Enable GDS with public package for v0.19.0
GDS feature is disabled for now due to no public cuFile package available.
Once GDS package is released publicly(expected to release this week), need to update the code to use cufile.h header from the public package and/or update conda package to use the public gds package.

gigony:
- Enable GDS with GDS v1.0.0
  - Re-enable GDS with public package
  - Found an issue with a GDS static library and working on WSL
    1. Segmentation fault error is happening if we link GDS static library because GDS static library file (libcufile.a) available does not support non-CXX11 ABI (`-D_GLIBCXX_USE_CXX11_ABI=0`).  Python wheel with manylinux2014 needs to be compiled with non-CXX ABI
        - Workaround the static library issue by using a shared library (if exists on the system) instead.
    2. Loading/using GDS shared library(libcufile.so) causes `Assertion failure, file index :cufio-udev  line :143` error on WSL
        - Workaround the WSL issue by not loading GDS shared library (calling `cuFileDriverOpen()`) if the platform is WSL.
- Do not use GDS if in WSL
- Add `__enter__` and `__exit__` for CuFileDriver
- Add `__enter__` and `__exit__` for CuImage

Resolves #3

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -103,7 +103,7 @@ include(ExternalProject)
 ################################################################################
 # Options
 ################################################################################
-option(CUCIM_SUPPORT_GDS "Support cufile library" OFF)
+option(CUCIM_SUPPORT_GDS "Support cufile library" ON)
 option(CUCIM_STATIC_GDS "Use static cufile library" OFF)
 option(CUCIM_SUPPORT_CUDA "Support CUDA" ON)
 
--- a/cpp/CMakeLists.txt
+++ b/cpp/CMakeLists.txt
@@ -65,6 +65,7 @@ add_library(${CUCIM_PACKAGE_NAME}
         include/cucim/memory/memory_manager.h
         include/cucim/util/cuda.h
         include/cucim/util/file.h
+        include/cucim/util/platform.h
         include/cucim/3rdparty/dlpack/dlpack.h
         include/cucim/3rdparty/dlpack/dlpackcpp.h
         src/cuimage.cpp
@@ -95,7 +96,8 @@ add_library(${CUCIM_PACKAGE_NAME}
         src/logger/logger.cpp
         src/logger/timer.cpp
         src/memory/memory_manager.cu
-        src/util/file.cpp)
+        src/util/file.cpp
+        src/util/platform.cpp)
 
 # Compile options
 set_target_properties(${CUCIM_PACKAGE_NAME}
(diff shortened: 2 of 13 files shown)


gigony:
CI/CD issue:
https://gpuci.gpuopenanalytics.com/job/rapidsai/job/gpuci/job/cucim/job/prb/job/cucim-cpu-cuda-build/CUDA=11.0/56/console

Line 958:
```
12:48:16     raise CondaMultiError(exceptions)
12:48:16 conda.CondaMultiError: Downloaded bytes did not match Content-Length
12:48:16   url: https://conda.anaconda.org/conda-forge/linux-64/cudatoolkit-11.0.3-h15472ef_8.tar.bz2
12:48:16   target_path: /opt/conda/envs/rapids/pkgs/cudatoolkit-11.0.3-h15472ef_8.tar.bz2
12:48:16   Content-Length: 998168165
12:48:16   downloaded bytes: 232106833
```

jakirkham (commented):
Thanks Gigon! 😄 Looks good

So are we going to include this in Conda and Wheels somehow?

gigony:
> So are we going to include this in Conda and Wheels somehow?

yes, We want to include this patch in v21.10.00

jakirkham:
Sorry what I mean is how are the GDS libraries included in the packages? Or is this up to user to install somewhere as well? If the latter, where should they be installing the GDS libraries and what happens if they are not there?

gigony:
> Sorry what I mean is how are the GDS libraries included in the packages? Or is this up to user to install somewhere as well? If the latter, where should they be installing the GDS libraries and what happens if they are not there?

It's up to user to install GDS libraries (GDS libraries are available through CUDA Toolkit >=11.x . I need to check from which version libcufile is included but I can confirm that CUDA Toolkit 11.4 include it), or can install manually by following the instruction https://docs.nvidia.com/gpudirect-storage/troubleshooting-guide/index.html#gds-installing

If GDS library is not available, the filesystem API still works with compatibility mode (our own implementation to do the same behavior).

jakirkham:
Ah ok it's in the CUDA Toolkit. That was all I needed to know. Thanks Gigon! 😄

jakirkham (approved):

jakirkham:
@gpucibot merge

PR #19, opened 2021-04-22
Check compression method used in the image (#17)

gigony:
SVS format is not supported for now but we loosen the restriction by
delegating its process to libtiff's TIFFRGBAImageGet() method.
cuCIM assumes that only one image with Subfile Type(=0) is available in
the image file but apparently, some SVS files have also only one Subfile
Type with zero so bypassing the check.

This patch explicitly checks SVS format and compression methods and
raises an exception if the given file doesn't match the criteria.
Once SVS format is supported, those changes need to be updated to not
raise exceptions for SVS files.

Resolves: #17

Issue #17, opened 2021-04-21
[BUG] Compression scheme 33003 tile decoding is not implemented.


OlivierDehaene:
**Describe the bug**

For Aperio svs files, I run into a cryptic error:

```/slide.svs: Compression scheme 33003 tile decoding is not implemented.```

However, cuCMI still loads a tile and doesn't raise but the color are wrong or it is entirely black. See the following grid of tiles:

![image](https://user-images.githubusercontent.com/23298448/115588945-93691700-a2cf-11eb-801a-b120fda30a6a.png)

Are all OpenSlide vendors supported? Is there a list of all supported vendors somewhere? I could not find it.

**Steps/Code to reproduce bug**

```python
# Download the TCGA-F4-6459-01Z-00-DX1 slide from https://portal.gdc.cancer.gov/cases/9fd08502-355b-4f5a-a25c-73ec7184f6d3?bioId=57f4136d-d6aa-4537-ba73-f21fe0374005 

import numpy as np

from PIL import Image
from cucmi import CuImage

slide = CuImage("TCGA-F4-6459-01Z-00-DX1")

tile = slide.read_region((0, 0), (224, 224), 0)

Image.fromarray(np.asarray(tile))
```

gigony:
Hello @OlivierDehaene !
Thank you for your interest in cuCIM!

Compression scheme 33003 is actually JPEG2000 which is used in Aperio SVS format.

cuCIM currently supports tiled RGB TIFF image(Generic TIFF or Philips TIFF) with JPEG or Deflate compressed and doesn't support SVS format yet.
However, we have a plan to support it in the near future.

The tentative plan is specified in the following link (I would move the content to Github wiki page later):
  https://github.com/rapidsai/cucim/blob/branch-0.20/python/cucim/docs/roadmap/index.md#facalendar-alttext-info-mr-1-v020

If cuCIM encounters TIFF (SVS is a variant of TIFF format) file that is not JPEG/Deflate compressed RGB format, it just delegates its process to libtiff's [TIFFRGBAImageGet()](https://linux.die.net/man/3/tiffrgbaimageget) method and it seems that warning/error are not properly handled.

Let me fix the issue by raising an error until cuCIM supports SVS format properly.
(I am currently working on providing a cache mechanism to cuCIM so that the performance could be greatly improved)

Thank you for the report! We are also eager to support various image formats so stay tuned!

OlivierDehaene:
Thank you for your answer @gigony!

[...]

PR #108, opened 2021-09-28
Support raw RGB tiled TIFF

gigony:
Due to #17 ([BUG] Compression scheme 33003 tile decoding is not implemented), #19 (Check compression method used in the image) was merged so cuCIM has been handling only particular formats (only jpeg/deflate-compressed image).

If the input TIFF image has no-compressed RAW tile image, cuCIM showed the following error message:

> RuntimeError: This format (compression: 1, sample_per_pixel: 1, planar_config: 1, photometric: 1) is not supported yet!. https://github.com/Project-MONAI/MONAI/pull/2987/checks?check_run_id=3667530814#step:7:15641

This patch is to support raw RGB tiled TIFF with fast-path, re-allowing non-compressed image with slow-path (in case the input is not tiled RGB image).

This supports https://github.com/Project-MONAI/MONAI/pull/2987.

### DIFF
--- a/cpp/plugins/cucim.kit.cuslide/src/cuslide/raw/raw.cpp
+++ b/cpp/plugins/cucim.kit.cuslide/src/cuslide/raw/raw.cpp
@@ -1,0 +1,71 @@
+/*
+ * Apache License, Version 2.0
+ * Copyright 2021 NVIDIA Corporation
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+/**
+ * Code below is using libdeflate library which is under MIT license
+ * Please see LICENSE-3rdparty.md for the detail.
+ */
+
+#include "raw.h"
+
+#include <cstring>
+#include <stdexcept>
+#include <unistd.h>
+
+
+namespace cuslide::raw
+{
+
+bool decode_raw(int fd,
+                unsigned char* raw_buf,
+                uint64_t offset,
+                uint64_t size,
+                uint8_t** dest,
+                uint64_t dest_nbytes,
+                const cucim::io::Device& out_device)
+{
+    (void)out_device;
+
+    if (raw_buf == nullptr)
+    {
+        if ((raw_buf = (unsigned char*)malloc(size)) == nullptr)
+        {
+            throw std::runtime_error("Unable to allocate buffer for raw data!");
+        }
+
+        if (pread(fd, raw_buf, size, offset) < 1)
+        {
+            throw std::runtime_error("Unable to read file for raw data!");
+        }
+    }
+    else
+    {
+        fd = -1;
+        raw_buf += offset;
+    }
+
+    memcpy(*dest, raw_buf, dest_nbytes);
+
+    if (fd != -1)
+    {
+        free(raw_buf);
+    }
+
+    return true;
+}
+
+} // namespace cuslide::raw
--- a/cpp/plugins/cucim.kit.cuslide/src/cuslide/raw/raw.cpp
+++ b/cpp/plugins/cucim.kit.cuslide/src/cuslide/raw/raw.cpp
@@ -40,6 +40,20 @@ bool decode_raw(int fd,
 {
     (void)out_device;
 
+    if (dest == nullptr)
+    {
+        throw std::runtime_error("'dest' shouldn't be nullptr in decode_raw()");
+    }
+
+    // Allocate memory only when dest is not null
+    if (*dest == nullptr)
+    {
+        if ((*dest = (unsigned char*)malloc(dest_nbytes)) == nullptr)
+        {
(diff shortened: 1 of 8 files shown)


grlee77 (commented):
This looks good, but is there a simple test we can add to ensure it is working as expected? Maybe read a simple raw TIFF generated in PIL via Python?

gigony:
> This looks good, but is there a simple test we can add to ensure it is working as expected? Maybe read a simple raw TIFF generated in PIL via Python?

Thanks @grlee77 for the review!
Sure. Will add a test case soon.

gigony:
@grlee77 Added unit tests for this : https://github.com/rapidsai/cucim/pull/108/commits/537c76ee9d85237cef605fe864ac750ac3c3da1e

grlee77 (approved):
Thanks, it looks like the test failures here will be resolved by #110

cpp/plugins/cucim.kit.cuslide/src/cuslide/raw/raw.cpp
```diff
@@ -0,0 +1,71 @@
...
+
+
+namespace cuslide::raw
+{
+
+bool decode_raw(int fd,
+                unsigned char* raw_buf,
+                uint64_t offset,
+                uint64_t size,
+                uint8_t** dest,
```
jakirkham, line 37: Why does this need to be `**` as opposed to `*`?

cpp/plugins/cucim.kit.cuslide/src/cuslide/raw/raw.cpp
```diff
@@ -0,0 +1,71 @@
...
+            throw std::runtime_error("Unable to read file for raw data!");
+        }
+    }
+    else
+    {
+        fd = -1;
+        raw_buf += offset;
+    }
+
+    memcpy(*dest, raw_buf, dest_nbytes);
```
jakirkham, line 61: Should we do a `nullptr` check for `dest` too?

cpp/plugins/cucim.kit.cuslide/src/cuslide/tiff/ifd.cpp
```diff
@@ -365,13 +366,22 @@ const std::vector<uint64_t>& IFD::image_piece_bytecounts() const
...
-            compression_ == COMPRESSION_DEFLATE);
+    switch (compression_)
+    {
+    case COMPRESSION_NONE:
+    case COMPRESSION_JPEG:
+    case COMPRESSION_ADOBE_DEFLATE:
+    case COMPRESSION_DEFLATE:
+        return true;
+    default:
+        return false;
```
jakirkham, line 377: Might be worth indenting these to make this a bit more readable

jakirkham (commented):
Thanks Gigon! 😄

Had a couple questions above

cpp/plugins/cucim.kit.cuslide/src/cuslide/raw/raw.cpp
gigony: For the same reason that cudaMalloc accept '**' (https://docs.nvidia.com/cuda/cuda-runtime-api/group__CUDART__MEMORY.html#group__CUDART__MEMORY_1g37d37965bfb4803b6d4e59ff26856356) -- decoders may need to allocate memory and setting the address to `dest` if needed.


See https://github.com/rapidsai/cucim/blob/048d53e6f99c2129b9febd08e0ae6d1b37d74451/cpp/plugins/cucim.kit.cuslide/src/cuslide/jpeg/libjpeg_turbo.cpp#L145

cpp/plugins/cucim.kit.cuslide/src/cuslide/raw/raw.cpp
gigony: Good catch!
Will check it and allocate memory if it is nullptr.

cpp/plugins/cucim.kit.cuslide/src/cuslide/tiff/ifd.cpp
gigony: Thanks! Will do.

gigony (commented):
Thanks @jakirkham for the review! Will update it soon.

cpp/plugins/cucim.kit.cuslide/src/cuslide/tiff/ifd.cpp
jakirkham: Ah ok. Maybe we can track C++ formatting clean as an issue

cpp/plugins/cucim.kit.cuslide/src/cuslide/raw/raw.cpp
jakirkham: Was mostly thinking about this in terms of the `nullptr` checks below. We may want to look into using references for this kind of thing in the future, but not urgent. Can be addressed later

cpp/plugins/cucim.kit.cuslide/src/cuslide/raw/raw.cpp
jakirkham: Thanks! 😄

jakirkham:
rerun tests

(looks like one of the jobs failed due to some sort of Jenkins issue)

cpp/plugins/cucim.kit.cuslide/src/cuslide/tiff/ifd.cpp
gigony: Yeah, will create an issue for that.

gigony: https://github.com/rapidsai/cucim/issues/116

jakirkham (approved):

jakirkham:
@gpucibot merge

PR #128, opened 2021-10-13
Update initial cpp unittests

gigony:
Update existing unit test code to use generated test image and allow external image specified by the parameter.

Updating `run` script to pre-generate the necessary test images and running C++ unit tests would be set up in a separate PR.


benchmarks/config.h
```diff
@@ -42,6 +43,38 @@ struct AppConfig
     std::string benchmark_color; //  {auto|true|false}
     std::string benchmark_counters_tabular;
     std::string v; // <verbosity>
+
+    std::string get_input_path(const char* default_value = "generated/tiff_stripe_4096x4096_256.tif") const
```
jakirkham, line 47: Should `default_value` be `std::string`? Similar lines below

benchmarks/config.h
```diff
@@ -21,7 +21,8 @@
 
 struct AppConfig
 {
-    std::string input_file = "test_data/private/generic_tiff_000.tif";
+    std::string test_folder;
+    std::string test_file;
```
jakirkham, line 25: We may want to revisit at some point and look at using [`std::filesystem::path`]( https://en.cppreference.com/w/cpp/filesystem/path ). Though that requires C++17 and we were already using `std::string`. So just a future note

jakirkham (commented):
Thanks Gigon! 😄

Had a couple suggestions above

benchmarks/config.h
gigony: Thanks! Updated to `const std::string`

benchmarks/config.h
gigony: Thanks!
Created a issue for following up:
- https://github.com/rapidsai/cucim/issues/136

jakirkham (approved):

jakirkham:
@gpucibot merge

PR #141, opened 2021-11-03
Support Aperio SVS with CPU LZW and jpeg2k decoder

gigony:
**Add tif_lzw.c from libtiff as it is**
Copy tif_lzw.c file from libtiff for porting lzw decoder.
The code is from the following link:
  https://gitlab.com/libtiff/libtiff/-/blob/8546f7ee994eacff0a563918096f16e0a6078fa2/libtiff/tif_lzw.c
, which is after v4.3.0

**Support Aperio SVS with CPU LZW and jpeg2k decoder**
- Update Jpeg decoder to consider colorspace of the JPEG image
  - We need to set a proper color space from TIFF metadata to the decoder
  - Update tjDecompress2() method call with jpeg_decode_buffer() that modified tjDecompress2() implementation
  - Add `jpeg_color_space_` to IFD Class.
- Move additional includes to their own CMake files
- Add Jpeg2k decoder with OpenJpeg
  - Added fast color conversion logic with pre-calculated table
- Add LZW decoder with libtiff's implementation
  - Take part of the code to provide only LZW decoder part
- Add `rows_per_strip_` and `predictor_` attributes to IFD class, to support LZW
- Refactor TIFF:resolve_vendor_format() to support Aperio SVS metadata
- Support LZW compressed image (with multi strips) for the associated image.

Addresses #17

image

image

image

PR #145, opened 2021-11-10
Fix thread contention in Cache

### LINKED ISSUE #146: [BUG] Performance is degraded for some range of threads when using multithreads
**Describe the bug**

image

When multiple threads are used, it can cause thread contentions so
performance can be degraded.

```bash
experiment_thread("per_process", "notebooks/input/CMU-1-JP2K-33005.svs", 9, 120, 240)
```

**Expected behavior**
Performance should be stable.

See https://nbviewer.org/github/gigony/cucim/blob/supporting_svs_notebook/notebooks/Supporting_Aperio_SVS_Format.ipynb

**Additional context**

gigony:
Needs to rebase once #141 is merged.

When multiple threads are used, it can cause thread contentions so
performance can be degraded.

image

It is found that the performance degradation is due to `lock()` method in cuCIM's cache implementation.

image

This patch fixes the issue by using a hash value for the index of the mutex pool.
Also, the default mutex pool size is increased from 11117 to 100003.

image

### After Fix

Note:: experiment on JPEG-compressed SVS had an incorrect configuration (start location and patch size was specified as (128,256) instead of (120, 240)) so updated it in both documents and [Google Spreadsheet](https://docs.google.com/spreadsheets/d/15D1EqNI_E9x_S8i3kJLwBxMcEmwk8SafW0WryMrAm6A/edit?usp=sharing).
image

image

image

Close #146

### DIFF
--- a/cpp/include/cucim/cache/image_cache_config.h
+++ b/cpp/include/cucim/cache/image_cache_config.h
@@ -28,7 +28,26 @@ constexpr uint64_t kOneMiB = 1024UL * 1024;
 constexpr std::string_view kDefaultCacheTypeStr = "nocache";
 constexpr CacheType kDefaultCacheType = cucim::cache::CacheType::kNoCache;
 constexpr uint64_t kDefaultCacheMemoryCapacity = 1024UL;
-constexpr uint32_t kDefaultCacheMutexPoolCapacity = 11117;
+/**
+ * @brief Mutex Pool size
+ *
+ * >>> from functools import reduce
+ * >>> def calc(pool_size, thread_size):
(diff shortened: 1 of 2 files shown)


jakirkham:
PR ( https://github.com/rapidsai/cucim/pull/141 ) is in. So this is ready for rebasing

gigony:
> PR ( #141 ) is in. So this is ready for rebasing

@jakirkham 
Please help merge this when you have time.

jakirkham (approved):

jakirkham:
@gpucibot merge

PR #148, opened 2021-11-16
Build with NumPy 1.17

jakirkham:
Instead of letting `host`'s `numpy` have a minimum, which could float to a high version that is in turn pinned to a high version in `run`, build using the minimum required `numpy` version. That way the minimum is set similarly in `host` and a wide range of `numpy`'s will still work at runtime.

cc @quasiben (for awareness)

### DIFF
--- a/conda/recipes/cucim/meta.yaml
+++ b/conda/recipes/cucim/meta.yaml
@@ -33,7 +33,7 @@ requirements:
     - libcucim {{ version }}.*
     - click
     - cupy 9.*
-    - numpy >=1.17.0
+    - numpy 1.17.0
     - scipy
     - scikit-image 0.18.1
   run:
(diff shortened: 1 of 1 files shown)


jakirkham:
In particular this caused some issues when installing `numba` into the same environment as it [doesn't always support the latest `numpy`]( https://github.com/conda-forge/numba-feedstock/blob/5d5c1498b41bce6804fac9c480eab32d39bba09f/recipe/meta.yaml#L61-L62 ), but `cucim` was effectively being built with and requiring the latest `numpy`.

raydouglass (approved):

jakirkham:
rerun tests

(one of the jobs failed, but it looks unrelated)

jakirkham:
rerun tests

(pending job was likely lost during rename)

jakirkham:
@gpucibot merge

jakirkham:
Debugging in PR ( https://github.com/rapidsai/cucim/pull/153 )

ajschmidt8:
the segfault failure in this PR is unrelated to the `numpy` version change. I will admin merge this PR to unblock some other `conda` solve issues downstream that this PR resolves.

PR #153, opened 2021-11-17
[DEBUG] Dump GDB back traces on Segmentation Fault

gigony:

PR #158, opened 2021-11-20
Fix a segmentation fault due to unloaded libcufile

gigony:
Do not unload the `cufile (GDS)` library because `libcufile` registers a cleanup function with `atexit()` and unloading the library will cause a segfault (calling the cleanup function that doesn't exist anymore).

It turns out that CUDA 11.5 bundled GDS library (libcufile) and it is available in GPUCI's build/test container image (such as `gpuci/rapidsai:21.12-cuda11.5-devel-ubuntu18.04-py3.7`).
cuCIM would dynamically load `libcufile.so` shared library and unload it when a global static variable in cuCIM is destroyed.
Since libcufile's cleanup function(through `atexit()`) is called after the libcufile is unloaded, it causes a segmentation fault at exit.
(See https://github.com/rapidsai/cucim/pull/153)

You can see discussions with `atexit in dynamically loaded shared library` keywords
- https://news.ycombinator.com/item?id=20228082
- https://forums.raspberrypi.com/viewtopic.php?t=237006

Maybe using [destructor attribute](https://www.geeksforgeeks.org/__attribute__constructor-__attribute__destructor-syntaxes-c/) could fix the issue from GDS(cufile) side.

This patch leaves the libcufile library loaded, without calling `::dlclose(library_handle)` method to unload `libcufile.so`.

Update: I couldn't find `atexit()` used in libcufile (though I can see `atexit()` call for a executable file[fio]) but it seems that a method is registered and called at the exit time so we cannot help but leave the dynamically loaded library without unloading.

### DIFF
--- a/cpp/src/filesystem/cufile_driver.cpp
+++ b/cpp/src/filesystem/cufile_driver.cpp
@@ -41,6 +41,7 @@ namespace cucim::filesystem
 {
 static constexpr unsigned int PAGE_SIZE = 4096;
 static constexpr uint64_t DEFAULT_MAX_CACHE_SIZE = 128 << 20; // 128MiB
+static CuFileStub s_cufile_stub;
 static CuFileDriverInitializer s_cufile_initializer;
 thread_local static CuFileDriverCache s_cufile_cache;
 Mutex CuFileDriver::driver_mutex_;
@@ -300,7 +301,7 @@ bool discard_page_cache(const char* file_path)
 CuFileDriverInitializer::CuFileDriverInitializer()
 {
     // Initialize libcufile library
-    open_cufile_stub();
+    s_cufile_stub.load();
 
     CUfileError_t status = cuFileDriverOpen();
     if (status.err == CU_FILE_SUCCESS)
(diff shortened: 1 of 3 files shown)


jakirkham (approved):

gigony:
Need to take some time to dig the libcufile in detail if it is actually due to `atexit()` call, or cuCIM's destructor (by a static variable) is calling an unloaded method in libcufile.

gigony:
@jakirkham More update on the root-cause:

libcufile.so started using `thread_local` variable since v1.1 which makes the shared library unloadable.
For this reason, this patch is a correct patch until libcufile is updated to make it possible.

Related information: 
- https://sourceware.org/glibc/wiki/Destructor%20support%20for%20thread_local%20variables
- https://github.com/emilk/loguru/issues/146
- https://maskray.me/blog/2021-02-14-all-about-thread-local-storage

gigony:
GDS team suggested using `RTLD_NODELETE` when calling dlopen() so that the library is not unloaded.
They will fix the issue by adding `-z nodelete` in the link flags when building libcufile.so
```
dlopen("/usr/local/cuda/targets/x86_64-linux/lib/libcufile.so", RTLD_NOW | RTLD_LOCAL| RTLD_NODELETE);
```
see #177

PR #172, opened 2021-12-01
add cucim.skimage.__api_version__

grlee77:
This attribute will indicate which version of the scikit-image API has been implemented (currently 0.18). Since we use an independent versioning scheme from scikit-image, this is otherwise not obvious to end users.

### DIFF
--- a/docs/source/api.rst
+++ b/docs/source/api.rst
@@ -17,6 +17,10 @@ CuImage
 skimage Submodules
 ==================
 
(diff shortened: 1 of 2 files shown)


jakirkham:
How will this be used? Also do you have a link to the previous discussion where this came up?

grlee77:
I am not positive if the attribute would see much use. Maybe we don't need it and should just clearly metnion the version we are targeting in the docs?

it is not a problem if this isn't merged for 21.12

PR #186, opened 2021-12-18
Stain Normalizer

### LINKED ISSUE #96: [FEA] Add Stain Normalization operation (in Python)

We need to add Stain Normalization algorithm to a part of cuCIM's operation:
   https://developer.nvidia.com/blog/accelerating-digital-pathology-pipelines-with-nvidia-clara-deploy

Need to move the implementation to cuCIM's package: https://github.com/Project-MONAI/MONAI/pull/2666

@drbeh helped push a draft PR
- https://github.com/rapidsai/cucim/pull/186

We would like to update and merge it.

**Describe alternatives you've considered**
N/A

**Additional context**

We also have C++ implementation.
Once Operator interface is available, we can provide the operation with C++ implementation.

drbeh:
Fix #96 

This PR adds an stain extractor and normalizer for Hematoxylin and Eosin implemented in CuPy

### DIFF
--- a/python/cucim/src/cucim/core/operations/color/stain_normalizer.py
+++ b/python/cucim/src/cucim/core/operations/color/stain_normalizer.py
@@ -1,0 +1,223 @@
+# Copyright (c) 2021, NVIDIA CORPORATION.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from typing import Union
+
+import cupy as cp
+
+
+class HEStainExtractor:
+    """Extract stain coefficients from an image.
+
+    Parameters
+    ----------
+        source_intensity: transmitted light intensity.
+            Defaults to 240.
+        alpha: percentiles to ignore for outliers, so to calculate min and max,
+            if only consider (alpha, 100-alpha) percentiles. Defaults to 1.
+        beta: absorbance threshold for transparent pixels.
+            Defaults to 0.15
+
+    Note:
+        Please refer to this paper for further information on the method:
+        http://wwwx.cs.unc.edu/~mn/sites/default/files/macenko2009.pdf
+    """
+
+    def __init__(
+        self,
+        source_intensity: float = 240,
+        alpha: float = 1,
+        beta: float = 0.15,
+    ) -> None:
+        self.source_intensity = source_intensity
+        self.alpha = alpha
+        self.beta = beta
+
+    def _stain_decomposition(self, image: cp.ndarray) -> cp.ndarray:
+        """Extract the matrix of stain coefficient fromthe image.
+
+        Parameters
+        ----------
+            image: RGB image to perform stain extraction on
+
+        Returns
+        -------
+            stain_coeff: stain attenuation coefficient matrix derive from the
+                image, where first column is H, second column is E, and
+                rows are RGB value
+
+        """
+        # calculate absorbance
+        image = image.astype(cp.float32, copy=False) + 1.0
+        absorbance = -cp.log(
+            image.clip(max=self.source_intensity) / self.source_intensity
+        )
+
+        # reshape to form a CxN matrix
+        c = absorbance.shape[0]
+        absorbance = absorbance.reshape((c, -1))
+
+        # remove transparent pixels
+        absorbance = absorbance[cp.all(absorbance > self.beta, axis=1)]
+        if len(absorbance) == 0:
+            raise ValueError(
+                "All pixels of the input image are below the threshold."
+            )
+
+        # compute eigenvectors
+        _, eigvecs = cp.linalg.eigh(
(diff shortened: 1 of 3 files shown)


GPUtester:

quasiben:
add to allowlist

quasiben:
Thanks for adding this @drbeh  .  However, I'm trying to find if there is an equivalent of this in scikit-image, do you know if there is ?

drbeh:
Hi @quasiben, I am not aware of an equivalent to this stain normalizer. The most relevant thin on scikit-image that I know is [`separate_stain`](https://github.com/scikit-image/scikit-image/blob/e37fc660aeb96d53db27c9e2f1a0a88920bc2b65/skimage/color/colorconv.py#L1507-L1580), which is related to stain extractor part of this PR but as far as I know it is based on different methods.

It might be useful to see how we can connect these two and maybe leverage skimage for improving our stain normalizer.

grlee77:
There has been a recent, related discussion on just having a normalization to Hematoxylin & Eosin in **scikit-image** at:
https://github.com/scikit-image/scikit-image/discussions/6065#discussioncomment-2278816

drbeh:
@gigony @quasiben taking a closer look at scikit-image, I realized that although the functionality of this PR is very related to stain separation in skimage, it is orthogonal to that.

- In skimage, @grlee77 please correct me if I'm wrong, the current functionality is based on a given a Stain Matrix (`hed_from_rgb`), it separates the stains in an RGB image using `skimage.color.rgb2hed`.
- However, here we are first extract the Stain Matrix from the image (`HEStainExtractor`), and then use this matrix to normalize the RGB image (`StainNormalizer`).

grlee77:
Yes, in scikit-image there is a [fixed matrix that gets applied](https://github.com/scikit-image/scikit-image/blob/5678c72c82e427f613883a3dd1cd5f2f8fa08151/skimage/color/colorconv.py#L541-L550) to do the stain normalization. The user could also apply their own matrix or use one of the other predefined ones. My understand is that the same values are not ideal across different labs/experiments and need to be adjusted to some extent? (my research background is in medical imaging / radiology as opposed to microscopy)

grlee77:
Ah, I see there is a reference to a Macenko et. al. paper in the docstrings. I will take a look at that to understand the method proposed here a bit better.

python/cucim/src/cucim/core/operations/color/stain_normalizer.py
```diff
@@ -0,0 +1,223 @@
...
+            image.clip(max=self.source_intensity) / self.source_intensity
+        )
+
+        # reshape to form a CxN matrix
+        c = absorbance.shape[0]
+        absorbance = absorbance.reshape((c, -1))
+
+        # remove transparent pixels
+        absorbance = absorbance[cp.all(absorbance > self.beta, axis=1)]
+        if len(absorbance) == 0:
```
grlee77, line 73:
```suggestion
        absorbance = absorbance[:, cp.all(absorbance > self.beta, axis=0)]
        if absorbance.size == 0:
```

grlee77: based on the prior line, it looks like this method expects an image with channels along the first axis. 

I have been refactoring things while testing locally. I think we can support specification of a `channel_axis` argument to make it clear which axis of the input is expected to be channels.

grlee77: It is not clear from the Marcenko et. al. publication whether this should ideally be `all(absorbance > self.beta)` as here or if some other operation like the mean or max absorbance would be used.

python/cucim/src/cucim/core/operations/color/stain_normalizer.py
```diff
@@ -0,0 +1,223 @@
...
+        Returns
+        -------
+            stain_coeff: stain attenuation coefficient matrix derive from the
+                image, where first column is H, second column is E, and
+                rows are RGB value
+
+        """
+        # calculate absorbance
+        image = image.astype(cp.float32, copy=False) + 1.0
+        absorbance = -cp.log(
```
grlee77, line 63: Marcenko et. al. uses `log10` rather than `log` in the publication, but existing open source implementations I found all use the natural log. I guess either is valid and it only effects the absolute scale of the `beta` threshold. Empirically, computation of `log` is better optimized than `log10`, so we should probably keep this as is.

grlee77:
I agree that this Marcenko method is highly cited an seems like a good starting point. I see that in determining the stain vectors, least squares is used here, which is readily available in CuPy. 

Other, related techniques use different solutions, but those can potentially be added later.

For example, there is a technique based on non-negative matrix factorization here:
https://github.com/InsightSoftwareConsortium/ITKColorNormalization

The Marcenko method is also implemented (under an MIT license) here: https://github.com/Peter554/StainTools
It differs from the one in this PR in that the normalization [uses LASSO (via the SPAMS library) instead of least squares](https://github.com/Peter554/StainTools/blob/master/staintools/utils/get_concentrations.py#L16). I don't think we already have an equivalent solver on the GPU (at least it is not in CuPy or cuCIM, but I haven't fully checked other RAPIDS projects)

I will update this PR with specific suggestions later today after testing a bit more with some actual images (such as the ones used for the demo in the ITKColorNormalization repo).

python/cucim/src/cucim/core/operations/color/stain_normalizer.py
drbeh: You are right! It only affect `beta` and as long as we are using the correct inverse (`exp` for `log` and `10**` for `log10`) both are fine.

drbeh:
Closing this in favor of #273

Issue #149, opened 2021-11-16
[Design] Supporting Multithreading and Batch Processing


gigony:
### Issue

cuCIM's read_region doesn't use multithreading and whether if using them is delegated to the user.
And, to get better performance using nvJPEG/GDS, batch-processing is needed.

The following requirements are needed:

- Support multithreading
  - Use multi-threading to load the entire image.
- Support batch processing
  - Batch-loading of patches from a huge image for AI model training/inferencing use cases.
  - Integration with nvJPEG&GDS:
    - Design and Implement image patch generator by loading raw compressed images at once using GDS and providing a series of image patches by decoding multiple tile images in batch using nvJPEG

Note that batch processing would be supported only for a single (large) image with the list of start locations of the patch, not for a series of images.

Supporting this feature would help AI model training/inferencing workflows and integration with [cuNumeric](https://developer.nvidia.com/cunumeric) (e.g., loading a whole slide image into a single NumPy Array on GPU) in the future.

Related issues:

- [[FEA] Multi-threading to load the entire image #139](https://github.com/rapidsai/cucim/issues/139)
- [[FEA] Integration between GDS & nvJPEG #123](https://github.com/rapidsai/cucim/issues/123)

#### Example API Usages

The following parameters would be added in the `read_region` method:

- `num_workers`: number of workers(threads) to use for loading the image. (default: `1`)
- `batch_size`: number of images to load at once. (default: `1`)
- `drop_last`: whether to drop the last batch if the batch size is not divisible by the number of images. (default: `False`)
- `preferch_factor`: number of samples loaded in advance by each worker. (default: `2`)
- `shuffle`: whether to shuffle the input locations (default: `False`)
- `seed`: seed value for random value generation (default: 0)

**Loading entire image by using multithreads**

```python
from cucim import CuImage

img = CuImage("input.tif")

region = img.read_region(level=1, num_workers=8)  # read whole image at level 1 using 8 workers
```

**Loading batched image using multithreads**

You can feed locations of the region through the list/tuple of locations or through the NumPy array of locations.
(e.g., `((<x for loc 1>, <y for loc 1>), (<x for loc 2>, <y for loc 2>)])`).
Each element in the location should be int type (int64) and the dimension of the location should be
equal to the dimension of the size.
You can feed any iterator of locations (dimensions of the input don't matter, flattening the item in the iterator once if the item is also an iterator).

For example, you can feed the following iterator:

[...]

Issue #139, opened 2021-11-03
[FEA] Multi-threading to load the entire image


drbeh:
**Is your feature request related to a problem? Please describe.**
There are use cases in digital pathology that we want to load the entire image (WSI at a certain level) to the memory. Doing so with `cuCIM` seems to be much slower than other libraries, like `tifffile`, which uses multi-threading to load the entire image. Thus it would be very helpful is `cuCIM` can also support multi-threading and be able to perform on par with other libraries to load the entire image into memory.

Related issue: 
- https://github.com/Project-MONAI/MONAI/issues/3161
- https://github.com/Project-MONAI/MONAI/issues/3251

gigony:
Loading the entire image is supported from cuCIM v21.12.01

https://github.com/rapidsai/cucim/wiki/release_notes_v21.12.01

drbeh:
That is fantastic, @gigony! Thanks!

Issue #123, opened 2021-10-07
[FEA] Integration between GDS & nvJPEG


gigony:
Design and Implement image patch generator by loading raw compressed image at once using GDS and provide a series of image patches by decoding multiple tile images in batch using nvJPEG

https://nbviewer.org/github/gigony/cucim/blob/v21.06.00/notebooks/Using_Cache.ipynb#Supporting-Generator-(iterator)

gigony:
We will follow up on the remaining issue with https://github.com/rapidsai/cucim/issues/222.
Closing this.

PR #191, opened 2022-01-10
Support multi-threads and batch, and support nvJPEG for JPEG-compressed images

gigony:
This patch is for supporting/addressing the following issues.
- #139
- #123
- #149

Please see #149 for the detailed design.

- Use [TaskFlow](https://github.com/taskflow/taskflow) to use a thread pool and distributes workload to multiple threads if `num_workers` parameter to `read_region()` API is larger than zero.
- Implement a buffered loader to load images in a batch manner.
- Uses nvJPEG to decode JPEG-compressed image if `device` parameter to `read_region()` API is `cuda`. (It requires performance improvement though).
- File handler is now shared among multiple CuImage objects.
- Clean up some code

This implementation is already released as part of cuCIM [v21.12.01](https://github.com/rapidsai/cucim/wiki/release_notes_v21.12.01) release.
The followings are added on top of existing changes:
- Use nvjpeg.a from the publicly available package.
- Fix GPU memory leak when using nvjpeg API (when `device='cuda'` parameter is used in `read_region` method).
- Get `libculibos.a` from /usr/local/lib[64] if the library cannot be found in Conda environment. (nvjpeg depends on the static library).


cpp/cmake/deps/concurrentqueue.cmake
```diff
@@ -0,0 +1,36 @@
+#
```
grlee77, line 1: Is `concurrentqueue` being used (I didn't see where it is)? If so, I think we need a `LICENSE.concurrentqueue`. If not, then we can remove this cmake file.

gigony: Thanks @grlee77 ! `concurrentqueue` is not used. let me remove the cmake file.

ajschmidt8 (approved):
Approving `ops-codeowner` file changes

cpp/cmake/deps/concurrentqueue.cmake
gigony: Removed it in the latest patchset.

cpp/plugins/cucim.kit.cuslide/src/cuslide/tiff/ifd.cpp
```diff
@@ -48,17 +52,16 @@ IFD::IFD(TIFF* tiff, uint16_t index, ifd_offset_t offset) : tiff_(tiff), ifd_ind
...
     PROF_SCOPED_RANGE(PROF_EVENT(ifd_ifd));
     auto tif = tiff->client();
-    int ret;
 
     char* software_char_ptr = nullptr;
     char* model_char_ptr = nullptr;
-    // TODO: error handling
 
-    ret = TIFFGetField(tif, TIFFTAG_SOFTWARE, &software_char_ptr);
+    // TODO: error handling
```
jakirkham, line 59: Would be good to convert this to an issue for easier tracking

cpp/src/cache/image_cache_per_process.cpp
```diff
@@ -53,19 +55,38 @@ struct PerProcessImageCacheItem
...
+        case io::DeviceType::kCUDA: {
+            cudaError_t cuda_status;
+            CUDA_TRY(cudaFree(data));
+            break;
+        }
+        case io::DeviceType::kPinned:
+        case io::DeviceType::kCPUShared:
+        case io::DeviceType::kCUDAShared:
+            fmt::print(stderr, "Device type {} is not supported!\n", device_type);
+            break;
```
jakirkham, line 82: Do we intend to cover these in the future? If so, should we raise an issue?

Note: This comes up again below.

cpp/src/cuimage.cpp
```diff
@@ -143,21 +144,25 @@ std::unique_ptr<plugin::ImageFormat> CuImage::image_format_plugins_ = std::make_
...
+
     ensure_init();
     image_format_ = image_format_plugins_->detect_image_format(path);
 
     // TODO: need to detect available format for the file path
     {
         PROF_SCOPED_RANGE(PROF_EVENT(cuimage_cuimage_open));
-        file_handle_ = image_format_->image_parser.open(path.c_str());
+        std::shared_ptr<CuCIMFileHandle>* file_handle_shared =
+            reinterpret_cast<std::shared_ptr<CuCIMFileHandle>*>(image_format_->image_parser.open(path.c_str()));
```
jakirkham, line 155: Curious why this is `shared_ptr<...>*`?

cpp/src/cuimage.cpp
```diff
@@ -273,18 +266,18 @@ CuImage::~CuImage()
...
+                    cudaError_t cuda_status;
+                    CUDA_TRY(cudaFree(image_data_->container.data));
                 }
+                image_data_->container.data = nullptr;
                 break;
             case io::DeviceType::kPinned:
             case io::DeviceType::kCPUShared:
             case io::DeviceType::kCUDAShared:
-                fmt::print(stderr, "Device type {} is not supported!", device_type);
+                fmt::print(stderr, "Device type {} is not supported!\n", device_type);
```
jakirkham, line 280: JFYI there is an issue about this upstream ( https://github.com/fmtlib/fmt/issues/428 )

jakirkham (commented):
Thanks Gigon! 😄

Had a few comments below.

python/pybind11/cucim_py.cpp
```diff
@@ -427,14 +506,26 @@ py::object py_read_region(const CuImage& cuimg,
...
+        return iter;
+    }
+    else
     {
         py::gil_scoped_acquire scope_guard;
 
         py::object region = py::cast(region_ptr);
 
-        // Add `__array_interace__` or `__cuda_array_interface__` in runtime.
+        // Add `__array_inteface__` or `__cuda_array_interface__` in runtime.
```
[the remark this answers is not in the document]
gigony: Thank you. Will update.

python/pybind11/cache/cache_py.h
```diff
@@ -40,7 +40,7 @@ py::dict py_config(ImageCache& cache);
 void py_image_cache_reserve(ImageCache& cache, uint32_t memory_capacity, py::kwargs kwargs);
 
 py::int_ py_preferred_memory_capacity(const py::object& img,
-                                      const std::optional<const std::vector<uint32_t>>& image_size,
+                                      const std::optional<const std::vector<uint64_t>>& image_size,
```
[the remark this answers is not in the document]
gigony: using `uint32_t` was just a naive thought that input image size (width or height) wouldn't be higher than 2,000,000,000px but it can cause problems in the distant future so I made sure it uses 64-bit type for image size.

cpp/plugins/cucim.kit.cuslide/cmake/deps/libopenjpeg.cmake
```diff
@@ -38,7 +38,6 @@ if (NOT TARGET deps::libopenjpeg)
     ###########################################################################
     # Build liblcms2 with the source in libopenjpeg
     ###########################################################################
-    message(" ##GG ${deps-libopenjpeg_SOURCE_DIR}")
```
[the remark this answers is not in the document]
gigony: The message was added during debugging (to see which path is used) so removed here.

Since `deps-libopenjpeg_SOURCE_DIR` would be populated by `FetchContent_GetProperties()` above, we don't need to check its existence.

cpp/plugins/cucim.kit.cuslide/src/cuslide/cuslide.cpp
```diff
@@ -82,15 +82,20 @@ static bool CUCIM_ABI checker_is_valid(const char* file_name, const char* buf, s
...
 
-static CuCIMFileHandle CUCIM_ABI parser_open(const char* file_path)
+static CuCIMFileHandle_share CUCIM_ABI parser_open(const char* file_path)
 {
     auto tif = new cuslide::tiff::TIFF(file_path, O_RDONLY);
     tif->construct_ifds();
-    return tif->file_handle();
+    // Move the ownership of the file handle object to the caller (CuImage).
+    auto handle_t = tif->file_handle();
+    tif->file_handle() = nullptr;
```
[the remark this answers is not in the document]
gigony: The method (`tif->file_handle()`) was used for holding shared_ptr object temporary.

Plugins(libcuslide.so) are using C-API and they open a file descriptor.
CuImage objects in main library (libcucim.so) may share the file descriptor so needed to handle by using shared_ptr.

CuImage class wants to use the shared file descriptor (std::shared_ptr<CuCIMFileHandle>) created from the plugin code so that CuImage object can close the file descriptor when no objects are using the file descriptor.

```
    // Move the ownership of the file handle object to the caller (CuImage).
    auto handle_t = tif->file_handle();    // (1)
    tif->file_handle() = nullptr;               // (2)
    CuCIMFileHandle_share handle = new std::shared_ptr<CuCIMFileHandle>(handle_t);   // (3)
    return handle;                                                                                                                  // (4)
```
(1) Assigning to handle_t would increase the reference count of the shared object (`CuCIMFileHandle` object). Now `CuCIMFileHandle` object has a reference count of 2.
(2) By assigning `nullptr` to tif->file_handle(), plugin code lose the access to the shared pointer so the reference count is 1 (owned by only `handle_t` local variable).
(3) Create a `pointer` whose type is `std::shared_ptr<CuCIMFileHandle>*` (dynamic allocation) and 
(4) return the handle

The returned pointer is restored to `std::shared_ptr<CuCIMFileHandle>` from CuImage class below:

https://github.com/gigony/cucim/blob/622f5bb579d50db41973bad84c435d2968364ca4/cpp/src/cuimage.cpp#L154
```
        std::shared_ptr<CuCIMFileHandle>* file_handle_shared =
            reinterpret_cast<std::shared_ptr<CuCIMFileHandle>*>(image_format_->image_parser.open(path.c_str()));
        file_handle_ = *file_handle_shared;
        delete file_handle_shared;
```

This answers a question here: https://github.com/rapidsai/cucim/pull/191#discussion_r792436016

Moving shared_ptr between C API and C++ API has a kind of boilerplate code and it can be simplified by using a template though. (Please teach me if you know a better/fancy way to solve the issue).

The below link is also talking about the same issue:
https://www.reddit.com/r/cpp_questions/comments/d2owlo/how_can_i_pass_shared_ownership_through_a_void/


https://github.com/gigony/cucim/blob/622f5bb579d50db41973bad84c435d2968364ca4/cpp/plugins/cucim.kit.cuslide/src/cuslide/tiff/tiff.h#L71
```
    std::shared_ptr<CuCIMFileHandle>& file_handle(); /// used for moving the ownership of the file handle to the caller.
                                                     /// Do not use for the application -- it will return nullptr.
```

cpp/src/cuimage.cpp
gigony: Please see the answer https://github.com/rapidsai/cucim/pull/191#discussion_r793135703

cpp/plugins/cucim.kit.cuslide/src/cuslide/loader/nvjpeg_processor.cpp
```diff
@@ -0,0 +1,438 @@
...
+    stopped_ = true;
+    cuda_batch_cond_.notify_all();
+}
+
+uint32_t NvJpegProcessor::preferred_loader_prefetch_factor()
+{
+    return preferred_loader_prefetch_factor_;
+}
+
+void NvJpegProcessor::update_file_block_info(int64_t* request_location, int64_t* request_size, uint64_t location_len)
```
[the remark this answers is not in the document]
gigony: Thanks! Updated.

cpp/plugins/cucim.kit.cuslide/src/cuslide/jpeg/libnvjpeg.h
```diff
@@ -0,0 +1,36 @@
...
+#ifndef CUSLIDE_LIBNVJPEG_H
+#define CUSLIDE_LIBNVJPEG_H
+
+#include <cucim/io/device.h>
+
+namespace cuslide::jpeg
+{
+
+EXPORT_VISIBLE bool decode_libnvjpeg(int fd,
+                                     unsigned char* jpeg_buf,
```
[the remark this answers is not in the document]
gigony: Updated.

cpp/plugins/cucim.kit.cuslide/src/cuslide/tiff/ifd.cpp
gigony: Good idea. Created https://github.com/rapidsai/cucim/issues/209

cpp/plugins/cucim.kit.cuslide/src/cuslide/loader/nvjpeg_processor.cpp
```diff
@@ -0,0 +1,438 @@
...
+
+    cudaError_t cuda_status;
+
+    // Initialize batch data with the first data
+    if (raw_cuda_inputs_.empty())
+    {
+        for (uint32_t i = 0; i < cuda_batch_size_; ++i)
+        {
+            uint8_t* mem_offset = nullptr;
+            mem_offset = file_block_ptr + tile_to_request[0].offset - file_start_offset_;
```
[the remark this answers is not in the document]
gigony: Oh, yes. Should be one line.

cpp/src/cache/image_cache_per_process.cpp
gigony: Created issue:
- https://github.com/rapidsai/cucim/issues/210

cpp/src/cuimage.cpp
gigony: Oh, that's good to know! Thanks for the pointer!

jakirkham (approved):

jakirkham:
@gpucibot merge

PR #197, opened 2022-01-19
Allow installation with CuPy 10

grlee77:
Now that CuPy 10.0 is out, we should also allow cuCIM to be installed with CuPy 10.x in the next release.

I checked the integration repo and [it has the same constraint proposed here](https://github.com/rapidsai/integration/blob/a76cfc32e684f53cf0a9b4cb113cf0fd9fad51e7/conda/recipes/versions.yaml#L47-L48)

### DIFF
--- a/conda/environments/env.yml
+++ b/conda/environments/env.yml
@@ -2,7 +2,7 @@ name: cucim
 channels:
   - conda-forge
 dependencies:
-  - cupy>=9,<11.0.0a0
+  - cupy>=9
   - scikit-image=0.18.1
   - openslide
   - zlib
--- a/conda/environments/env.yml
+++ b/conda/environments/env.yml
@@ -2,7 +2,7 @@ name: cucim
 channels:
   - conda-forge
 dependencies:
-  - cupy=9
+  - cupy>=9,<11.0.0a0
   - scikit-image=0.18.1
   - openslide
   - zlib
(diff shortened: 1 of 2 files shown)


grlee77:
okay, I guess `/conda/recipes/cucim/meta.yaml` had already been updated, so this is now just an update of `conda/environments/env.yml`

conda/environments/env.yml
```diff
@@ -2,7 +2,7 @@ name: cucim
 channels:
   - conda-forge
 dependencies:
-  - cupy=9
+  - cupy>=9,<11.0.0a0
```
jakirkham, line 5: Perhaps we should just drop this constraint? I think the original motivation was to ensure we got CuPy 9 as we needed the very latest features. Do things work without a constraint here?

grlee77: I am fine with just keeping the lower bound. I think it should tend to find 10.0 without the constraint, but it doesn't hurt to indicate that it is incompatible with 8.x.

grlee77, line 5:
```suggestion
  - cupy>=9
```

jakirkham (approved):

ajschmidt8 (approved):

jakirkham:
@gpucibot merge

PR #205, opened 2022-01-20
Do not use cudatoolkit from conda for building libcucim

gigony:
When `find_package(CUDAToolkit REQUIRED)` is called, CUDA library path is determined by the location of `libcudart.so`.

- https://gitlab.kitware.com/cmake/cmake/-/blob/v3.22.1/Modules/FindCUDAToolkit.cmake#L922
- https://gitlab.kitware.com/cmake/cmake/-/blob/v3.22.1/Modules/FindCUDAToolkit.cmake#L848
- https://gitlab.kitware.com/cmake/cmake/-/blob/v3.22.1/Modules/FindCUDAToolkit.cmake#L834
- https://gitlab.kitware.com/cmake/cmake/-/blob/v3.22.1/Modules/FindCUDAToolkit.cmake#L793

For https://github.com/rapidsai/cucim/pull/191 to be built, it needs `CUDA::culibos` (libculibos.a) which is a static library that will be needed to link a custom nvjpeg static library(libnvjpeg.a).

If `cudatoolkit` is set to `host` of `libcucim`'s conda recipe, it cannot find `CUDA::culibos` component because conda's `cudatoolkit` (runtime libraries) takes precedence.

This patch removes `cudatoolkit` dependency for libcucim recipe so it can build the library without error.

### DIFF
--- a/conda/recipes/libcucim/meta.yaml
+++ b/conda/recipes/libcucim/meta.yaml
@@ -29,7 +29,6 @@ requirements:
     - sysroot_linux-64 2.17
     - yasm # [x86_64]
   host:
-    - cudatoolkit {{ cuda_version }}.*
     - openslide
     - zlib
     - jpeg


gigony:
Necessary change is merged in #191

PR #208, opened 2022-01-26
Apply fixes to skimage.transform scheduled for scikit-image 0.19.2

### LINKED ISSUE #199: [BUG] PiecewiseAffineTransform can trigger TypeError
**Describe the bug**

With certain inputs, [PiecewiseAffineTransform](https://github.com/rapidsai/cucim/blob/d6d3af5695451971167f64fe30174ab1a231a6b1/python/cucim/src/cucim/skimage/transform/_geometric.py#L987) will trigger a code path that raises

```
  File "/home/croat/.local/share/virtualenvs/starmap-T47byR32/lib/python3.8/site-packages/cucim/skimage/transform/_geometric.py", line 1092, in __call__
    out[index_mask, :] = affine(coords[index_mask, :])
  File "/home/croat/.local/share/virtualenvs/starmap-T47byR32/lib/python3.8/site-packages/cucim/skimage/transform/_geometric.py", line 651, in __call__
    return self._apply_mat(coords, self.params)
  File "/home/croat/.local/share/virtualenvs/starmap-T47byR32/lib/python3.8/site-packages/cucim/skimage/transform/_geometric.py", line 612, in _apply_mat
    dst = src @ matrix.T
  File "cupy/_core/core.pyx", line 1574, in cupy._core.core.ndarray.__array_ufunc__
  File "/home/croat/.local/share/virtualenvs/starmap-T47byR32/lib/python3.8/site-packages/cupy/linalg/_product.py", line 42, in matmul
    return _gu_func_matmul(x1, x2, out=out, axes=axes)
  File "/home/croat/.local/share/virtualenvs/starmap-T47byR32/lib/python3.8/site-packages/cupy/_core/_gufuncs.py", line 652, in __call__
    args, dimsizess, loop_output_dims, outs, m_dims = self._get_args_transposed(  # NOQA
  File "/home/croat/.local/share/virtualenvs/starmap-T47byR32/lib/python3.8/site-packages/cupy/_core/_gufuncs.py", line 466, in _get_args_transposed
    args = [_manipulation.broadcast_to(
  File "/home/croat/.local/share/virtualenvs/starmap-T47byR32/lib/python3.8/site-packages/cupy/_core/_gufuncs.py", line 466, in <listcomp>
    args = [_manipulation.broadcast_to(
TypeError: Argument 'array' has incorrect type (expected cupy._core.core.ndarray, got numpy.ndarray)
```

```
import cupy as cp
from cucim.skimage.transform import PiecewiseAffineTransform


SRC = cp.asarray(
    [
        [0, 192, 256],
        [0, 256, 256],
        [5, 0, 192],
        [5, 64, 0],
        [5, 64, 64],
        [5, 64, 256],
        [5, 192, 192],
        [5, 256, 256],
        [0, 192, 256],
    ]
)

DST = cp.asarray(
    [
        [0, 142, 206],
        [0, 206, 206],
        [5, -50, 142],
        [5, 14, 0],
        [5, 14, 64],
        [5, 14, 206],
        [5, 142, 142],
        [5, 206, 206],
        [0, 142, 206],
    ]
)

tform = PiecewiseAffineTransform()
tform.estimate(SRC, DST)
tform(SRC)
```

**Expected behavior**

The above code should run without error.

 - Environment location: Bare metal
 - Method of cuCIM install: pip (cucim==21.12.1)
 - Run on Ubuntu 20.04 with an NVIDIA TITAN RTX
 
**Additional context**

Not every input causes the problem.  For example, removing a pair of points from SRC/DST in the above example does not trigger the error.

grlee77:
Closes #199

Corresponds to:
scikit-image/scikit-image#6207
scikit-image/scikit-image#6211
scikit-image/scikit-image#6214

Additionally, the affines contained within the `PiecewiseAffineTransform` will have parameters in CuPy arrays when the inputs to `estimate` are CuPy arrays. This is one to make it consistent with the other transform classes.

@chrisroat, can you confirm if this fixes the issue for you?

### DIFF
--- a/python/cucim/src/cucim/skimage/transform/_geometric.py
+++ b/python/cucim/src/cucim/skimage/transform/_geometric.py
@@ -1051,7 +1051,7 @@ class PiecewiseAffineTransform(GeometricTransform):
         Returns
         -------
         success : bool
-            True, if model estimation succeeds.
+            True, if all pieces of the model are successfully estimated.
 
         """
 
@@ -1059,31 +1059,30 @@ class PiecewiseAffineTransform(GeometricTransform):
         # forward piecewise affine
         # triangulate input positions into mesh
         xp = cp.get_array_module(src)
-        if xp is cp:
-            # TODO: grlee77 :update if spatial.Delaunay is implemented for GPU
-            # transfer to CPU for use of spatial.Delaunay
-            src = cp.asnumpy(src)
-            dst = cp.asnumpy(dst)
+        # TODO: update if spatial.Delaunay become available in CuPy
+        self._tesselation = spatial.Delaunay(cp.asnumpy(src))
 
-        self._tesselation = spatial.Delaunay(src)
+        ok = True
+
         # find affine mapping from source positions to destination
         self.affines = []
-        for tri in self._tesselation.vertices:
+        for tri in xp.asarray(self._tesselation.vertices):
             affine = AffineTransform(dimensionality=ndim)
-            affine.estimate(src[tri, :], dst[tri, :])
+            ok &= affine.estimate(src[tri, :], dst[tri, :])
             self.affines.append(affine)
 
         # inverse piecewise affine
         # triangulate input positions into mesh
-        self._inverse_tesselation = spatial.Delaunay(dst)
+        # TODO: update if spatial.Delaunay become available in CuPy
+        self._inverse_tesselation = spatial.Delaunay(cp.asnumpy(dst))
         # find affine mapping from source positions to destination
         self.inverse_affines = []
-        for tri in self._inverse_tesselation.vertices:
+        for tri in xp.asarray(self._inverse_tesselation.vertices):
             affine = AffineTransform(dimensionality=ndim)
-            affine.estimate(dst[tri, :], src[tri, :])
+            ok &= affine.estimate(dst[tri, :], src[tri, :])
             self.inverse_affines.append(affine)
 
-        return True
+        return ok
 
     def __call__(self, coords):
         """Apply forward transformation.
@@ -1103,12 +1102,12 @@ class PiecewiseAffineTransform(GeometricTransform):
         """
 
         xp = cp.get_array_module(coords)
-        if xp == cp:
-            coords = coords.get()
-        out = np.empty_like(coords, np.double)
+        out = xp.empty_like(coords, xp.double)
 
         # determine triangle index for each coordinate
-        simplex = self._tesselation.find_simplex(coords)
+        # coords must be on host for calls to _tesselation methods
+        simplex = self._tesselation.find_simplex(cp.asnumpy(coords))
+        simplex = xp.asarray(simplex)
 
         # coordinates outside of mesh
         out[simplex == -1, :] = -1
@@ -1121,8 +1120,6 @@ class PiecewiseAffineTransform(GeometricTransform):
 
             out[index_mask, :] = affine(coords[index_mask, :])
 
-        if xp == cp:
-            out = xp.asarray(out)
         return out
 
     def inverse(self, coords):
@@ -1143,12 +1140,12 @@ class PiecewiseAffineTransform(GeometricTransform):
         """
 
         xp = cp.get_array_module(coords)
-        if xp == cp:
-            coords = coords.get()
-        out = np.empty_like(coords, np.double)
+        out = xp.empty_like(coords, xp.double)
 
         # determine triangle index for each coordinate
-        simplex = self._inverse_tesselation.find_simplex(coords)
+        # coords must be on host for calls to _tesselation methods
+        simplex = self._inverse_tesselation.find_simplex(cp.asnumpy(coords))
+        simplex = xp.asarray(simplex)
 
         # coordinates outside of mesh
         out[simplex == -1, :] = -1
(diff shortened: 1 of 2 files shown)


grlee77:
@chrisroat, Can you provide some context of the sizes of the `src` and `dst` arrays that are being used with the `estimate` method in your application? I did not spend much time thinking about this `_geometry.py` module on the cuCIM side and it is probably one area where we are unlikely to get a benefit over staying on the CPU given the small sizes of the affine matrices. `PiecewiseAffineTransform` has a particular difficulty in that it relies on methods from `scipy.spatial.Delaunay` that do not have a corresponding GPU implementation in CuPy, requiring us to transfer to/from the host in multiple places! 

For that reason this `_geometry.py` is possibly the only place in cuCIM where the classes accept both NumPy or CuPy arrays as inputs. Most of the rest of the library only allows CuPy arrays. This is probably something we should discuss further to decide on the desired behavior long term.

chrisroat:
Thanks for consolidating everything, and adding the conversion to cupy arrays.  My incremental approach was too slow.  :)

The application in question is warping a chunk of data to a new coordinate system.  The code basically looks like the following. The `shape` is as large as 128x512x512, which gives `coords` roughly 100M elements.

```
    ndim = chunk.ndim
    shape = chunk.shape

    coords = np.indices(shape, np.int32)
    coords = coords.reshape(ndim, -1).T
    coords = tform(coords)
    coords = coords.T.reshape((-1,) + shape)
    ...
    return ndimage.map_coordinates(chunk, coords, ...)
```

chrisroat:
Oh, sorry -- I realize now that you were asking about the `estimate` function, the `__call__`.  Yes, the arrays are small that calculate the transform -- typically less than 1k elements.  I could use CPU skimage to calculate the transform, and then switch over to cucium based affines for the use of the transform and the warp.

python/cucim/src/cucim/skimage/transform/_geometric.py
```diff
@@ -1050,39 +1051,38 @@ def estimate(self, src, dst):
...
         # forward piecewise affine
         # triangulate input positions into mesh
         xp = cp.get_array_module(src)
-        if xp is cp:
-            # TODO: grlee77 :update if spatial.Delaunay is implemented for GPU
-            # transfer to CPU for use of spatial.Delaunay
-            src = cp.asnumpy(src)
-            dst = cp.asnumpy(dst)
+        # TODO: update if spatial.Delaunay become available in CuPy
+        self._tesselation = spatial.Delaunay(cp.asnumpy(src))
```
jakirkham, line 1063: Should we open an issue about this?

python/cucim/src/cucim/skimage/transform/_geometric.py
```diff
@@ -1102,12 +1102,12 @@ def __call__(self, coords):
...
         xp = cp.get_array_module(coords)
-        if xp == cp:
-            coords = coords.get()
-        out = np.empty_like(coords, np.double)
+        out = xp.empty_like(coords, xp.double)
 
         # determine triangle index for each coordinate
-        simplex = self._tesselation.find_simplex(coords)
+        # coords must be on host for calls to _tesselation methods
+        simplex = self._tesselation.find_simplex(cp.asnumpy(coords))
```
jakirkham, line 1109: Why do we coerce to a NumPy array here?

Note: This comes up again below, but only commenting here for simplicity.

jakirkham (commented):
Thanks Greg! 😄

Had a couple questions below

python/cucim/src/cucim/skimage/transform/_geometric.py
grlee77: We could, it is probably not very easy to address though.  We would need a Delaunay triangulation class with a `vertices` attribute and `find_simplex` method for the GPU (There are several other attributes/methods on this class in SciPy, but they are not used here). On the SciPy side the implementation relies on the QHull C library.

grlee77: If we did find one, ideally it would be available via upstream `cupyx.scipy.spatial`

grlee77:
> Oh, sorry -- I realize now that you were asking about the estimate function, not the __call__. Yes, the arrays are small that calculate the transform -- typically less than 1k elements. I could use CPU skimage to calculate the transform, and then switch over to cucium based affines for the use of the transform and the warp.

Yeah, that is probably the best approach here.

python/cucim/src/cucim/skimage/transform/_geometric.py
jakirkham: Idk maybe it could happen. There has been work in the past like [CudaHull]( https://doi.org/10.1016/j.cag.2012.02.012 )

jakirkham:
So RAPIDS is [releasing next week]( https://docs.rapids.ai/maintainers ). Should we retarget this for the next release?

jakirkham:
Retargeted to `branch-22.04` per our discussion in the meeting today. Though ok to include PyPI 22.02.01

jakirkham (approved):

jakirkham:
@gpucibot merge

jakirkham:
Thanks Greg! 😄

Let's follow up on anything else in a new PR 🙂

Issue #225, opened 2022-02-22
[BUG] Transforms on cucim.clara use a fixed probability for the images in a batch


gigony:
**Describe the bug**
Transforms introduced by https://github.com/rapidsai/cucim/wiki/release_notes_v21.10.00#1-add-high-performance-transforms-for-digital-pathology-100 use a fixed probability for the images in a batch.

**Steps/Code to reproduce bug**

Use this tutorial.

https://github.com/Project-MONAI/tutorials/tree/master/pathology/tumor_detection

**Additional context**

Using fixed probability reduced the accuracy of the results.

jakirkham:
With PR ( https://github.com/rapidsai/cucim/pull/231 ) in, are we good to close or is there more needed here?

cc @shekhardw (for vis)

gigony:
I think it's good to close this. Closing.

The remaining issue can be followed up by 
- #259

PR #231, opened 2022-03-02
randomization per image per batch

shekhardw:
As per https://github.com/rapidsai/cucim/issues/225

Randomness updated to be applied to per image in the batch. Existing functionality of applying randomness on entire batch at once is unchanged.

### DIFF
--- a/python/cucim/src/cucim/core/operations/spatial/rotate_and_flip.py
+++ b/python/cucim/src/cucim/core/operations/spatial/rotate_and_flip.py
@@ -66,6 +66,7 @@ def image_flip(
         result = cupy.flip(cupy_img, spatial_axis)
         if to_cupy is True:
             result = cupy.asnumpy(result)
+
         return result
     except Exception as e:
         _logger.error("[cucim] " + str(e), exc_info=True)
@@ -130,7 +131,8 @@ def image_rotate_90(
 def rand_image_flip(
     img: Any,
     spatial_axis: tuple(),
-    prob: float = 0.1
+    prob: float = 0.1,
+    whole_batch: bool = False
 ) -> Any:
     """
     Randomly flips the image along axis.
@@ -142,6 +144,9 @@ def rand_image_flip(
     prob: Probability of flipping.
     spatial_axis : tuple
         spatial axis along which to flip over the input array
+    whole_batch: Flag to apply transform on whole batch.
+        If False, each image in the batch is randomly transformed
+        It True, entire batch is transformed randomly.
 
     Returns
     -------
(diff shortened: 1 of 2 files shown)


python/cucim/src/cucim/core/operations/color/jitter.py
```diff
@@ -298,3 +298,87 @@ def adjust_hue(input_arr, hue):
...
+def rand_color_jitter(
+    img: Any,
+    brightness=0,
+    contrast=0,
+    saturation=0,
+    hue=0,
+    prob: float = 0.1,
+    whole_batch: bool = False
+):
+    """Randomly Applies color jitter by random sequential application of
```
gigony, line 312:
```suggestion
    """Randomly applies color jitter by random sequential application of
```

python/cucim/src/cucim/core/operations/spatial/tests/test_random_flip.py
```diff
@@ -49,7 +49,10 @@ def test_rand_flip_batchinput():
     arr = get_input_arr()
     flip_arr = get_flipped_data()
     arr_batch = np.stack((arr,) * 8, axis=0)
-    np_output = spt.rand_image_flip(arr_batch, prob=1.0, spatial_axis=(2, 3))
+    np_output = spt.rand_image_flip(arr_batch,
+                                    prob=1.0,
+                                    spatial_axis=(2, 3),
```
gigony, line 54: Parameter input is not consistent. Here it uses `(2, 3)`, but below code, it uses `[2, 3]`.

python/cucim/src/cucim/core/operations/spatial/rotate_and_flip.py
```diff
@@ -130,7 +131,8 @@ def image_rotate_90(
 def rand_image_flip(
     img: Any,
     spatial_axis: tuple(),
-    prob: float = 0.1
+    prob: float = 0.1,
```
gigony, line 134: What would be the good default value for this?
Having `0.5` for the default value looks more useful here?

gigony (approved):
Looks good to me.
Have some minor suggestions.

python/cucim/src/cucim/core/operations/spatial/rotate_and_flip.py
shekhardw: Will update it.

jakirkham:
Generally seems reasonable. Do we want to test `whole_batch=False` as well?

jakirkham (approved):

jakirkham:
@gpucibot merge

GPUtester:

PR #246, opened 2022-03-25
Expose data type of CuImage object for interoperability with NumPy

### LINKED ISSUE #243: [FEA] Expose typestr to CuImage object for converting data type of the image to numpy.dtype
It would be better if there is an easier way for converting the data type of the image to NumPy's datatype.

currently, a workaround is np.dtype(img.__array_interface__["typestr"]).

Will expose typestr property to the image object so `np.dtype(img.typestr)` is possible.

**Additional context**
This feature is necessary for interoperability with other libraries/frameworks.

gigony:
- Expose typestr property on CuImage object
- Expose DLDataType and DLDataTypeCode under cucim.clara

Without this patch,
DLDataType and DLDataTypecode can be accessible with the below workaround.

```python
>>> from cucim import CuImage
>>> a = CuImage("notebooks/input/image.tif")
>>> b = a.read_region((0,0), (10,10))
>>> import numpy as np
>>> np.dtype(b.__array_interface__["typestr"]) # b would expose `__cuda_array_interface__` if memory is in GPU.
dtype('uint8')
```

With this patch, we can convert data type to NumPy's dtype easily, and also can access cuCIM's DLDataType.

```python
>>> from cucim import CuImage
>>> a = CuImage("notebooks/input/image.tif")
>>> b = a.read_region((0,0), (10,10))
>>> import numpy as np
>>> b.typestr
'|u1'
>>> np.dtype(b.typestr) == np.uint8
True
>>> from cucim.clara import DLDataType, DLDataTypeCode
>>> b.dtype == DLDataType(DLDataTypeCode.DLUInt, 8, 1)
True
```

Fixes #243

### DIFF
--- a/cpp/include/cucim/cuimage.h
+++ b/cpp/include/cucim/cuimage.h
@@ -37,6 +37,14 @@
 #include <string>
 #include <vector>
 
+
+// Forward declarations for DLDataType's equality operator.
+template <>
+struct std::hash<DLDataType>;
+
+EXPORT_VISIBLE bool operator==(const DLDataType& lhs, const DLDataType& rhs);
+EXPORT_VISIBLE bool operator!=(const DLDataType& lhs, const DLDataType& rhs);
+
 namespace cucim
 {
 
--- a/cpp/include/cucim/cuimage.h
+++ b/cpp/include/cucim/cuimage.h
@@ -148,6 +148,8 @@ public:
 
     DLDataType dtype() const;
 
+    std::string typestr() const;
+
     std::vector<std::string> channel_names() const;
 
     std::vector<float> spacing(std::string dim_order = std::string{}) const;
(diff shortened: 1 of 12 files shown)


gigony:
It seems that there is an infra issue.

Even if `libnvjpeg-dev-11-6_11.6.0.55-1_arm64.deb` is listed <https://developer.download.nvidia.com/compute/cuda/repos/ubuntu2004/sbsa/>, cannot download <https://developer.download.nvidia.com/compute/cuda/repos/ubuntu2004/sbsa/libnvjpeg-dev-11-6_11.6.0.55-1_arm64.deb>

jakirkham:
Generally this seems reasonable. Would be good to have a test. Are we thinking about other endianness?

gigony:
> Generally this seems reasonable. Would be good to have a test. Are we thinking about other endianness?

Since our major platform (x86-64 and aarch64 on Jetson) uses little-endian only and most TIFF files are using little-endian, supporting big-endian is lower-priority.

Will add tests later today when I am able to access my machine.

jakirkham (approved):

jakirkham (approved):

PR #252, opened 2022-03-31
Fix ImportError from vendored code

### LINKED ISSUE #251: [BUG] ImportError with development branch of CuPy
**Describe the bug**

While running the tests today, I observed the following ImportError from `match_template`.

```
=================================== FAILURES ===================================
____________________________ test_template[float32] ____________________________
feature/tests/test_template.py:28: in test_template
    result = match_template(image, target)
feature/template.py:124: in match_template
    from cucim.skimage import _vendored as signal
_vendored/__init__.py:6: in <module>
    from cucim.skimage._vendored.signaltools import *  # noqa
_vendored/signaltools.py:6: in <module>
    from cupyx.scipy.ndimage import _util, filters
E   ImportError: cannot import name 'filters' from 'cupyx.scipy.ndimage' (/home/lee8rx/my_git/pyir/cupy/cupyx/scipy/ndimage/__init__.py)
```

I think this is because `cupyx.scipy.ndimage.filters` was renamed `cupyx.scipy.ndimage._filters`. I will make a PR now to update this

grlee77:
closes #251 by avoiding imports from private CuPy code

The new files here are just subsets of files present in CuPy that are not part of its public API. Only a few functions use functionality from thi`s `_vendored` folder.

The whole folder is mainly here in order to have a version of `cupyx.scipy.signal.fftconvolve` with a modified `choose_conv_mode`. We should eventually just update the convolution selection code in CuPy so that we can remove these files.

I think `fftconvolve` itself was not present at all in CuPy 9.x, but even in the current development branch, the [convolution choice function is still restricted to 1D](https://github.com/cupy/cupy/blob/ec6d2c9a98440b19a133868576fb299972c21f6c/cupy/_math/misc.py#L23-L25)

### DIFF
--- a/python/cucim/src/cucim/skimage/_shared/utils.py
+++ b/python/cucim/src/cucim/skimage/_shared/utils.py
@@ -738,6 +738,11 @@ new_float_type = {
     cp.float16().dtype.char: cp.float32,
     'g': cp.float64,      # cp.float128 ; doesn't exist on windows
     'G': cp.complex128,   # cp.complex256 ; doesn't exist on windows
+    # int types where the full range can be represented exactly in float32
+    cp.int8: cp.float32,
+    cp.uint8: cp.float32,
+    cp.int16: cp.float32,
+    cp.uint16: cp.float32,
 }
 
 
--- a/python/cucim/src/cucim/skimage/_shared/utils.py
+++ b/python/cucim/src/cucim/skimage/_shared/utils.py
@@ -738,11 +738,6 @@ new_float_type = {
     cp.float16().dtype.char: cp.float32,
     'g': cp.float64,      # cp.float128 ; doesn't exist on windows
     'G': cp.complex128,   # cp.complex256 ; doesn't exist on windows
-    # int types where the full range can be represented exactly in float32
-    cp.int8: cp.float32,
-    cp.uint8: cp.float32,
-    cp.int16: cp.float32,
-    cp.uint16: cp.float32,
 }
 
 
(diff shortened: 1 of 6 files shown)


jakirkham:
Thanks Greg! 😄

Instead of vendoring what if we do a `try...except...` to handle `import`ing with the relocation?

Agree in the medium term would be good to contribute upstream to avoid needing these internal bits.

grlee77:
> what if we do a try...except... to handle importing with the relocation?

Do you mean modifying `_ndimage_filters.py` and others to something like this? 

```Python
try:
    import cupyx.scipy.ndimage

    if hasattr(cupyx.scipy.ndimage, '_filters'):
        # CuPy 11.x
        from cupyx.scipy.ndimage._filters import _get_correlate_kernel
    else:
        from cupyx.scipy.ndimage.filters import _get_correlate_kernel

except ImportError:

    import cupy

    from cucim.skimage._vendored import _ndimage_filters_core as _filters_core


    @cupy.memoize(for_each_device=True)
    def _get_correlate_kernel(mode, w_shape, int_type, offsets, cval):
        return _filters_core._generate_nd_kernel(
            'correlate',
            'W sum = (W)0;',
            'sum += cast<W>({value}) * wval;',
            'y = cast<Y>(sum);',
            mode, w_shape, int_type, offsets, cval, ctype='W')
```

That would let us potentially benefit from any upstream improvements or fixes to `_get_correlate_kernel`, but I don't think that outweighs the risk that the behavior may change in unexpected ways?

jakirkham:
@gpucibot merge

PR #271, opened 2022-05-04
parameterization of `filters` and `features` benchmarks

grlee77:
This PR introduces changes to allow selection of specific benchmarks from the command line for the `cucim.skimage.filters` and `cucim.skimage.feature` benchmarks. Previously, the scripts would run over all functdions which is undesirable when wanting to focus on a specific case. The new shell scripts here allow looping over all benchmarks as before while also potentially looping over the image shape and/or data types.

@anaszain89, can you confirm that this is the correct branch from your repository to consider? I think the changes here are quite helpful and we can proceed to update the other modules in a similar fashion.

The one thing I would like to address is having an option to enable the CPU-based comparison rather than just commenting it out in `_image_bench.py`. We could have it off by default, but enable it via another command line argument if requested. What do you think?


grlee77:
> Regarding the CPU option, I would actually like to keep it ON by default and needed we can turn it OFF. We had commented it out for our internal testing.

Yeah, keeping it on by default sounds good. Do you want to update your branch or should I make the changes in a separate branch and create the PR from that?

anaszain89:
@grlee77 - I have updated my branch with the parameterizations. This should be in line with what we talked about earlier.

PR #272, opened 2022-05-10
Fix: return object-typed properties as NumPy arrays in `skimage.measure.regionprops_table`

alxndrkalinin:
This PR fixes adding properties whose results are objects to the dictionary that is returned by `skimage.measure.regionprops_table`.

Minimal example reproducing the issue:
```python
from cucim.skimage.measure import regionprops_table

SAMPLE_MULTIPLE = cp.eye(10, dtype=cp.int32)
SAMPLE_MULTIPLE[3:5, 7:8] = 2

props = regionprops_table(SAMPLE_MULTIPLE, properties=['coords'])
```

raises `TypeError: Implicit conversion to a NumPy array is not allowed.`, while returning correct values is expected (also see modified unit tests).

For some reason, only `slice` property was [checked for](https://github.com/rapidsai/cucim/pull/272/commits/380bcc27584a7b4ec5dc2d27f98dd9b9d9433a70#diff-0019b8f1fc21c156da7d64a20335392909b847c2bf6f4cb115915065ce8353e5L821), although the list of such properties already exists in `OBJECT_COLUMNS`.

### DIFF
--- a/python/cucim/src/cucim/skimage/measure/_regionprops.py
+++ b/python/cucim/src/cucim/skimage/measure/_regionprops.py
@@ -84,11 +84,6 @@ PROPS = {
     'weighted_moments_normalized': 'moments_weighted_normalized',
 }
 
-OBJECT_COLUMNS = {
-    'image', 'coords', 'image_convex', 'slice',
-    'image_filled', 'image_intensity'
-}
-
 COL_DTYPES = {
     'area': int,
     'area_bbox': int,
@@ -131,6 +126,8 @@ COL_DTYPES = {
     'slice': object,
     'solidity': float,
 }
+
+OBJECT_COLUMNS = [col for col, dtype in COL_DTYPES.items() if dtype == object]
 
 PROP_VALS = set(PROPS.values())
 
--- a/python/cucim/src/cucim/skimage/measure/_regionprops.py
+++ b/python/cucim/src/cucim/skimage/measure/_regionprops.py
@@ -818,18 +818,16 @@ def _props_to_dict(regions, properties=('label', 'bbox'), separator='-'):
             or prop in OBJECT_COLUMNS
             or dtype is np.object_
         ):
-            if prop == 'slice':
-                # keep slice objects in a NumPy array
+            if prop in OBJECT_COLUMNS:
+                # keep objects in a NumPy array
                 column_buffer = np.empty(n, dtype=dtype)
                 for i in range(n):
                     column_buffer[i] = regions[i][prop]
                 out[orig_prop] = np.copy(column_buffer)
             else:
-                column_buffer = []
+                column_buffer = cp.empty(n, dtype=dtype)
                 for i in range(n):
-                    p = regions[i][prop]
-                    column_buffer.append(p)
-                column_buffer = cp.array(column_buffer)
+                    column_buffer[i] = regions[i][prop]
                 out[orig_prop] = column_buffer
         else:
             if isinstance(rp, cp.ndarray):
--- a/python/cucim/src/cucim/skimage/measure/_regionprops.py
+++ b/python/cucim/src/cucim/skimage/measure/_regionprops.py
@@ -825,9 +825,11 @@ def _props_to_dict(regions, properties=('label', 'bbox'), separator='-'):
                     column_buffer[i] = regions[i][prop]
                 out[orig_prop] = np.copy(column_buffer)
             else:
-                column_buffer = cp.empty(n, dtype=dtype)
+                column_buffer = []
                 for i in range(n):
-                    column_buffer[i] = regions[i][prop]
+                    p = regions[i][prop]
+                    column_buffer.append(p)
+                column_buffer = cp.array(column_buffer)
                 out[orig_prop] = column_buffer
         else:
             if isinstance(rp, cp.ndarray):
--- a/python/cucim/src/cucim/skimage/measure/tests/test_regionprops.py
+++ b/python/cucim/src/cucim/skimage/measure/tests/test_regionprops.py
@@ -604,7 +604,7 @@ def test_props_to_dict():
     regions = regionprops(SAMPLE_MULTIPLE)
     out = _props_to_dict(regions, properties=('coords',))
     coords = np.empty(2, object)
-    coords[0] = cp.array((cp.arange(10), cp.arange(10))).T
+    coords[0] = cp.stack((cp.arange(10),) * 2, axis=-1)
     coords[1] = cp.array([[3, 7], [4, 7]])
     assert out['coords'].shape == coords.shape
     assert_array_equal(out['coords'][0], coords[0])
@@ -625,7 +625,7 @@ def test_regionprops_table():
 
     out = regionprops_table(SAMPLE_MULTIPLE, properties=('coords',))
     coords = np.empty(2, object)
-    coords[0] = cp.array((cp.arange(10), cp.arange(10))).T
+    coords[0] = cp.stack((cp.arange(10),) * 2, axis=-1)
     coords[1] = cp.array([[3, 7], [4, 7]])
     assert out['coords'].shape == coords.shape
     assert_array_equal(out['coords'][0], coords[0])
--- a/python/cucim/src/cucim/skimage/measure/tests/test_regionprops.py
+++ b/python/cucim/src/cucim/skimage/measure/tests/test_regionprops.py
@@ -601,6 +601,15 @@ def test_props_to_dict():
                    'bbox+0': cp.array([0]), 'bbox+1': cp.array([0]),
                    'bbox+2': cp.array([10]), 'bbox+3': cp.array([18])}
 
+    regions = regionprops(SAMPLE_MULTIPLE)
+    out = _props_to_dict(regions, properties=['coords'])
+    coords = np.empty(2, object)
+    coords[0] = cp.array((cp.arange(10), cp.arange(10))).T
+    coords[1] = cp.array([[3, 7], [4, 7]])
+    assert out['coords'].shape == coords.shape
+    assert_array_equal(out['coords'][0], coords[0])
+    assert_array_equal(out['coords'][1], coords[1])
+
 
 def test_regionprops_table():
     out = regionprops_table(SAMPLE)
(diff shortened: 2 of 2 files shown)


GPUtester:

jakirkham:
> Can one of the admins verify this patch?

@quasiben, could you please look at this one?

quasiben:
add to allowlist

jakirkham:
@grlee77, could you please review this? 🙂

python/cucim/src/cucim/skimage/measure/_regionprops.py
```diff
@@ -818,18 +818,16 @@ def _props_to_dict(regions, properties=('label', 'bbox'), separator='-'):
...
                     column_buffer[i] = regions[i][prop]
                 out[orig_prop] = np.copy(column_buffer)
             else:
-                column_buffer = []
+                column_buffer = cp.empty(n, dtype=dtype)
                 for i in range(n):
-                    p = regions[i][prop]
-                    column_buffer.append(p)
-                column_buffer = cp.array(column_buffer)
+                    column_buffer[i] = regions[i][prop]
```
grlee77, line 830: I don't think we should change this part to preallocate the array. Assigning element by element into the array seems substantially more expensive than creating a list and then copying that to the GPU at the end.

Example test script:
```Python
import math

import cupy as cp
from cupyx.profiler import benchmark


def set_via_list(vals):
   n = len(vals)
   out = []
   for i in range(n):
      out.append(vals[i])
   return cp.asarray(out)


def set_in_preallocated_array(vals):
   n = len(vals)  
   out = cp.empty((n,), dtype=cp.float32)
   for i in range(n):
      out[i] = vals[i]
   return out


print(f"n       | Duration (list) | Duration (preallocated array)")
print(f"--------|-----------------|-----------------")
for n in [10, 100, 1000, 10000, 100000]:
   vals = [0,] * n

   # determine number of repetitions corresponding to ~2 seconds
   temp = benchmark(set_in_preallocated_array, (vals,), n_warmup=1, n_repeat=4)
   n_repeat = math.ceil(2 / temp.gpu_times.mean())

   perf1 = benchmark(set_via_list, (vals,), n_warmup=100, n_repeat=n_repeat)
   perf2 = benchmark(set_in_preallocated_array, (vals,), n_warmup=100, n_repeat=n_repeat)
   print(f"{n} | {1000*perf1.gpu_times.mean():0.4f} ms | {1000*perf2.gpu_times.mean():0.4f} ms")

```

result on my system
n       | Duration (list) | Duration (preallocated array)
--------|-----------------|-----------------
10 | 0.0270 ms | 0.0402 ms
100 | 0.0351 ms | 0.3310 ms
1000 | 0.1207 ms | 3.3150 ms
10000 | 0.9659 ms | 33.3445 ms
100000 | 9.4307 ms | 339.0651 ms

python/cucim/src/cucim/skimage/measure/tests/test_regionprops.py
```diff
@@ -601,6 +601,15 @@ def test_props_to_dict():
                    'bbox+0': cp.array([0]), 'bbox+1': cp.array([0]),
                    'bbox+2': cp.array([10]), 'bbox+3': cp.array([18])}
 
+    regions = regionprops(SAMPLE_MULTIPLE)
+    out = _props_to_dict(regions, properties=('coords',))
+    coords = np.empty(2, object)
+    coords[0] = cp.array((cp.arange(10), cp.arange(10))).T
```
grlee77, line 607: equivalent, but seems a bit clearer to me

```suggestion
    coords[0] = cp.stack((cp.arange(10),) * 2, axis=-1)
```

another option would be
```Python
coords[0] = cp.stack(cp.where(SAMPLE_MULTIPLE == 1), axis=-1)
```

python/cucim/src/cucim/skimage/measure/tests/test_regionprops.py
```diff
@@ -614,6 +623,14 @@ def test_regionprops_table():
                    'bbox+0': cp.array([0]), 'bbox+1': cp.array([0]),
                    'bbox+2': cp.array([10]), 'bbox+3': cp.array([18])}
 
+    out = regionprops_table(SAMPLE_MULTIPLE, properties=('coords',))
+    coords = np.empty(2, object)
+    coords[0] = cp.array((cp.arange(10), cp.arange(10))).T
```
grlee77, line 628:
```suggestion
    coords[0] = cp.stack((cp.arange(10),) * 2, axis=-1)
```

grlee77 (changes requested):
Thanks, @alxndrkalinin, the main fix here looks good to me! 

I have one request to revert changes to the non-object code path in `_regionprops.py` as I think it will be slower. If you measured an improvement, though, please let me know.

python/cucim/src/cucim/skimage/measure/_regionprops.py
alxndrkalinin: Ah, of course, don't want to move elements to GPU one by one – thank you for this example.

It's not very important after fixing the bug, but I have 2 small suggestions to make adding new properties easier.

1. Properties that return objects are listed in `OBJECT_COLUMNS`. but also in `COL_DTYPES` along with other props. Perhaps, reducing it to one place would make adding new props less error prone:
```python
OBJECT_COLUMNS = [k for k, v in COL_DTYPES.items() if v == object]
```
2. Calling `cp.asarray` on a list of objects threw an error that was confusing: `*** TypeError: Implicit conversion to a NumPy array is not allowed. Please use '.get()' to construct a NumPy array explicitly.`. Another option is to preallocate numpy array, then `cp.asarray` throws more informative `*** ValueError: Unsupported dtype object`. 

I added this to your test and it seems to be on par with the list:

```python
def set_in_preallocated_array(vals):
   n = len(vals)  
   out = np.empty((n,), dtype=cp.float32)
   for i in range(n):
      out[i] = vals[i]
   return cp.asarray(out)
```

n       | Duration (list) | Duration (preallocated array) | Duration (NumPy array)
--------|-----------------|-----------------|-----------------
10 | 0.0594 ms | 0.0879 ms | 0.0594 ms
100 | 0.0827 ms | 0.7180 ms | 0.0817 ms
1000 | 0.3054 ms | 7.0958 ms | 0.3044 ms
10000 | 2.4602 ms | 71.2040 ms | 2.5394 ms
100000 | 23.9010 ms | 710.5080 ms | 24.8934 ms

WDYT?

python/cucim/src/cucim/skimage/measure/_regionprops.py
```diff
@@ -818,18 +818,16 @@ def _props_to_dict(regions, properties=('label', 'bbox'), separator='-'):
...
-                # keep slice objects in a NumPy array
+            if prop in OBJECT_COLUMNS:
+                # keep objects in a NumPy array
                 column_buffer = np.empty(n, dtype=dtype)
                 for i in range(n):
                     column_buffer[i] = regions[i][prop]
                 out[orig_prop] = np.copy(column_buffer)
             else:
-                column_buffer = []
+                column_buffer = cp.empty(n, dtype=dtype)
```
alxndrkalinin, line 828:
```suggestion
                column_buffer = np.empty(n, dtype=dtype)
```

python/cucim/src/cucim/skimage/measure/_regionprops.py
```diff
@@ -818,18 +818,16 @@ def _props_to_dict(regions, properties=('label', 'bbox'), separator='-'):
...
                 for i in range(n):
-                    p = regions[i][prop]
-                    column_buffer.append(p)
-                column_buffer = cp.array(column_buffer)
+                    column_buffer[i] = regions[i][prop]
                 out[orig_prop] = column_buffer
```
alxndrkalinin, line 831:
```suggestion
                out[orig_prop] = cp.asarray(column_buffer)
```

alxndrkalinin (commented):
Thanks for the feedback @grlee77!

Please see 2 small suggestions I have – if you think they're unnecessary, I'll be happy to just revert back to the list for non-object props.

python/cucim/src/cucim/skimage/measure/_regionprops.py
grlee77: > Perhaps, reducing it to one place would make adding new props less error prone:

That seems like a good idea. I think that logic was copied from upstream skimage, so we could follow up with a similar PR there.

> Another option is to preallocate numpy array, then cp.asarray throws more informative *** ValueError: Unsupported dtype object.

Yes, preallocating a numpy array as you suggest is a reasonable alternative. It does still have some performance penalty on my system, though, so I would prefer to leave it as-is. Hopefully suggestion 1.) above will avoid encountering this opaque error message in the future.

```
n       | Duration (list) | Duration (NumPy array)
--------|-----------------|-----------------
10 | 0.0216 ms | 0.0212 ms
100 | 0.0304 ms | 0.0309 ms
1000 | 0.1168 ms | 0.1271 ms
10000 | 0.8842 ms | 1.1074 ms
100000 | 8.7250 ms | 10.6378 ms
```

Actually, I don't see the confusing message you mentioned when running
```Python
cp.asarray([slice(5), slice(10)])
```

but instead get a more helpful "ValueError: Unsupported dtype object". Perhaps the difference is because I am on a development branch of CuPy, though.

```
alueError                                Traceback (most recent call last)
Input In [12], in <module>
----> 1 cp.asarray([slice(5), slice(10)])

File ~/mambaforge/envs/cucim/lib/python3.8/site-packages/cupy/_creation/from_data.py:76, in asarray(a, dtype, order)
     49 def asarray(a, dtype=None, order=None):
     50     """Converts an object to array.
     51 
     52     This is equivalent to ``array(a, dtype, copy=False)``.
   (...)
     74 
     75     """
---> 76     return _core.array(a, dtype, False, order)

File cupy/_core/core.pyx:2249, in cupy._core.core.array()

File cupy/_core/core.pyx:2270, in cupy._core.core.array()

File cupy/_core/core.pyx:2398, in cupy._core.core._array_default()

ValueError: Unsupported dtype object
```

alxndrkalinin: That's a nice GPU 😃 Agreed to leave as-is.

New commits:
* revert changes in non-object path back to adding props to a list first
* make OBJECT_COLUMNS to be derived from COL_DTYPES (happy to make a quick PR to scikit-image to reflect this change)


Re: error message
> I don't see the confusing message you mentioned ... but instead get a more helpful "ValueError: Unsupported dtype object"

That is only true for `slice`, other props like `image_` ones or `coords` return a list of multiple CuPy arrays with mismatching shapes, so better example would be 

```python
cp.asarray([cp.empty(5), cp.empty(10)])
```

or a snippet in my original post https://github.com/rapidsai/cucim/pull/272#issue-1230621346

grlee77 (approved):
Thanks @alxndrkalinin, this looks good now.

jakirkham (approved):

PR #273, opened 2022-05-11
Stain Normalizer (v2)

### LINKED ISSUE #96: [FEA] Add Stain Normalization operation (in Python)

grlee77:
closes #96

This PR resumes work that was started in #186. Given the large overall refactoring, it was not feasible to make the suggestions there as individual comments.

Overall the approach is the same Macenko method that was was proposed in #186. 

I spent quite a bit of time refactoring for performance and to separate out some aspects so that it will be easier to add additional related methods in the future. I find about 3x improvement for the case here vs. the one in #186.

A summary of the changes relative to #186 are:

### Enhancements
- added a `channel_axis` argument that can be used to specify which axis of the
  input array corresponds to color channels.
- more descriptive docstrings

### General refactoring
- Now also provides a function-based interface as well to more closely match
  the typical `cucim.skimage` style. The existing class-based interface was
  kept as well. There is a small amount of redundancy in providing both, so we
  should decide if this is worth it.
- Aspects like conversion to/from absorbance units was separated out into
  separate functions.

### Performance Related Changes

- Used `cupy.fuse` to fuse multiple kerenel operations needed for absorbance
  calculations into a single GPU kernel. This gives ~4x improvement in
  conversions to/from absorbance space.
- Added an `image_type` argument that defaults to 'intensity', but can be set
  to 'absorbance' to indicate that the image is already in absorbance space.
  This is used to avoid redundant conversions during stain normalization.
- Added a `_covariance` function that is a simplified and optimized version of
  `cupy.cov`. It runs 4x faster for me for the float32 test case I tried on a
  roughly size 2000x2000 image.
- Added a second 'ortho' method aside from least-squares for estimating the raw
  concentrations. The `_complement_stain_matrix` helper adds a third column
  that is orthogonal to the two estimated stain vectors, so that a standard
  matrix inverse can be used. This is much faster in practice than calling
  `cupy.linalg.lstsq` and gives identical result for almost all voxels in
  test images. A tiny fraction of voxels differed in uint8 intensity by a
  magnitude of 1, but this is likely just due differences in the rounding
  result of finite precision floating point values. This approach is based on
  the one used by HistomicsTK software (Apache 2.0 licensed).

### Test Changes
- refactored to use `pytest.mark.parametrize` instead of adding a dependency on [parameterized](https://github.com/wolever/parameterized)
- There is an increase by 1 in some expected pixel values in the "expected" result for some test cases due to replacement of casting to integers (floor operation) with the use of rounding upon conversion of `float32` back to `uint8` during color normalization.


grlee77:
@drbeh, please see this refactored version of the stain normalization proposed previously. If there is no desire for a function-based interface we can remove the `normalize_colors_macenko` method and just keep the `StainNormalizer` class.

drbeh:
Thank you very much @grlee77 for this upgraded version. I'll review it today.

grlee77:
Also, @thewtex pointed me to this method that was implemented in ITK:
https://github.com/InsightSoftwareConsortium/ITKColorNormalization

 I can try to take a closer look at it at some point to estimate how easy it is to adapt to the GPU.

python/cucim/src/cucim/core/operations/color/stain_normalizer.py
```diff
@@ -0,0 +1,821 @@
...
+    # Make Hematoxylin (H) first and eosin (E) second by comparing the
+    # R channel value
+    if v1[0] < v2[0]:
+        v1, v2 = v2, v1
+    stain_coeff = np.stack((v1, v2), axis=-1)
+
+    # renormalize columns to reduce numerical error
+    stain_coeff /= np.linalg.norm(stain_coeff, axis=0, keepdims=True)
+    if append_third_column:
+        stain_coeff = _complement_stain_matrix(stain_coeff)
```
drbeh, line 402: Is there any reason to do this calculations on the host?

python/cucim/src/cucim/core/operations/color/stain_normalizer.py
```diff
@@ -0,0 +1,821 @@
...
+    cov = _covariance(absorbance, rowvar=True)
+    cov = cp.asnumpy(cov).astype(np.float32, copy=False)
+    _, ev = np.linalg.eigh(cov)
+    ev = ev[:, [2, 1]]
+    # flip to ensure positive first coordinate so arctan2 angles are about 0
+    if ev[0, 0] < 0:
+        ev[:, 0] *= -1
+    if ev[0, 1] < 0:
+        ev[:, 1] *= -1
+
```
grlee77, line 377: This flipping was added here based on an MIT-licensed implementation in:
https://github.com/Peter554/StainTools/blob/2089900d11173ee5ea7de95d34532932afd3181a/staintools/stain_extraction/macenko_stain_extractor.py#L29-L37

python/cucim/src/cucim/core/operations/color/stain_normalizer.py
```diff
@@ -0,0 +1,821 @@
...
+    # convert to absorbance (optical density) matrix of shape (n_pixels, 3)
+    absorbance = _image_to_absorbance_matrix(
+        image,
+        source_intensity=source_intensity,
+        image_type=image_type,
+        channel_axis=channel_axis
+    )
+
+    # channels_axis=0 for the shape (3, n_pixels) absorbance matrix
+    src_stain_coeff = stain_decomposition_macenko(
```
grlee77, line 563: should perhaps change "decomposition"->"extraction". not sure why I had named it this way

python/cucim/src/cucim/core/operations/color/stain_normalizer.py
```diff
@@ -0,0 +1,821 @@
...
+        conc_raw = cp.linalg.lstsq(
+            src_stain_coeff, absorbance, rcond=None
+        )[0]
+    elif method == 'ortho':
+        if src_stain_coeff.shape != (3, 3):
+            raise ValueError("expected a set of three stain vectors")
+        # This approach relies on a square stain coeffs matrix as used by
+        # HistomicsTK. In practice, it gives nearly identical results to the
+        # least-squares approach.
+        coeff_inv = cp.linalg.inv(src_stain_coeff)
```
drbeh, line 425: Instead of creating the extra column for statin coefficients, can't we use pseudo-inverse (for rectangular matrices)? `{X^T.X}^{-1}X^T`

python/cucim/src/cucim/core/operations/color/stain_normalizer.py
```diff
@@ -0,0 +1,821 @@
...
+        )
+    # restore original shape
+    return image_norm
+
+
+def normalize_colors_macenko(
+        image,
+        source_intensity: float = 240.0,
+        alpha: float = 1.0,
+        beta: float = 0.15,
```
drbeh, line 486: As you mentioned before since we are using `log` instead of `log10`, maybe we should adjust this value to `0.15/log10(e) ~ 0.345`, what do you think?

drbeh (commented):
@grlee77 thank you very much for refactoring this stain normalizer. Overall it looks great but I left some comments in the code.

drbeh:
> @drbeh, please see this refactored version of the stain normalization proposed previously. If there is no desire for a function-based interface we can remove the `normalize_colors_macenko` method and just keep the `StainNormalizer` class.

Either way is fine but I think it depends on `cucim`'s approach for transforms. The previous ones had functional interfaces so it makes sense to keep only functional interfaces here.

python/cucim/src/cucim/core/operations/color/stain_normalizer.py
```diff
@@ -0,0 +1,821 @@
...
+]
+
+
+@cp.fuse()
+def _image_to_absorbance(image, min_val, max_val):
+    image = cp.minimum(image, max_val)
+    image = cp.maximum(image, min_val)
+    image /= max_val
+    image = cp.log(image)
+    return -image
```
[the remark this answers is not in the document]
grlee77: how about:

```Python
@cp.fuse()
def _image_to_absorbance(image, min_val, max_val):
    image = cp.maximum(cp.minimum(image, max_val), min_val) / max_val
    return -cp.log(image)
```

(performance is equivalent)

python/cucim/src/cucim/core/operations/color/stain_normalizer.py
grlee77: I will change it and add a note to indicate it is equivalent to the default in the publication

python/cucim/src/cucim/core/operations/color/stain_normalizer.py
```diff
@@ -0,0 +1,821 @@
...
+        alpha: float = 1.0,
+        beta: float = 0.15,
+        ref_stain_coeff: Union[tuple, cp.ndarray] = (
+            (0.5626, 0.2159),
+            (0.7201, 0.8012),
+            (0.4062, 0.5581),
+        ),
+        ref_max_conc: Union[tuple, cp.ndarray] = (1.9705, 1.0308),
+        image_type: str = 'intensity',
+        channel_axis: int = -1,
```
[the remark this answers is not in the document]
grlee77: Yes. I was in the habit of using -1 as that is the scikit-image default. It makes sense to adapt default this file to MONAI/PyTorch conventions, though.

python/cucim/src/cucim/core/operations/color/stain_normalizer.py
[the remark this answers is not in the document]
drbeh: Sure, that's fine too but just for the sake of slightly more readability, what do you think of separating them into two lines, first line about image value cliping and second line about absorbance calculation:
```python
@cp.fuse()
def _image_to_absorbance(image, min_val, max_val):
    image = cp.maximum(cp.minimum(image, max_val), min_val)
    absorbance = -cp.log(image / max_val)
    return absorbance
```

python/cucim/src/cucim/core/operations/color/stain_normalizer.py
```diff
@@ -0,0 +1,821 @@
...
+        fact = 0.0
+
+    X -= X.mean(axis=1, keepdims=True)
+    if not X.flags.f_contiguous:
+        # TODO: Empirically, it is faster to use F order for .dot() call below.
+        #       F-order was ~4x faster for float32 when measured, but slightly
+        #       slower for float64, which seems odd. Should further validate on
+        #       additional hardware.
+        X = cp.asfortranarray(X)
+    out = X.dot(X.T.conj())
```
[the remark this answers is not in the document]
grlee77: I think in practice `.conj()` is near-instantaneous for real-valued inputs. Calling .T similarly is cheap as it doesn't make a copy, but just modifies the strides. Example

```Python
import cupy as cp
a = cp.ones((3, 1000000))
d = cp.cuda.Device()
%timeit a.T.conj(); d.synchronize()
```
gives:
`1.21 µs ± 4.21 ns per loop (mean ± std. dev. of 7 runs, 1,000,000 loops each)`

whereas just making a copy of `a` takes ~125 µs

python/cucim/src/cucim/core/operations/color/stain_normalizer.py
```diff
@@ -0,0 +1,821 @@
...
+    out *= 1 / cp.float64(fact)
```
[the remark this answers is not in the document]
grlee77: It was just to force the multiplication to be done in-place. Otherwise, I think another temporary array is created?

python/cucim/src/cucim/core/operations/color/stain_normalizer.py
grlee77: Out of curiousity, I ran the following
```Python
ev = cp.random.randn(3, 3).astype(np.float32)
min_phi = cp.asarray(-.5)
max_phi = cp.asarray(0.5)

def _on_host(cov, min_phi, max_phi):
    cov = cp.asnumpy(cov).astype(np.float32, copy=False)
    _, ev = np.linalg.eigh(cov)
    ev = ev[:, [2, 1]]
    # flip to ensure positive first coordinate so arctan2 angles are about 0
    if ev[0, 0] < 0:
        ev[:, 0] *= -1
    if ev[0, 1] < 0:
        ev[:, 1] *= -1
    min_phi, max_phi = float(min_phi), float(max_phi)

    # project back to absorbance space
    v_min = np.array([math.cos(min_phi), math.sin(min_phi)], dtype=np.float32)
    v_max = np.array([math.cos(max_phi), math.sin(max_phi)], dtype=np.float32)
    v1 = np.dot(ev, v_min)
    v2 = np.dot(ev, v_max)

    # Make Hematoxylin (H) first and eosin (E) second by comparing the
    # R channel value
    if v1[0] < v2[0]:
        v1, v2 = v2, v1
    stain_coeff = np.stack((v1, v2), axis=-1)

    # renormalize columns to reduce numerical error
    stain_coeff /= np.linalg.norm(stain_coeff, axis=0, keepdims=True)
    return cp.asarray(stain_coeff)
    
def _on_device(cov, min_phi, max_phi):

    # project back to absorbance space
    _, ev = cp.linalg.eigh(cov)
    ev = ev[:, [2, 1]]
    # flip to ensure positive first coordinate so arctan2 angles are about 0
    if ev[0, 0] < 0:
        ev[:, 0] *= -1
    if ev[0, 1] < 0:
        ev[:, 1] *= -1    
    v_min = cp.array([cp.cos(min_phi), cp.sin(min_phi)], dtype=np.float32)
    v_max = cp.array([cp.cos(max_phi), cp.sin(max_phi)], dtype=np.float32)
    v1 = cp.dot(ev, v_min)
    v2 = cp.dot(ev, v_max)

    # Make Hematoxylin (H) first and eosin (E) second by comparing the
    # R channel value
    if v1[0] < v2[0]:
        v1, v2 = v2, v1
    stain_coeff = cp.stack((v1, v2), axis=-1)

    # renormalize columns to reduce numerical error
    stain_coeff /= cp.linalg.norm(stain_coeff, axis=0, keepdims=True)
    return cp.asarray(stain_coeff)

```

which gives
```
In [22]: %timeit _on_host(ev, min_phi, max_phi); d.synchronize()
128 µs ± 1.52 µs per loop (mean ± std. dev. of 7 runs, 10,000 loops each)

In [23]: %timeit _on_device(ev, min_phi, max_phi); d.synchronize()
677 µs ± 119 µs per loop (mean ± std. dev. of 7 runs, 1 loop each)
```

So, approximately 0.5 ms additional overhead regardless of image size when all of these tiny calculations are done on the GPU.

python/cucim/src/cucim/core/operations/color/stain_normalizer.py
grlee77: Isn't this pinv-based solution just "least squares"? Perhaps we should just remove the `method` argument for now and always use the psuedo-inverse code path.

I was thinking `method` could be useful later if we were to add additional methods like LASSO or a non-negative matrix factorization that would ensure non-negative concentrations.

python/cucim/src/cucim/core/operations/color/stain_normalizer.py
```diff
@@ -0,0 +1,821 @@
...
+    if conc_raw.ndim != 2 or conc_raw.shape[0] != 2:
+        raise ValueError(
+            "`conc_raw` must be a 2D array of concentrations with size 2 on "
+            "axis 0."
+        )
+    if ref_stain_coeff.shape != (3, 2):
+        raise ValueError(
+            "`ref_stain_coeff` must be a shape (3, 2) matrix, representing "
+            "two stain vectors."
+        )
```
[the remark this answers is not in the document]
grlee77: yes, I will change it to check shape `(3, n)` here and length should be `n` for `ref_max_conc`

python/cucim/src/cucim/core/operations/color/__init__.py
```diff
@@ -1,6 +1,11 @@
...
+from .stain_normalizer import StainNormalizer, HEStainExtractor
 
 __all__ = [
     "color_jitter",
     "rand_color_jitter"
+    "absorbance_to_image",
+    "image_to_absorbance",
+    "StainNormalizer",
+    "HEStainExtractor",
 ]
```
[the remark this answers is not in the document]
grlee77: Yes, and I still need to change the tests to use those as well.

python/cucim/src/cucim/core/operations/color/stain_normalizer.py
```diff
@@ -0,0 +1,821 @@
...
+    if not dtype.kind == 'f':
+        raise ValueError("dtype must be a floating point type")
+
+    image = image.astype(dtype, copy=False)
+    if source_intensity < 0:
+        raise ValueError(
+            "Source transmitted light intensity must be a positive value."
+        )
+    source_intensity = float(source_intensity)
+    if dtype.kind == 'f':
```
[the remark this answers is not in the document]
drbeh: Shouldn't the `min_val` always be `1`? Otherwise, we will end up having negative values for absorbance, right?

python/cucim/src/cucim/core/operations/color/stain_normalizer.py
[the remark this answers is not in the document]
drbeh: Right but anyways you have a division, so there isn't any gain over:
```python
 out = X.dot(X.T.conj()) / cp.float64(fact)
```

python/cucim/src/cucim/core/operations/color/stain_normalizer.py
drbeh: I like that you added different methods but if the gain is not that much for now, we can keep only one of the methods and add the argument if needed in the future.

python/cucim/src/cucim/core/operations/color/stain_normalizer.py
drbeh: Both are fine to me since you need to decompose stains to extract them.
However, speaking on naming, do you think that we can give it a more representative name to this method? Rather than the family name of the first author of that paper, maybe we can come use the name of underlying method, since it is not a new method per se, and the paper just showed how they have used this method in histopathology.

python/cucim/src/cucim/core/operations/color/stain_normalizer.py
```diff
@@ -0,0 +1,821 @@
...
+        iinfo = cp.iinfo(dtype)
+        image = _absorbance_to_image_int(
+            absorbance, source_intensity, iinfo.min, iinfo.max
+        )
+        return image.astype(dtype, copy=False)
+    else:
+        return _absorbance_to_image_float(absorbance, source_intensity)
+
+
+def _covariance(a, rowvar=True):
```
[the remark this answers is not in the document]
grlee77: done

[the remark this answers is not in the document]
drbeh: Great! Thanks!

python/cucim/src/cucim/core/operations/color/__init__.py
```diff
@@ -1,6 +1,13 @@
...
+from .stain_normalizer import (absorbance_to_image, image_to_absorbance,
+                               normalize_colors_macenko,
+                               stain_extraction_macenko)
 
 __all__ = [
     "color_jitter",
     "rand_color_jitter"
+    "absorbance_to_image",
+    "image_to_absorbance",
+    'stain_extraction_macenko',
```
gigony, line 11: Nit: quotation mark is not consistent. May want to use " instead of ' in line 11 and 12.
Or, please stick to one and apply the same rule to other cases.

python/cucim/tests/unit/core/test_stain_normalizer.py
```diff
@@ -0,0 +1,251 @@
+# Copyright (c) 2021, NVIDIA CORPORATION.
```
gigony, line 1:
```suggestion
# Copyright (c) 2022, NVIDIA CORPORATION.
```

LICENSE-3rdparty.md
```diff
@@ -275,3 +275,9 @@ Taskflow
   - https://github.com/taskflow/taskflow
 - Copyright: Dr. Tsung-Wei Huang
 - Usage: Threadpool implementation for batch processing.
+
+StainTools
+- License: MIT License
+  - https://github.com/Peter554/StainTools/blob/master/LICENSE.txt
```
gigony, line 281: It would be nice if you could copy & paste the raw text to 3rdparty/LICENSE.StainTools.

python/cucim/src/cucim/core/operations/color/stain_normalizer.py
```diff
@@ -0,0 +1,534 @@
+# Copyright (c) 2021, NVIDIA CORPORATION.
```
gigony, line 1:
```suggestion
# Copyright (c) 2022, NVIDIA CORPORATION.
```

gigony (approved):
Thank you so much @grlee77 ! It's been a long time since this implementation has been around.
Neha(author of original algorithm implementation, @nsrivathsa) would love this work :)

And thank you @drbeh for the great feedback!

python/cucim/src/cucim/core/operations/color/stain_normalizer.py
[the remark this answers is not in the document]
grlee77: Okay, I will do the division on the host, so there is just one in-place multiplication
```Python
    out *= 1 / float(fact)
```

LICENSE-3rdparty.md
grlee77: Done, thanks!

jakirkham (approved):

PR #276, opened 2022-05-13
improve efficiency of histogram-based thresholding functions

grlee77:
A number of image thresholding functions involve computations on the histogram counts and/or bins. Given that the histogram outputs are small (only 256 elements by default), it is much more efficient to perform computations on them via the CPU.

Overall, computation time of these functions is dominated by the histogram computation itself (at least for larger images), so that is the part it is important to perform on the GPU.

I benchmarked `threshold_otsu` on a (512, 512) grayscale uint8 image (`skimage.data.camera()`). Prior to the change in this PR duration was: 0.894 ms which is reduced to 0.686 ms by the changes made here. The improvement should be around 0.2 ms regardless of image size (since the number of histogram bins is independent of image size).

### DIFF
--- a/python/cucim/src/cucim/skimage/filters/thresholding.py
+++ b/python/cucim/src/cucim/skimage/filters/thresholding.py
@@ -366,6 +366,9 @@ def threshold_otsu(image=None, nbins=256, *, hist=None):
             return first_pixel
 
     counts, bin_centers = _validate_image_histogram(image, hist, nbins)
+    # small size counts, bin_centers -> faster on the host
+    counts = cp.asnumpy(counts)
+    bin_centers = cp.asnumpy(bin_centers)
 
     # class probabilities for all possible thresholds
     weight1 = np.cumsum(counts)
@@ -379,10 +382,10 @@ def threshold_otsu(image=None, nbins=256, *, hist=None):
     # ``weight2``/``mean2``, which do not exist.
     variance12 = weight1[:-1] * weight2[1:] * (mean1[:-1] - mean2[1:]) ** 2
 
-    idx = cp.argmax(variance12)
+    idx = np.argmax(variance12)
     threshold = bin_centers[idx]
 
-    return threshold
+    return cp.asarray(threshold)
 
 
 def threshold_yen(image=None, nbins=256, *, hist=None):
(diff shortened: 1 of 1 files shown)


gigony (approved):
Thanks @grlee77! Looks great to me.

jakirkham:
Would it make sense to just use these functions from scikit-image if they are running on host? Perhaps we could have some wrapper function for shepherding data between host and devices on the boundaries of these calls

python/cucim/src/cucim/skimage/filters/thresholding.py
```diff
@@ -366,6 +366,9 @@ def threshold_otsu(image=None, nbins=256, *, hist=None):
             return first_pixel
 
     counts, bin_centers = _validate_image_histogram(image, hist, nbins)
+    # small size counts, bin_centers -> faster on the host
+    counts = cp.asnumpy(counts)
+    bin_centers = cp.asnumpy(bin_centers)
```
grlee77, line 371: @jakirkham, as a concrete example of what you were asking, now that we have computed the histogram and transfered back to the host, we could replace the rest of this function with a call to the skimage function:

```Python
return skimage.threshold_otsu(image=None, nbins=nbins, hist=(counts, bin_centers))
```
It does look like the `hist` argument has been available since scikit-image 0.18

grlee77: I have now removed redundant code and call the scikit-image functions with the histogram that was computed on the GPU. See commit fc73695

python/cucim/src/cucim/skimage/filters/thresholding.py
```diff
@@ -353,6 +353,8 @@ def threshold_otsu(image=None, nbins=256, *, hist=None):
     -----
     The input image must be grayscale.
     """
+    from skimage.filters import threshold_otsu as threshold_otsu_cpu
```
jakirkham, line 356: Would it make sense to move this `import` to the top-level? We could add an `_` before it if we are concerned with users thinking this is a public API

grlee77: Okay, I went ahead and moved all these up top

gigony:
@grlee77 Please help resolve the test errors.

https://gpuci.gpuopenanalytics.com/job/rapidsai/job/gpuci/job/cucim/job/prb/job/cucim-gpu-test/CUDA=11.0,GPU_LABEL=driver-450,LINUX_VER=centos7,PYTHON=3.8/292/console
```
12:40:07 /opt/conda/envs/rapids/lib/python3.8/site-packages/_pytest/assertion/rewrite.py:170: in exec_module
12:40:07     exec(co, module.__dict__)
12:40:07 src/cucim/skimage/filters/tests/test_thresholding.py:18: in <module>
12:40:07     from cucim.skimage.filters.thresholding import (_cross_entropy,
12:40:07 /opt/conda/envs/rapids/lib/python3.8/site-packages/cucim/skimage/filters/thresholding.py:10: in <module>
12:40:07     from skimage.filters import threshold_isodata_cpu as _threshold_isodata_cpu
12:40:07 E   ImportError: cannot import name 'threshold_isodata_cpu' from 'skimage.filters' (/opt/conda/envs/rapids/lib/python3.8/site-packages/skimage/filters/__init__.py)
```

grlee77:
There was a typo. It is now fixed (verified tests pass locally)

jakirkham (approved):

jakirkham:
@gpucibot merge

jakirkham:
Thanks Greg! 🙏

PR #278, opened 2022-05-17
Promote small integer types to single rather than double precision

grlee77:
This is a performance-related PR that will result in casting of small integer dtypes (8 and 16-bit) to 32-bit floats rather than the current use of 64-bit floats. The current behavior is consistent with scikit-image, although I have raised an issue there to potentially change to the behavior proposed here: https://github.com/scikit-image/scikit-image/issues/6310. 

The changes required are quite small with the key one being the change to the `new_float_type` dict that gets used when promoting dtypes to a floating point type. Most other changes are in the tests where we sometimes have to bump up the tolerance when computations that were previously in double precision now get run in single precision instead.

I marked this as **breaking** because the output of various floating-point functions for integer inputs may now be `cp.float32` in cases where it was previously `cp.float64`.

### DIFF
--- a/python/cucim/src/cucim/skimage/_shared/tests/test_utils.py
+++ b/python/cucim/src/cucim/skimage/_shared/tests/test_utils.py
@@ -158,7 +158,7 @@ def test_validate_interpolation_order(dtype, order):
 def test_supported_float_dtype_real(dtype):
     float_dtype = _supported_float_type(dtype)
     if dtype in [np.float16, np.float32, np.int8, np.uint8, np.int16,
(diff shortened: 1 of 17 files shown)


jakirkham:
Just curious, are there any cases where half-precision floats would makes sense for us?

gigony (approved):
Looks good to me!

jakirkham (approved):

jakirkham:
@gpucibot merge

jakirkham:
Thanks Greg! 🙏

PR #284, opened 2022-05-20
Canny: improve performance with Elementwise kernel

grlee77:
This PR refactors the non-maximum suppression stage of the Canny filter into a single ElementwiseKernel instead of dozens of indivdual kernel calls. This stage of the algorithm became ~50x faster which translated into an overall ~3.5x reduction in runtime for Canny filtering.

Further benchmarks for `canny` in comparison to scikit-image are given below.

### Larger Image Benchmark (float32, mode='nearest')

Implementation   | shape        | duration (ms)
-----------------|--------------|--------------
skimage (0.19.2) | (2512, 2512) | 751.90
skimage (dev)    | (2512, 2512) | 346.67
cucim (22.04)    | (2512, 2512) |  29.45
cucim (dev)      | (2512, 2512) |   7.57

We see that for an image of shape (2512, 2512) there is a 47x advantage to the GPU implementation here as compared to the scikit-image dev branch.

### Small Image Benchmark (float32, mode='nearest')
Implementation   | shape        | duration (ms)
-----------------|--------------|--------------
skimage (0.19.2) | (157, 157)   |   3.30
skimage (dev)    | (157, 157)   |   1.10
cucim (22.04)    | (157, 157)   |   4.45
cucim (dev)      | (157, 157)   |   1.29

Even for a much smaller image of shape (157, 157), the implementation here is better than scikit-image 0.19 and competitive with the proposed improvement in https://github.com/scikit-image/scikit-image/pull/6387.

### DIFF
--- a/python/cucim/src/cucim/skimage/feature/_canny.py
+++ b/python/cucim/src/cucim/skimage/feature/_canny.py
@@ -13,18 +13,11 @@ Original author: Lee Kamentsky
 """
 import cupy as cp
 import cupyx.scipy.ndimage as ndi
-from cupyx.scipy.ndimage import binary_erosion, generate_binary_structure
 
 from cucim.skimage.util import dtype_limits
 
 from .._shared.filters import gaussian
 from .._shared.utils import _supported_float_type, check_nD
-
-
-# fuse several commonly paired ufunc operations into a single kernel call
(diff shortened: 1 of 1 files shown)


gigony (approved):
Looks good to me!
Great to see the huge improvement!

jakirkham (approved):

Issue #230, opened 2022-03-01
Images with no color channel!


bhashemian:
**Describe the bug**
When using cuCIM to load TIFF files that does not have color channel (a grayscale 2D image), no error is thrown and instead the values will be repeated to create an RGBA image, representing the grayscale image. This may cause confusion, like here: https://github.com/Project-MONAI/MONAI/issues/3858

**Steps/Code to reproduce bug**
```python
import numpy as np                                                                                                                                                                                                                           
from skimage import io    
from cucim import CuImage

image_path = "test.tif"
x = np.random.randint(low=0, high=256, size=(128,128), dtype=np.uint8)
io.imsave(image_path, x) 

img = np.array(CuImage(image_path).read_region())
print(f"{img.shape = }")
print(f"{img[0, 0, :] = }")
```

```
img.shape = (128, 128, 4)
img[0, 0, :] = array([121, 121, 121, 255], dtype=uint8)
```

drbeh:
@gigony is there any update on this? thanks

gigony:
@drbeh we would eventually need to support multi-band and other data types
- https://github.com/rapidsai/cucim/issues/33

Currently, if the image is not RAW/LZW/JPEG/JPEG2000-compressed tiled/multi-resolution RGB images, it delegates its processing to libtiff's `TIFFRGBAImageGet()` method that creates a 4-channel RGBA array.

What would you suggest until we support single-channel image data? Maybe raising a warning message would be one we can do?

drbeh:
> @drbeh we would eventually need to support multi-band and other data types
> 
> * [[FEA] Supporting multi bands (multi-channel) images for geospatial (remote sensing) and microscopy #33](https://github.com/rapidsai/cucim/issues/33)
> 
> Currently, if the image is not RAW/LZW/JPEG/JPEG2000-compressed tiled/multi-resolution RGB images, it delegates its processing to libtiff's `TIFFRGBAImageGet()` method that creates a 4-channel RGBA array.
> 
> What would you suggest until we support single-channel image data? Maybe raising a warning message would be one we can do?

Thank you, @gigony! Raising an error would satisfy the purpose of this open issue. Currently, we don't have any need for single-channel images in digital pathology but it might be a good feature to have.

gigony:
> Thank you, @gigony! Raising an error would satisfy the purpose of this open issue. Currently, we don't have any need for single-channel images in digital pathology but it might be a good feature to have.


[...]

PR #293, opened 2022-05-25
Show warning message when loading an image with a slow method

gigony:
Show a warning message if the image loading is handled by libtiff so its output has a 4-channel (RGBA).

Address #230 


grlee77 (approved):
Built locally and it was working as expected (had not been aware of `std::call_once`!).

Here is a local test I ran. I'm not sure if there is an easier way to capture the stderr text without needing to call a subprocess?

```Python
import subprocess

import numpy as np
from skimage import io

img = np.ones((128, 128), dtype=np.uint8)
io.imsave('test.tif', img)

cmd = (
    "python -c \""
    "from cucim import CuImage\n"
    "i = CuImage('test.tif')\n"
    "i.read_region()\n"
    "i.read_region()\n"
    "i.read_region()\""
)

result = subprocess.run(cmd, shell=True, check=True, capture_output=True)
# called read_region 3 times, but expect only one warning message on stderr
assert result.stderr.decode().count('slow-path') == 1
```

jakirkham (approved):

jakirkham:
Thanks Gigon! 🙏

Can you please resolve the merge conflicts here?

jakirkham:
@grlee77 would you have time to look at the conflicts?

gigony:
> @gigony would you have time to look at the conflicts?

I am sorry for the late response. I just rebased it.

Issue #262, opened 2022-04-11
Accessing metadata can cause segmentation fault!


drbeh:
The following code results in `Segmentation fault (core dumped)`:
```python
import numpy as np
from tifffile import imwrite
from cucim import CuImage

# Create image array
img_array = np.ones((128, 128, 3)).astype(np.uint8)
print(f'RGB image shape: {img_array.shape}')
img_array = np.concatenate([img_array, 255 * np.ones_like(img_array[..., 0])[...,np.newaxis]], axis=2)
print(f'RGBA image shape: {img_array.shape}')

filename = 'test_cucim.tiff'
imwrite(filename, img_array, shape=img_array.shape, tile=(16, 16))

obj = CuImage(filename)
print(obj.metadata)
```

while it works if:
- the image is RGB instead of RGBA, or
- access the metadata through the extracted region instead:
```python
img = obj.read_region()
print(img.metadata)
```
```
{'cucim': {'associated_images': [], 'channel_names': ['R', 'G', 'B', 'A'], 'coord_sys': 'LPS', 'dims': 'YXC', 'direction': [[1.0, 0.0, 0.0], [0.0, 1.0, 0.0], [0.0, 0.0, 1.0]], 'dtype': {'bits': 8, 'code': 1, 'lanes': 1}, 'ndim': 3, 'origin': [0.0, 0.0, 0.0], 'path': 'test_cucim.tiff', 'resolutions': {'level_count': 1, 'level_dimensions': [[0, 1065353216]], 'level_downsamples': [1.0], 'level_tile_sizes': [[0, 0]]}, 'shape': [128, 128, 4], 'spacing': [1.0, 1.0, 1.0], 'spacing_units': ['micrometer', 'micrometer', 'color'], 'typestr': '|u1'}}
```

gigony:
Thanks @drbeh for reporting this issue!

PR #294, opened 2022-05-25
Populate correct channel names for RGBA image

gigony:
Fix a segmentation fault raised when loading RGBA image and accessing metadata.

Addresses #262

### DIFF
--- a/cpp/plugins/cucim.kit.cuslide/src/cuslide/cuslide.cpp
+++ b/cpp/plugins/cucim.kit.cuslide/src/cuslide/cuslide.cpp
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2020-2021, NVIDIA CORPORATION.
+ * Copyright (c) 2020-2022, NVIDIA CORPORATION.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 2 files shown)


cpp/plugins/cucim.kit.cuslide/src/cuslide/cuslide.cpp
```diff
@@ -147,9 +147,22 @@ static bool CUCIM_ABI parser_parse(CuCIMFileHandle_ptr handle_ptr, cucim::io::fo
...
+    uint8_t n_ch = level0_ifd->samples_per_pixel();
+    std::pmr::vector<std::string_view> channel_names(&resource);
+    channel_names.reserve(n_ch);
+    if (n_ch == 3)
+    {
+        channel_names.emplace_back(std::string_view{ "R" });
+        channel_names.emplace_back(std::string_view{ "G" });
+        channel_names.emplace_back(std::string_view{ "B" });
+    }
+    else
```
grlee77, line 159: This `else` case also gets run for n_ch = 1 grayscale images, so as-before a single channel image will be listed as having one channel named "R" when printing the metadata from Python

We would still run into problems for any case with > 4 channels (or does TIFF not support that?)

How about using `else if (n_ch == 4)` here and then having a final else clause that just stores a dummy name (maybe "X"?) to all `n_ch` channels?

grlee77 (changes requested):
One question/suggestion, but otherwise this looks good. 

I appreciate the new test case. Do you think we should also have a version with a 2D grayscale image?

cpp/plugins/cucim.kit.cuslide/src/cuslide/cuslide.cpp
gigony: Thanks @grlee77 for the comment!

Didn't check 1-ch image case before. Since loaded image will have 4-ch images when it goes through a slow-path using libtiff, I also updated the output shape of the image.

grlee77: Thanks

jakirkham (approved):

jakirkham:
@gpucibot merge

jakirkham:
Thanks Gigon! 🙏

PR #308, opened 2022-06-07
fix RawKernel bug for canny filter when quantiles are used

grlee77:
This PR fixes a bug in `canny` that was leading to failures in `test_use_quantiles`. It also expands that test case via `pytest.parametrize` to cover additional dtypes

The changes to other test classes are merely refactoring to avoid use of `unittest.TestCase` so that `pytest.parametrize` can be used


grlee77:
closed this one and reopened as #310 targeting branch-22.08

PR #316, opened 2022-06-23
Stain extraction: use a less strict condition across channels when thresholding

grlee77:
I think we want to allow pixels with substantial attenuation in **any** of the channels rather than requiring substantial attenuation in **all** channels. The Macenko et. al. reference does not actually explain how the threshold was applied (min, max or mean across channels). However, qualitatively, results looked closer to expected with the `any` condition proposed here when I was working on a pending stain normalization example/demo.

I have also seen third party implementations where there is an RGB->LUV conversion and thresholding is done on the L channel rather than in luminance space. That seems reasonable and perhaps is more intuitive in choice of the threshold number (easier to think in fraction of brightness rather than absorbance units). However, making that change would be a bigger deviation from the publication and would involve some small additional overhead of the RGB->luminance conversion.

I have marked this as breaking as it will affect the estimated stain vectors produced by `stain_extraction_pca`.

### DIFF
--- a/python/cucim/src/cucim/core/operations/color/stain_normalizer.py
+++ b/python/cucim/src/cucim/core/operations/color/stain_normalizer.py
@@ -314,7 +314,7 @@ def stain_extraction_pca(image, source_intensity=240, alpha=1, beta=0.345,
     )
 
     # remove transparent pixels
-    absorbance = absorbance[:, cp.all(absorbance > beta, axis=0)]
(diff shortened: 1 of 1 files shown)


gigony (approved):
Looks good to me!

jakirkham (approved):

jakirkham:
@gpucibot merge

jakirkham:
Thanks Greg for the PR and Gigon for the review! 🙏

PR #317, opened 2022-06-24
Add a fast histogram-based median filter

### LINKED ISSUE #320: [FEA] fast histogram-based median implementation
For large filter kernel sizes, the median filter becomes inefficient. For integer-valued images, a histogram-based approach can give much better performance.

I am working on adding this in #317

grlee77:
### Background

closes #320

This PR adds a fast histogram-based median filter. It is based on a paper by Oded Green [1] as available in [`opencv_contrib`](https://github.com/opencv/opencv_contrib/blob/4.x/modules/cudafilters/src/cuda/median_filter.cu). It is a GPU-based adaptation of an earlier CPU-based publication [2].

This new functionality has been added to the existing `median` function and will be automatically used when possible (although the user can override the automated choice via a new keyword-only `algorithm` parameter.

The CUDA file [here](https://github.com/opencv/opencv_contrib/blob/4.x/modules/cudafilters/src/cuda/median_filter.cu) has a BSD 3-clause license up top, although the repository states that it is Apache 2.0 licensed, so that is likely out of date. We do already list OpenCV's extra modules and the Apache 2.0 license for it in our 3rdparty license folder.

### References

[1] O. Green, "Efficient Scalable Median Filtering Using Histogram-Based Operations," in IEEE Transactions on Image Processing, vol. 27, no. 5, pp. 2217-2228, May 2018, https://doi.org/10.1109/TIP.2017.2781375.

[2] S. Perreault and P. Hebert, "Median Filtering in Constant Time," in IEEE Transactions on Image Processing, vol. 16, no. 9, pp. 2389-2394, Sept. 2007, https://doi.org/10.1109/TIP.2007.902329.

### Improvements

There are a number of generalizations in the implementation here as compared to the one in OpenCV:
- supports odd-sized rectangular footprints as well as square ones
- supports int8, uint16 and int16 in addition to uint8
- the dtype used to store the histogram counts is adapted to the filter's footprint size
- handles many additional boundary extension modes via explicit pre-pad and post-crop (in principle could adapt the kernel for each case to avoid this overhead, but likely not worth the substantial additional effort)
- automatically switches between the sorting-based and histogram-based implementations depending on kernel size (sorting is much faster for very small kernels while the histogram-based approach is much faster for large ones)

### Notes/Limitations

The GPU I tested this on has a static shared memory limit of 48k per block. This means that the `HFine` shared memory array can exceed the available GPU memory once there are more than ~12000 gray levels in the histogram. So, for int16 or uint16, we I can only run the algorithm on images where the min/max range does not exceed that value. One fairly common case in medical imaging is DICOM images with 12-bit range (0-4095) that are stored at 16-bit int. This shared memory limit is compatible with this case.

### Performance

see summary in the next comment


grlee77:
## Benchmark results

Results are shown below for two different workstations and a comparison of 3 implementations on each:
- CPU: ITK (multithreaded using all available cores)
- GPU: CuPy's sorting-based median
- GPU: The histogram-based median from this PR

The top and bottom plots are the same thing except the bottom one is on the log scale. This makes it visible that the sorting based approach is faster for small footprint sizes. The histogram-based approach has a computation time that is nearly independent of the footprint size.

median_benchmark_two_system


## partitions parameter benchmark

Also in tuning the histogram-based approach, there is a `partitions` parameter that has a large effect on performance. Higher values up to approximately approximately 512 on the RTX A6000 system can improve performance, although minimal benefit is typically seen beyond half the number of image rows in smaller images.

Also, as can be seen in the tables, the working memory requirement is directly proportional to `partitions`. It is also directly proportional to the number of image columns as well as the number of histogram bins, so memory required for uint16 cases is much more than for uint8. For this reason, the image min/max are used in determining the histogram range for 16-bit inputs.


### image shape (512, 512), uint8

footprint shape | # partitions | extra working memory required (MB) | duration (s)
----------------|--------------|------------------------------------|-------------
(25, 25) | 1 | 0.2578125 | 0.3180733642578125
(25, 25) | 2 | 0.515625 | 0.15946286468505858
(25, 25) | 4 | 1.03125 | 0.07990570269132916
(25, 25) | 8 | 2.0625 | 0.040577709094898114
(25, 25) | 16 | 4.125 | 0.020654160146843896
(25, 25) | 32 | 8.25 | 0.01171679826825857
(25, 25) | 64 | 16.5 | 0.007104467095921955
(25, 25) | 128 | 33.0 | 0.004121047191567474
(25, 25) | 256 | 66.0 | 0.0028324458865001657
(25, 25) | 384 | 99.0 | 0.003215217651215746
(25, 25) | 512 | 132.0 | 0.0030073736240586123


### image shape (2048, 2048), uint8

footprint shape | # partitions | extra working memory required (MB) | duration (s)
----------------|--------------|------------------------------------|-------------
(25, 25) | 1 | 1.03125 | 5.17657177734375
(25, 25) | 2 | 2.0625 | 2.58887255859375
(25, 25) | 4 | 4.125 | 1.295036376953125
(25, 25) | 8 | 8.25 | 0.6912278849283854
(25, 25) | 16 | 16.5 | 0.40077769470214847
(25, 25) | 32 | 33.0 | 0.20581830596923828
(25, 25) | 64 | 66.0 | 0.10558810882568359
(25, 25) | 128 | 132.0 | 0.05465728460039411
(25, 25) | 256 | 264.0 | 0.030111923828125
(25, 25) | 384 | 396.0 | 0.025255038452148437
(25, 25) | 512 | 528.0 | 0.021404803558134695
(25, 25) | 768 | 792.0 | 0.023774978041648867
(25, 25) | 1024 | 1056.0 | 0.02653733614871376
(25, 25) | 1536 | 1584.0 | 0.03566272451711256
(25, 25) | 2048 | 2112.0 | 0.04358610033307756

### image shape (2048, 2048), uint16 with range [0, 4095]

footprint shape | # partitions | extra working memory required (MB) | duration (s)
----------------|--------------|------------------------------------|-------------
(25, 25) | 1 | 16.125 | 6.932091796875
(25, 25) | 2 | 32.25 | 3.493909423828125
(25, 25) | 4 | 64.5 | 2.0387840576171876
(25, 25) | 8 | 129.0 | 1.1526460571289063
(25, 25) | 16 | 258.0 | 0.5829215494791667
(25, 25) | 32 | 516.0 | 0.295801508585612
(25, 25) | 64 | 1032.0 | 0.1556665939331055
(25, 25) | 128 | 2064.0 | 0.08680433824327258
(25, 25) | 256 | 4128.0 | 0.05798199330843412
(25, 25) | 384 | 6192.0 | 0.05386503669193814
(25, 25) | 512 | 8256.0 | 0.05406031567709787
(25, 25) | 768 | 12384.0 | 0.0623766049194336
(25, 25) | 1024 | 16512.0 | 0.07788800430297853
(25, 25) | 1536 | 24768.0 | 0.0982440299987793
(25, 25) | 2048 | 33024.0 | 0.12620956166585287

python/cucim/src/cucim/skimage/filters/_median.py
```diff
@@ -79,5 +112,64 @@ def median(image, footprint=None, out=None, mode='nearest', cval=0.0,
...
+                out[:] = temp
+            else:
+                if out_dtype is not None:
+                    temp = temp.astype(out_dtype, copy=False)
+                out = temp
+            return out
+        except KernelResourceError as e:
+            # Fall back to sorting-based implementation if we encounter a
+            # resource limit (e.g. insufficient shared memory per block).
+            print(e)
```
jakirkham, line 167: Should this `print` still be here?

python/cucim/src/cucim/skimage/filters/_median.py
```diff
@@ -79,5 +112,64 @@ def median(image, footprint=None, out=None, mode='nearest', cval=0.0,
...
+                    except TypeError:
+                        raise TypeError(
+                            "out must be either a cupy.array or a valid input "
+                            "to cupy.dtype"
+                        )
+
+            # TODO: Can't currently pass an output array into _median_hist as a
+            #       new array currently needs to be created during padding.
+            temp = _median_hist(image, footprint, mode=mode, cval=cval,
+                                **algorithm_kwargs)
```
jakirkham, line 156: Would it make sense to turn this TODO into an issue?

jakirkham (commented):
Thanks Greg! 🙏

Had a couple questions above

python/cucim/src/cucim/skimage/filters/_median.py
grlee77: Good catch. I have now changed it to a warning instead. If the user wants to avoid the warning in the future they could then set the `algorithm='sorting'` kwarg to bypass trying the histogram-based algorithm.

python/cucim/src/cucim/skimage/filters/_median.py
grlee77: Sure, I opened #352 to mention this as well as potentially extending to 3D

grlee77:
I see there is an isolated test failure with 16-bit that I can't reproduce locally. I am tempted for now to only enable this histogram-based algorithm for 8-bit integer types.

jakirkham:
That seems reasonable. We can always relax it later after we fix issues with other types

python/cucim/src/cucim/skimage/filters/_median_hist.py
```diff
@@ -0,0 +1,733 @@
...
+
+# This preamble code is common across all generated kernels
+preamble_common = """
+
+__device__ void histogramAddAndSubCoarse(HIST_INT_T* H, const HIST_INT_T* hist_colAdd, const HIST_INT_T* hist_colSub){
+    int tx = threadIdx.x;
+    if (tx<HIST_SIZE_COARSE){
+        H[tx]+=hist_colAdd[tx]-hist_colSub[tx];
+    }
+}
```
jakirkham, line 95: Maybe it is worth pulling this code out into C files that we load and then compile?

python/cucim/src/cucim/skimage/filters/_median_hist.py
```diff
@@ -0,0 +1,733 @@
...
+                *count  = array[threadIndex];                          \\
+            }                                                          \\
+        }
+
+
+"""  # noqa
+
+
+# TODO: look into using CUB's Block-wise collectives (e.g. BlockScan)
+def _gen_preamble_median(hist_size_coarse, hist_size_fine):
```
jakirkham, line 180: Should we file an issue about this?

python/cucim/src/cucim/skimage/filters/tests/test_median.py
```diff
@@ -41,12 +56,167 @@ def test_selem_kwarg_deprecation(image):
...
+    'int_dtype, irange',
+    [
+        (cp.uint16, (0, 256)),
+        (cp.uint16, (0, 15)),
+        (cp.uint16, (128, 384)),
+        (cp.uint16, (0, 200)),
+        (cp.uint16, (0, 510)),
+        (cp.uint16, (500, 550)),
+        (cp.uint16, (0, 1024)),
+        # (cp.uint16, (0, 2048)),
```
jakirkham, line 152: Maybe we should mark as `xfail` instead of commenting? Would make it easier to see these issues and fix them later. Also may be worth filing an issue to track. This could be included in the `reason` message

```suggestion
        pytest.param(cp.uint16, (0, 2048), marks=pytest.mark.xfail(reason="<insert reason here>")),
```

python/cucim/src/cucim/skimage/filters/tests/test_median.py
```diff
@@ -41,12 +56,167 @@ def test_selem_kwarg_deprecation(image):
...
+        # (cp.int16, (150, 2048)),
+    ]
+)
+def test_median_hist_16bit_offsets(mode, footprint_shape, int_dtype, irange):
+    """Make sure 16-bit cases are robust to various value ranges"""
+    footprint = cp.ones(footprint_shape, dtype=bool)
+    rng = cp.random.default_rng(123)
+    shape = (350, 407)
+    if int_dtype == cp.uint16:
+        # test with 12-bit range stored in 16-bit integers (e.g. DICOM)
```
jakirkham, line 169: Possibly also worth an issue to track

jakirkham (commented):
Thanks Greg! 🙏

Generally this looks good. Noted a few things above that we might want to follow up on

jakirkham:
@gigony could you please look at this as well?

python/cucim/src/cucim/skimage/filters/tests/test_median.py
grlee77: If I put `xfail` then I will not be able to run locally as the failure doesn't occur for me. I have only seen these fail on CI.

python/cucim/src/cucim/skimage/filters/_median_hist.py
grlee77: That was the route I ended up going with the distance transform PR. We could do something similar here.

python/cucim/src/cucim/skimage/filters/tests/test_median.py
grlee77: added to #352

python/cucim/src/cucim/skimage/filters/_median_hist.py
grlee77: added to #352

python/cucim/src/cucim/skimage/filters/_median_hist.py
grlee77: Please see the change in [2608b2e](https://github.com/rapidsai/cucim/pull/317/commits/2608b2e9771ea4143f10ca8c29660cc0a19150ca) and [a364ee4](https://github.com/rapidsai/cucim/pull/317/commits/a364ee46e90bb16531fc3b0b1f21f4948413bb6c) where I moved all of the non-dynamically generated code into a separate .cu file.

python/cucim/src/cucim/skimage/filters/tests/test_median.py
jakirkham: We could do `mark.skip` instead

python/cucim/src/cucim/skimage/filters/_median_hist.py
jakirkham: Thanks Greg! 🙏

Think we may want to start doing this in more places going forward. Filed issue ( https://github.com/rapidsai/cucim/issues/375 ) about it

gigony (approved):
Thanks Greg! Looks good to me!

python/cucim/src/cucim/skimage/filters/tests/test_median.py
grlee77: Okay this is now using `mark.skip`! Thanks for pointing out the notation to do this viia `pytest.param`.

grlee77: I guess with xfail it doesn't actually cause the tests to fail if they pass, it just gets reported as XPASS in the summary. I think skip is fine, though as we ideally don't expect it to fail.

jakirkham: Great thanks Greg! 🙏

jakirkham (approved):

jakirkham:
@gpucibot merge

jakirkham:
Thanks Greg for the PR and Gigon for helping review! 🙏

PR #318, opened 2022-06-28
Add Euclidean distance transform for images/volumes

### LINKED ISSUE #319: [FEA] add euclidean distance transform
It will be useful to have an implementation of `scipy.ndimage.distance_transform_edt` (currently unimplemented in SciPy).

I am working on this in #318

grlee77:
### Overview

closes #319

This PR adds an implementation of `scipy.ndimage.distance_transform_edt`. For each foreground pixel in a binary image, this function computes the minimal Euclidean distance to reach a background pixel. This is a SciPy function rather than a scikit-image one so I have put it under `cucim.core` instead of `cucim.skimage`. Longer term we could move this upstream to CuPy, but I think we want to have a copy here so that we can use it in the near term to implement some of the missing scikit-image functionality. This function is used by the following functions in the scikit-image API, so this PR will help us implement these in future PRs:

`skimage.morphology.medial_axis` (**update**: see #342)
`skimage.segmentation.expand_labels` (**update**: see #341)
`skimage.segmentation.chan_vese` (**update**: see #343)
It is also used in examples related to `skimage.segmentation.watershed` and is needed for implementation of ITK's `SignedMaurerDistanceMapImageFilter` for [itk-cucim](https://github.com/InsightSoftwareConsortium/itk_cucim/issues/13).

The algorithm used here is adapted from the C++ code in the [PBA+ repository](https://github.com/orzzzjq/Parallel-Banding-Algorithm-plus) (MIT-licensed)

### extensions made to PBA+ kernel source
- larger sizes can be supported due to runtime code generation of the integer datatypes and encode/decode functions.

### current known limitations
- 2D and 3D only (this should cover the most common use cases)
- The `sampling` argument is not fully supported. This can likely be done with a bit of effort, but will require modifications to the CUDA kernels
- User-specified output arguments are not currently supported. We can potentially add this in the future
- The indices returned by `return_indices=True` are equally valid, but not always identical to those returned by SciPy. This is because there can be multiple indices with an identical distance, so which one gets chosen in that case is implementation-dependent.

### initial benchmarks relative to SciPy

Here `% true` is the percentage of the image that corresponds to foreground pixels. Even for fairly small images there is some benefit, with the benefit becoming two orders of magnitude at larger sizes.

shape      | % true |  cuCIM |  SciPy  | acceleration
-----------------|--------|---------|---------|--------------
(256, 256) | 5 | 0.00108 | 0.00353 | 3.25
(512, 512) | 5 | 0.00108 | 0.01552 | 14.42
(2048, 2048) | 5 | 0.00252 | 0.34434 | 136.86
(4096, 4096) | 5 | 0.00765 | 1.58948 | 207.88
(32, 32, 32) | 5 | 0.00103 | 0.00305 | 2.98
(128, 128, 128) | 5 | 0.00153 | 0.30103 | 196.26
(256, 256, 256) | 5 | 0.00763 | 3.17872 | 416.37
(384, 384, 384) | 5 | 0.02460 | 14.28779 | 580.89
(256, 256) | 50 | 0.00107 | 0.00430 | 4.01
(512, 512) | 50 | 0.00109 | 0.01878 | 17.30
(2048, 2048) | 50 | 0.00299 | 0.39304 | 131.60
(4096, 4096) | 50 | 0.00896 | 1.84686 | 206.19
(32, 32, 32) | 50 | 0.00102 | 0.00361 | 3.53
(128, 128, 128) | 50 | 0.00163 | 0.31657 | 194.66
(256, 256, 256) | 50 | 0.00914 | 3.35914 | 367.49
(384, 384, 384) | 50 | 0.03005 | 13.83219 | 460.30
(256, 256) | 95 | 0.00107 | 0.00344 | 3.22
(512, 512) | 95 | 0.00108 | 0.01638 | 15.17
(2048, 2048) | 95 | 0.00314 | 0.36996 | 117.81
(4096, 4096) | 95 | 0.00943 | 1.90475 | 202.05
(32, 32, 32) | 95 | 0.00102 | 0.00314 | 3.07
(128, 128, 128) | 95 | 0.00180 | 0.28843 | 159.80
(256, 256, 256) | 95 | 0.01073 | 3.23450 | 301.41
(384, 384, 384) | 95 | 0.03577 | 12.40526 | 346.82

### other comments
- can likely reduce memory overhead and improve performance a bit by refactoring some of the pre/post-processing code into elementwise kernels. (e.g. `encode3d`/`decode3d`, etc.)

- (JK) This may be able to leverage CuPy in the future ( https://github.com/cupy/cupy/issues/6919 )

### DIFF
--- a/LICENSE-3rdparty.md
+++ b/LICENSE-3rdparty.md
@@ -281,3 +281,9 @@ StainTools
   - https://github.com/Peter554/StainTools/blob/master/LICENSE.txt
 - Copyright: Peter Byfield
 - Usage: reference for stain color normalization algorithm
+
+PBA+
+- License: MIT License
+  - https://github.com/orzzzjq/Parallel-Banding-Algorithm-plus/blob/master/LICENSE
+- Copyright: School of Computing, National University of Singapore
+- Usage: PBA+ is used to implement the Euclidean distance transform.
--- a/python/cucim/src/cucim/core/operations/morphology/tests/test_distance_transform.py
+++ b/python/cucim/src/cucim/core/operations/morphology/tests/test_distance_transform.py
@@ -25,60 +25,34 @@ def assert_percentile_equal(arr1, arr2, pct=95):
     assert mismatch < pct_mismatch
 
 
-@pytest.mark.parametrize('block_params', [(1, 1, 1), None])
-@pytest.mark.parametrize('sampling', [None, (1.5, 1.5)])
-@pytest.mark.parametrize('shape', [(256, 256), (537, 236)])
-@pytest.mark.parametrize('density', [20, 50, 80])
-def test_distance_transform_edt_2d(shape, sampling, block_params, density):
-
-    kwargs_scipy = dict(sampling=sampling)
-    kwargs_cucim = dict(sampling=sampling, block_params=block_params)
-    img = binary_image(shape, pct_true=density)
-    out = distance_transform_edt(img, **kwargs_cucim)
-    expected = ndi_cpu.distance_transform_edt(cp.asnumpy(img), **kwargs_scipy)
-    cp.testing.assert_allclose(out, expected)
-
-
-@pytest.mark.parametrize('sx', list(range(32)))
-@pytest.mark.parametrize('sy', list(range(16)))
-def test_distance_transform_edt_2d_aniso_block_params(sx, sy):
-    """ensure default block_params is robust to anisotropic shape."""
-    shape = (128 + sx, 128 + sy)
-    img = binary_image(shape, pct_true=80)
-    out = distance_transform_edt(img)
-    expected = ndi_cpu.distance_transform_edt(cp.asnumpy(img))
-    cp.testing.assert_allclose(out, expected)
-
-
-@pytest.mark.parametrize('value', [0, 1, 3])
-def test_distance_transform_edt_2d_uniform_valued(value):
-    """ensure default block_params is robust to anisotropic shape."""
-    img = cp.full((64, 64), value, dtype=cp.uint8)
-    # ensure there is at least 1 pixel at background intensity
-    img[13, 13] = 0
-    out = distance_transform_edt(img)
-    expected = ndi_cpu.distance_transform_edt(cp.asnumpy(img))
-    cp.testing.assert_allclose(out, expected)
-
-
 @pytest.mark.parametrize('return_indices', [False, True])
 @pytest.mark.parametrize('return_distances', [False, True])
-@pytest.mark.parametrize('sampling', [None, (1.5, 1.5)])
-@pytest.mark.parametrize('shape', [(256, 256), (65, 128)])
-def test_distance_transform_edt_2d_returns(shape, sampling, return_distances,
-                                           return_indices):
+@pytest.mark.parametrize('shape, sampling',
+    [
+        ((256, 128), None),
+        ((384, 256), (1.5, 1.5)),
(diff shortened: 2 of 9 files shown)


grlee77:
The failure here is due to use of `math.lcm` in a helper function. Apparently this function was only recently introduced in Python 3.9, so will need to introduce a workaround for that.

LICENSE-3rdparty.md
```diff
@@ -281,3 +281,9 @@ StainTools
   - https://github.com/Peter554/StainTools/blob/master/LICENSE.txt
 - Copyright: Peter Byfield
 - Usage: reference for stain color normalization algorithm
+
+PBA+
```
gigony, line 285: Thank you for updating license texts! :)

gigony (approved):
I don't understand detailed algorithms so I feel I am not capable of reviewing this patch in detail.
Overall, it looks good to me.

python/cucim/src/cucim/core/operations/morphology/tests/test_distance_transform.py
```diff
@@ -0,0 +1,147 @@
...
+    shape = (16 + sz, 32 + sy, 48 + sx)
+    img = binary_image(shape, pct_true=80)
+    out = distance_transform_edt(img)
+    expected = ndi_cpu.distance_transform_edt(cp.asnumpy(img))
+    cp.testing.assert_allclose(out, expected)
+
+
+@pytest.mark.parametrize('ndim', [1, 4, 5])
+def test_distance_transform_edt_unsupported_ndim(ndim):
+    """ensure default block_params is robust to anisotropic shape."""
```
grlee77, line 135:
```suggestion
```

python/cucim/src/cucim/core/operations/morphology/tests/test_distance_transform.py
```diff
@@ -0,0 +1,147 @@
...
+    with pytest.raises(NotImplementedError):
+        distance_transform_edt(cp.zeros((8,) * ndim))
+
+
+@pytest.mark.skip(reason="excessive memory requirement")
+def test_distance_transform_edt_3d_int64():
+    """ensure default block_params is robust to anisotropic shape."""
```
grlee77, line 142:
```suggestion
```

grlee77:
> I don't understand detailed algorithms so I feel I am not capable of reviewing this patch in detail.
> Overall, it looks good to me.

Thanks @gigony. I think the kernels are pretty well tested both here and in the original PBA+ repo, so probably doesn't need to much review in the details of the kernels in the `.h` files. I can hopefully revisit adding support for non-uniform values for `sampling` in a follow-up PR (for a future release).

This PR is a bit unique for cuCIM in that I kept the bulk of the kernel code in the form of CUDA C++ header files so that it is easier to view than if it was stored within a Python string. The file gets loaded into a string for purposes of runtime compilation, but keeping it in a separate file makes it easier to view with proper code highlighting either on GitHub or in an IDE.

jakirkham:
Made an edit to the OP to xref the related upstream CuPy issue. Hope that is ok 🙂

jakirkham (approved):
Thanks Greg! 🙏

Would suggest converting any TODOs into issues so we can track them better

jakirkham:
@gpucibot merge

jakirkham:
Filed upstream issue ( https://github.com/cupy/cupy/issues/7400 ) about moving this to CuPy

PR #323, opened 2022-06-28
expand cucim.skimage.segmentation benchmarks

grlee77:
I noticed that benchmark coverage for the segmentation functions was missing several functions. This PR adds them. 

Unfortunately functions in the segmentation module take different mixtures of image and label image inputs so there are multiple different benchmarking classes required to set up appropriate inputs. The appropriate benchmarking class is chosen based on the function name.

### DIFF
--- a/benchmarks/skimage/cucim_segmentation_bench.py
+++ b/benchmarks/skimage/cucim_segmentation_bench.py
@@ -3,6 +3,8 @@ import os
 import pickle
 
 import cucim.skimage
+import cucim.skimage.data
+import cucim.skimage.exposure
 import cucim.skimage.segmentation
 import cupy as cp
 import numpy as np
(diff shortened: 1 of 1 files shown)


benchmarks/skimage/cucim_segmentation_bench.py
```diff
@@ -187,5 +291,10 @@ def set_args(self, dtype):
 fbase = os.path.splitext(pfile)[0]
 all_results.to_csv(fbase + ".csv")
 all_results.to_pickle(pfile)
-with open(fbase + ".md", "wt") as f:
-    f.write(all_results.to_markdown())
+try:
+    import tabulate
```
jakirkham, line 295: Should we add this dependency somewhere?

grlee77: Not sure where the best place would be? We could add a `requirements-bench.txt` or something, but we should not add either `tabulate` or `pandas` to the main package dependecies as they aren't used outside of the benchmarks. We don't distribute these benchmark files with the conda or PyPI packages.

jakirkham: Yeah we could do that.

Another option would be to add it to [this environment file]( https://github.com/rapidsai/cucim/blob/branch-22.08/conda/environments/env.yml ). We can comment sections and specify these are benchmark requirements for clarity.

grlee77: We do have a `requirements-test.txt` within the python/cucim folder which has various test requirements that are not listed in that `env.yml`. I can just add a `requirements-bench.txt` to the changes in #324

jakirkham: Maybe we should include these in the `env.yml` as well

PR #327, opened 2022-07-05
Protect against obscure divide by zero error in edge case of `normalize_colors_pca`

grlee77:
This PR updates `normalize_colors_pca` to raise a more helpful error message in the case that there is only a single pixel above the specified `beta` threshold. Currently this leads to a not very helpful `DivideByZeroError` because the degrees of freedom during an internal covariance calculation becomes zero in this case.

This does also change the error type when there are no pixels above the threshold from `ValueError` to `RuntimeError` which seems more appropriate to me.

### DIFF
--- a/python/cucim/src/cucim/core/operations/color/stain_normalizer.py
+++ b/python/cucim/src/cucim/core/operations/color/stain_normalizer.py
@@ -13,7 +13,6 @@
 # limitations under the License.
 
 import math
-import warnings
 from typing import Union
 
 import cupy as cp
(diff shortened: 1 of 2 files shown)


jakirkham (approved):
Thanks Greg! 🙏

jakirkham (approved):

jakirkham:
@gpucibot merge

jakirkham:
@gpucibot merge

jakirkham:
Thanks Greg! 🙏

PR #338, opened 2022-07-26
Add cucim.skimage.feature.match_descriptors

### LINKED ISSUE #193: [FEA] gpu enabled match_descriptors
I've heard from some folks that they would like to see an implementation of [match_descriptors](https://scikit-image.org/docs/dev/api/skimage.feature.html?highlight=match_descriptor#skimage.feature.match_descriptors) in cuCIM.  This might be challenging as I don't think [cdist](https://docs.scipy.org/doc/scipy/reference/generated/scipy.spatial.distance.cdist.html) is implemented in CuPy yet.

xref: https://github.com/scikit-image/scikit-image/blob/main/skimage/feature/match.py#L5

grlee77:
closes #193

This PR adds `cucim.skimage.feature.match_descriptors`. It is a very straightforward adaptation of the scikit-image code, only substituting numpy->cupy. The only differences of note are:
- when new enough CuPy and pylibraft are not available it warns and falls back to `cdist` on the CPU
- test cases involving BRIEF temporarily involve round trip to the host until we implement BRIEF here (looks not too hard)

This PR also removes the deprecated `masked_register_translation` and `register_translation` (these were moved to cucim.skimage.registration.phase_cross_correlation`) from the `feature` module. Removing those was missed when previously updating the API here to match scikit-image 0.19.

### DIFF
--- a/python/cucim/src/cucim/skimage/feature/__init__.py
+++ b/python/cucim/src/cucim/skimage/feature/__init__.py
@@ -6,38 +6,9 @@ from .corner import (corner_foerstner, corner_harris, corner_kitchen_rosenfeld,
                      corner_peaks, corner_shi_tomasi, hessian_matrix,
                      hessian_matrix_det, hessian_matrix_eigvals, shape_index,
                      structure_tensor, structure_tensor_eigenvalues)
+from .match import match_descriptors
 from .peak import peak_local_max
 from .template import match_template
-
-
-@deprecated(
-    alt_func="cucim.skimage.registration.phase_cross_correlation",
-    removed_version="0.19",
-)
-def masked_register_translation(
-    src_image, target_image, src_mask, target_mask=None, overlap_ratio=0.3
-):
-    from ..registration import phase_cross_correlation
-
-    return phase_cross_correlation(
-        src_image,
-        target_image,
-        reference_mask=src_mask,
-        moving_mask=target_mask,
-        overlap_ratio=overlap_ratio,
-    )
-
-
-@deprecated(
-    alt_func="cucim.skimage.registration.phase_cross_correlation",
-    removed_version="0.19",
-)
-def register_translation(
-    src_image, target_image, upsample_factor=1, space="real", return_error=True
-):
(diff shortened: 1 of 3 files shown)


grlee77:
The acceleration provided here will depend on the number and size of the descriptors. Running scikit-image's SIFT on the small `astronaut` image results in 1234 descriptors each of size 128. For that size of input, the relative acceleration on the GPU is quite good.

When cross_check=False, max_distance=np.inf, max_ratio = 1.0
**Acceleration = 333.44040448425943**

When cross_check=True, max_distance=np.inf, max_ratio = 1.0
**Acceleration = 138.47554030822292**

When cross_check=False, max_distance=100.0, max_ratio < 1.0
**Acceleration = 58.34578386444441**

In the latter two cases there are progressively more logical comparisons and indexing operations which end up reducing the overall acceleration factor as compared to the first case without any additional checks.

jakirkham:
Moved to 22.10

There's not a project board for this yet. So moved to feature planning for now. Can update that later

ajschmidt8:
Removing `ops-codeowners` from the required reviews since it doesn't seem there are any file changes that we're responsible for. Feel free to add us back if necessary.

jakirkham:
Yeah I accidentally got 21.10 the first time 🤦‍♂️ Sorry for the noise 😞

Should be fixed now 🍀

grlee77:
This PR now has that same issue we saw previously where the `distance_transform_edt` import fails even though the files are in this branch.

```
src/cucim/core/operations/morphology/tests/test_distance_transform.py:8: in <module>
    from cucim.core.operations.morphology import distance_transform_edt
E   ModuleNotFoundError: No module named 'cucim.core.operations.morphology'
```


I need to look into how we are doing the install when testing on CI. Perhaps the files are there during build but are getting omitted somehow during the package install stage?

grlee77:
rerun tests

gigony (approved):
Looks good to me. Thank you Greg!

gigony:
@gpucibot merge

PR #340, opened 2022-07-26
Performance optimizations to morphological segmentation functions

grlee77:
This PR is purely a performance refactor without a change in behavior. The main changes to `morphological_chan_vese` are:

- Use a simplified version of `cupy.gradient` that does not promote all integer types to float64 and omits a couple of division operations by assuming unit spacing. Currently we don't have any places where non-uniform spacing or non first-order edges are needed, but we can always still use `cupy.gradient` for those if they come up in the future. 

- apply kernel fusion via `cupy.fuse()` decorators.

- avoid repeated transfer of footprints to the GPU and reuse the same workspace array for repeated `_curveop` calculations.

Overall, these changes give around 1/3 reduction in run-time for `morphological_chan_vese`.

All other places using `cp.gradient` were also changed to use the simplified `gradient` defined here. These include:
`cucim.skimage.features.hessian_matrix`
`cucim.skimage.features.multiscale_basic_features`
`cucim.skimage.registration.optical_flow_tvl1`

### DIFF
--- a/python/cucim/src/cucim/skimage/_shared/_gradient.py
+++ b/python/cucim/src/cucim/skimage/_shared/_gradient.py
@@ -11,7 +11,7 @@ import cupy
 from cucim.skimage._shared.utils import _supported_float_type
 
 
-def gradient(f, axis=None):
+def gradient(f, axis=None, output_as_array=False):
     """Return the gradient of an N-dimensional array.
 
     The gradient is computed using second order accurate central differences
@@ -27,6 +27,7 @@ def gradient(f, axis=None):
             (axis = None) is to calculate the gradient for all the axes of the
             input array. axis may be negative, in which case it counts from the
             last to the first axis.
+        output_as_array
 
     Returns:
         gradient (cupy.ndarray or list of cupy.ndarray): A set of ndarrays
@@ -50,8 +51,6 @@ def gradient(f, axis=None):
     # use central differences on interior and one-sided differences on the
     # endpoints. This preserves second order-accuracy over the full domain.
 
-    outvals = []
-
     # create slice objects --- initially all are [:, :, ..., :]
     slice1 = [slice(None)] * ndim
     slice2 = [slice(None)] * ndim
@@ -68,6 +67,12 @@ def gradient(f, axis=None):
             f = f.astype(float_dtype)
         otype = float_dtype
 
+    if output_as_array:
+        out = cupy.empty((ndim,) + f.shape, dtype=otype)
+        outvals = out
+    else:
+        outvals = []
+
     for axis in axes:
         if f.shape[axis] < 2:
             raise ValueError(
@@ -75,7 +80,8 @@ def gradient(f, axis=None):
                 "at least 2 elements are required."
             )
         # result allocation
-        out = cupy.empty_like(f, dtype=otype)
+        if not output_as_array:
+            out = cupy.empty_like(f, dtype=otype)
 
         # Numerical differentiation: 2nd order interior
         slice1[axis] = slice(1, -1)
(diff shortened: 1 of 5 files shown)


jakirkham:
Thanks for working on this Greg! 🙏

Am curious would it be possible for us to promote some cases to `float32` ourselves and then call `cupy.gradient` to avoid the promotion to `float64`? Or does it still promote in that case as well?

python/cucim/src/cucim/skimage/_shared/_gradient.py
```diff
@@ -0,0 +1,111 @@
...
+        slice4[axis] = slice(2, None)
+
+        out[tuple(slice1)] = (f[tuple(slice4)] - f[tuple(slice2)]) / 2.0
+
+        # Numerical differentiation: 1st order edges
+        slice1[axis] = 0
+        slice2[axis] = 1
+        slice3[axis] = 0
+        # 1D equivalent -- out[0] = (f[1] - f[0]) / (x[1] - x[0])
+        out[tuple(slice1)] = f[tuple(slice2)] - f[tuple(slice3)]
```
grlee77, line 93: Note: by supporting only unit spacing this version has removed division by the distance on this line as well as on line 99.

grlee77:
> Am curious would it be possible for us to promote some cases to float32 ourselves and then call cupy.gradient to avoid the promotion to float64? Or does it still promote in that case as well?

That is a good idea in that we could just have a thin wrapper to the CuPy functions. The other optimization that is implemented here is the one in this comment: https://github.com/rapidsai/cucim/pull/340/files#r932285232

I haven't benchmarked just that change in isolation, but it should help a bit.

One other reason to potentially keep our own copy is to further extend this one to allow writing a set of `n` gradients to a single `ndim + 1` dimensional output array instead of a tuple of `n` arrays. That would avoid some places where I have had to call essentially getting the equivalent of the following directly from a single `gradient` call:
```
grads = cp.stack(cp.gradient(x), axis=0)
```

grlee77:
It is likely worth at some point creating a custom kernel for computing all gradient terms in a single pass over the array (ideally contributing that to CuPy itself).

jakirkham:
> > Am curious would it be possible for us to promote some cases to float32 ourselves and then call cupy.gradient to avoid the promotion to float64? Or does it still promote in that case as well?
> 
> That is a good idea in that we could just have a thin wrapper to the CuPy functions. The other optimization that is implemented here is the one in this comment: https://github.com/rapidsai/cucim/pull/340/files#r932285232

Thanks for the clarification. For some reason the link is just taking me to the file with no selection. Is there a particular line (or lines) that I should be looking at?

grlee77:
> Thanks for the clarification. For some reason the link is just taking me to the file with no selection. Is there a particular line (or lines) that I should be looking at?

There were just a couple of lines that are now
```Python
out[out_sl] = f[tuple(slice2)] - f[tuple(slice3)]
```
whereas in cupy the difference is divided by dx
```Python
out[out_sl] = (f[tuple(slice2)] - f[tuple(slice3)]) / dx
```

jakirkham:
> One other reason to potentially keep our own copy is to further extend this one to allow writing a set of `n` gradients to a single `ndim + 1` dimensional output array instead of a tuple of `n` arrays. That would avoid some places where I could get the equivalent of the following directly from a single `gradient` call:
> 
> ```python
> grads = cp.stack(cp.gradient(x), axis=0)
> ```
> 
> **note:** currently looks like only the functions in registration/_optical_flow.py explicitly stack the gradient like this

Following up on this point. Agree this is a reasonable concern regarding memory pressure. That said, maybe we should discuss with others on how we can improve this implementation across libraries to alleviate the need to have this code ( https://github.com/data-apis/array-api/issues/468 ).

jakirkham (approved):

jakirkham:
@gpucibot merge

jakirkham:
Thanks Greg! 🙏

PR #357, opened 2022-07-28
Allow CuPy 11

jakirkham:
Allows cuCIM to be installed with CuPy 11.

xref: https://github.com/rapidsai/integration/pull/508

PR #362, opened 2022-08-01
Revert "Allow CuPy 11"

galipremsagar:
Reverts rapidsai/cucim#357

### DIFF
--- a/conda/recipes/cucim/meta.yaml
+++ b/conda/recipes/cucim/meta.yaml
@@ -29,7 +29,7 @@ requirements:
     - python {{ python_version }}.*
     - libcucim {{ version }}.*
     - click
-    - cupy >=9,<12.0.0a0
+    - cupy >=9,<11.0.0a0
     - numpy 1.19
     - scipy
     - scikit-image >=0.18.1,<0.20.0a0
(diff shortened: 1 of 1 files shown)


quasiben:
I think cuCIM wants to use CuPy 11 even though RAPIDS need to exclude it for now.  This has happened in the past where cuCIM needs to specifically add CuPy 11 to the CI infra to properly test

ajschmidt8 (approved):

ajschmidt8:
> I think cuCIM wants to use CuPy 11 even though RAPIDS need to exclude it for now. This has happened in the past where cuCIM needs to specifically add CuPy 11 to the CI infra to properly test

@quasiben, will `cucim` be using CuPy features that are exclusively in CuPy 11? If so, I think this will cause issues with the `rapids` package on Anaconda ([src](https://github.com/rapidsai/integration/blob/53e92859fc0a5d6c84d5fc50bbf171731ad05d74/conda/recipes/rapids/meta.yaml#L39-L64)) if `cucim` needs CuPy 11, but the rest of RAPIDS can only use CuPy 10.

jakirkham:
Discussed with @grlee77, we are going to bump the feature that needed CuPy 11 to 22.10. So sticking to CuPy 10 for 22.08 should be ok for us

jakirkham (approved):

jakirkham:
@gpucibot merge

jakirkham:
Thanks all! 🙏

PR #355, opened 2022-07-28
more efficient separable convolution

grlee77:
closes #321 

This PR adds a different implementation of separable convolution that was adapted from `opencv_contrib`. It is not yet ready for review.

Still needs:
- [x] support other boundary modes
- [x] tests
- [x] ~extension to nD~  extension to 3D
- [x] wrappers for existing filters to call this one instead of CuPy's

The key changes in this PR are the new files:
`skimage/filters/_separable_filtering.py`
`skimage/filters/tests/test_separable_filtering.py`

It is based on the approach taken in OpenCV-contrib's [row](https://github.com/opencv/opencv_contrib/blob/4.6.0/modules/cudafilters/src/cuda/row_filter.hpp) and [column]([https://github.com/opencv/opencv_contrib/blob/4.6.0/modules/cudafilters/src/cuda/column_filter.hpp] ) filters, but also supports:
- 3D
- additional dtypes (e.g. complex64)
- all boundary modes from SciPy
- not restricted to kernel size <= 32
- casting behavior to the output matches SciPy rather than OpenCV conventions.
 
A simpler version of the same approach has long been in the CUDA samples [convolutionSeparable.cu example](https://github.com/NVIDIA/cuda-samples/blob/master/Samples/2_Concepts_and_Techniques/convolutionSeparable/convolutionSeparable.cu). The basic idea is:
1.) First stage loads the current patch of the image and its boundaries into shared memory
2.) After synchronization, convolution is performed on the shared memory array.

A lot of CuPy's ndimage code is vendored here for the following reasons:
- We need any `cupyx.scipy.ndimage` functions we call, like `gaussian_filter`, `uniform_filter`, etc. to dispatch to the new convolution implementation when possible.
- the `_get_weights_dtype` utility was changed to promote 8 and 16-bit integers to float32 instead of float64 during convolutions.
- the `_get_output` utility was changed to allocate the output arrays as `empty` rather than `zeros` which is more efficient.
- the `_run_1d_filters` utility was improved so that it avoids an extra copy when the number of filters is even

I will submit PRs for these to CuPy. The `_get_output` change in particular also impacts other morphology and interpolation functions that we use in cuCIM, so I have vendored those here as well. For reference regarding these non-filtering changes, there is modest performance improvement. Grayscale `erosion` on a 4k image became 10% faster and `resize` of a 4k image to HD became 20% faster.

[...]

PR #374, opened 2022-08-02
Drop unneeded code

jakirkham:
cc @grlee77


grlee77:
Actually we do need those includes for CUDA as well, I think it is only the comment line that was accidentally kept.

I did already fix this yesterady in #355 commit https://github.com/rapidsai/cucim/pull/355/commits/e59badbefdc0708438387b657b503bc8b9e98ab2 of #355, though. Given that I think we can close this issue?

jakirkham:
Sounds good

PR #386, opened 2022-08-05
Add missing imports of euler_number and perimeter_crofton

grlee77:
These functions had been implemented, but were missing from the `__init__.py` in `skimage.measure`. This was not caught because the test cases had imported them from the private submodule!


jakirkham (approved):

jakirkham:
@gpucibot merge

grlee77:
Currently users could only obtain these only via import from a private submodule:
```
from cucim.skimage.measure._regionprops import euler_number, perimeter_crofton
```
while the fix here makes it available via
```
```
from cucim.skimage.measure import euler_number, perimeter_crofton
```

jakirkham:
Thanks Greg! 🙏

PR #393, opened 2022-08-22
Euclidean distance transform: fix bad choice of block parameters

### LINKED ISSUE #392: [BUG] distance_transform_edt gives bad output for 2D images > 1024 in size on either axis
**Describe the bug**
There is currently a bug in the 2D implementation of `distance_transform_edt` where certain lines of the output have incorrect values due to a bad default choice of internal CUDA grid/block parameters at this size.  Fortunately, the 3D distance transform does not suffer from the same issue.

A temporary workaround in release 22.08.00 is to manually specify the `block_params` argument and set the first parameter to 1. This is smaller than needed for optimal performance, but will ensure a correct result. 

In general for 2D images where one axis is larger than 1024, please set `block_params=(1, 32, 2)`. Other than as a workaround for this bug or in case of advanced performance profiling use cases, we recommend leaveling `block_params=None` to allow internal choice of these kernel launch parameters.

a minimal example
```py
import math
import cupy as cp

import matplotlib.pyplot as plt

from cucim.core.operations.morphology import distance_transform_edt

shape = (1200, 1200)
size = math.prod(shape)
ntrue = .0005* size
p_true =  ntrue / size
p_false = 1 - p_true

cp.random.seed(123)
image = cp.random.choice([0, 1], size=shape, p=(p_false, p_true))

distances, coords = distance_transform_edt(
    image == 0, return_distances=True, return_indices=True
)
plt.figure()
plt.imshow(distances.get(), vmin=0, vmax=116)
plt.show()
```
gives this `distances` output where the bright horizontal lines are invalid distances
edt_bad

whereas calling with
```py
distances, coords = distance_transform_edt(
    image == 0, return_distances=True, return_indices=True, block_params(1, 32, 2)
)
```
gives the correct result:
edt_good

**Expected behavior**
no artifacts with the default choice of `block_params`

cuCIM v22.08.,00

**Additional context**

grlee77:
closes #392

This PR fixes #392 and also makes it more friendly for use with user-provided `block_params`. In general, most users should not be providing that argument, but it can be used to compare different settings for performance optimization. In case of user-provided `block_params`, the implementation now automatically pad the shape to an appropriate least common multiple of the warp_size and the `m1`, `m2` and `m3` block parameters.

More extensive unit tests over a range of image sizes and `block_params` settings are now implemented.

### DIFF
--- a/python/cucim/src/cucim/core/operations/morphology/_pba_2d.py
+++ b/python/cucim/src/cucim/core/operations/morphology/_pba_2d.py
@@ -154,7 +154,7 @@ def _pba_2d(arr, sampling=None, return_distances=True, return_indices=False,
     if block_params is None:
         padded_size = math.ceil(max(arr.shape) / block_size) * block_size
 
-        # should be <= size / 64. sy must be a multiple of m1
+        # should be <= size / block_size. sy must be a multiple of m1
         m1 = padded_size // block_size
         # size must be a multiple of m2
         m2 = max(1, min(padded_size // block_size, block_size))
@@ -162,23 +162,28 @@ def _pba_2d(arr, sampling=None, return_distances=True, return_indices=False,
         m2 = 2**math.floor(math.log2(m2))
         if padded_size % m2 != 0:
             raise RuntimeError("error in setting default m2")
-        # should be <= 64. image size must be a multiple of m3
         m3 = min(min(m1, m2), 2)
     else:
         if any(p < 1 for p in block_params):
             raise ValueError("(m1, m2, m3) in blockparams must be >= 1")
-        m1, m2, m3 = map(int, block_params)
+        m1, m2, m3 = block_params
         if math.log2(m2) % 1 > 1e-5:
             raise ValueError("m2 must be a power of 2")
         multiple = lcm(block_size, m1, m2, m3)
         padded_size = math.ceil(max(arr.shape) / multiple) * multiple
 
     if m1 > padded_size // block_size:
-        raise ValueError("m1 too large. must be <= arr.shape[0] // 32")
+        raise ValueError(
+            f"m1 too large. must be <= padded arr.shape[0] // {block_size}"
+        )
     if m2 > padded_size // block_size:
-        raise ValueError("m2 too large. must be <= arr.shape[1] // 32")
+        raise ValueError(
+            f"m2 too large. must be <= padded arr.shape[1] // {block_size}"
+        )
     if m3 > padded_size // block_size:
-        raise ValueError("m3 too large. must be <= arr.shape[1] // 32")
+        raise ValueError(
+            f"m3 too large. must be <= padded arr.shape[1] // {block_size}"
+        )
     for m in (m1, m2, m3):
         if padded_size % m != 0:
             raise ValueError(
--- a/python/cucim/src/cucim/core/operations/morphology/_pba_2d.py
+++ b/python/cucim/src/cucim/core/operations/morphology/_pba_2d.py
@@ -1,7 +1,39 @@
+import functools
 import math
+import numbers
 import os
 
+
 import cupy
+
+try:
+    # math.lcm was introduced in Python 3.9
+    from math import lcm
+except ImportError:
+
+    """Fallback implementation of least common multiple (lcm)
+
+    TODO: remove once minimum Python requirement is >= 3.9
+    """
+
+    def _lcm(a, b):
+        return abs(b * (a // math.gcd(a, b)))
+
+    @functools.lru_cache()
+    def lcm(*args):
+        nargs = len(args)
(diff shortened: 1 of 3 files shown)


python/cucim/src/cucim/core/operations/morphology/_pba_2d.py
```diff
@@ -1,8 +1,40 @@
...
+    from math import lcm
+except ImportError:
+
+    """Fallback implementation of least common multiple (lcm)
+
+    TODO: remove once minimum Python requirement is >= 3.9
+    """
+
+    def _lcm(a, b):
+        return abs(b * (a // math.gcd(a, b)))
```
jakirkham, line 20: NumPy also has an `lcm` implementation FWIW

python/cucim/src/cucim/core/operations/morphology/_pba_2d.py
```diff
@@ -1,8 +1,40 @@
...
+
+    @functools.lru_cache()
+    def lcm(*args):
```
jakirkham, line 23: Curious what the motivation is for caching here

python/cucim/src/cucim/core/operations/morphology/_pba_2d.py
```diff
@@ -119,27 +151,34 @@ def _pba_2d(arr, sampling=None, return_distances=True, return_indices=False,
...
             raise RuntimeError("error in setting default m2")
-
-        # should be <= 64. texture size must be a multiple of m3
+        # should be <= 64. image size must be a multiple of m3
         m3 = min(min(m1, m2), 2)
     else:
-        m1, m2, m3 = block_params
+        if any(p < 1 for p in block_params):
+            raise ValueError("(m1, m2, m3) in blockparams must be >= 1")
+        m1, m2, m3 = map(int, block_params)
```
jakirkham, line 170: Are these allowed to be `float`s or something? Should we do a type check?

jakirkham (commented):
Thanks Greg! 🙏

Had a couple comments below

python/cucim/src/cucim/core/operations/morphology/_pba_2d.py
```diff
@@ -119,27 +151,34 @@ def _pba_2d(arr, sampling=None, return_distances=True, return_indices=False,
...
         # size must be a multiple of m2
-        m2 = max(1, min(padded_size // block_size, 32))
+        m2 = max(1, min(padded_size // block_size, block_size))
         # m2 must also be a power of two
         m2 = 2**math.floor(math.log2(m2))
         if padded_size % m2 != 0:
             raise RuntimeError("error in setting default m2")
-
-        # should be <= 64. texture size must be a multiple of m3
+        # should be <= 64. image size must be a multiple of m3
```
gigony, line 165: Do we need  a statement for checking the value is `<=64` somewhere?

gigony (approved):
Looks good to me besides John's comments and a minor question.

python/cucim/src/cucim/core/operations/morphology/_pba_2d.py
grlee77: Okay, I didn't know that. This way is not quite as fast as `math.lcm` for scalar `a`, `b` but still faster than `numpy.lcm` in that case.

python/cucim/src/cucim/core/operations/morphology/_pba_2d.py
grlee77: When using `math.lcm` directly in Python 3.8+ it is much faster (~150 ns vs. 2.8 µs for this fallback) so the cache was just to avoid that overhead prior to kernel launch. I think in practice it is not very important and can remove it if you prefer.

I figured often the (m1, m2, m3) block parameters would usually be the same. This function is not used at all when `block_params` is left at the default of None, though.

python/cucim/src/cucim/core/operations/morphology/_pba_2d.py
grlee77: No, they definitely should be integers. I am not sure why I had that explicit `map` call. I can just remove it.

python/cucim/src/cucim/core/operations/morphology/_pba_2d.py
grlee77: I think that was an outdated/stray comment and will just remove it. The check should be relative to  "block_size" as in the check shortly below that:
```py
    if m3 > padded_size // block_size:
        raise ValueError("m3 too large. must be <= arr.shape[1] // 32")
```

I will update that string to use `block_size` instead of 32 in case we change block_size in the future.

grlee77:
Thanks for reviewing, I think I have addressed the comments.

jakirkham (approved):

PR #403, opened 2022-09-01
Forward-merge branch-22.08 to branch-22.10 [skip gpuci]

jakirkham:
Fix conflicts with the bot generated forward-merger ( https://github.com/rapidsai/cucim/pull/391 ).

Issue #397, opened 2022-08-30
Two READMEs & CHANGELOGs?


jakirkham:
We have a CHANGELOG & README at the top-level and for Python. Maybe we should consolidate these into a single CHANGELOG & README?

* https://github.com/rapidsai/cucim/blob/branch-22.10/CHANGELOG.md
* https://github.com/rapidsai/cucim/blob/branch-22.10/python/cucim/CHANGELOG.md
* https://github.com/rapidsai/cucim/blob/branch-22.10/README.md
* https://github.com/rapidsai/cucim/blob/branch-22.10/python/cucim/README.md

jakirkham:
Addressing in PR ( https://github.com/rapidsai/cucim/pull/415 )

PR #404, opened 2022-09-02
update pypi CHANGELOG.md for releases 22.08.00 and 22.08.01

grlee77:


jakirkham:
Think this got resolved by the forward-merge PR ( https://github.com/rapidsai/cucim/pull/403 ), but please double check

grlee77:
That content is mostly the same, but this is a second copy of the CHANGELOG.md that is under the `python/cucim` folder and gets included in the `long_description` in `setup.py` so that it shows up on PyPI. I didn't realize we even had this second changelog when I did the 22.08.00 release, so it was not updated at that time. That is why both 22.08.00 and 22.08.01 changes are both here.

It seems we could migrate to just a single file going forward. (as for the README issue you recently raised in #397)

grlee77:
This is where these second copies are used:
https://github.com/rapidsai/cucim/blob/90fc72aa149b2bd2fa89622e2fb5f0ab5010b596/python/cucim/setup.py#L33-L37

jakirkham:
Oh missed that. Thanks for clarifying Greg 🙏

Yeah think tracking these duplicates down and coming up with a way to consolidate should save us additional maintenance work in the future

jakirkham (approved):

jakirkham:
@gpucibot merge

jakirkham:
Thanks Greg! 🙏

PR #406, opened 2022-09-06
Improve performance of Euclidean distance transform

grlee77:
This PR gives a further 2-4x speedup to the 2D and 3D distance transform functions by creating dedicated elementwise CUDA kernels for the pre- and post-processing steps. These include, for example, packing and upacking the 3D coordinates into a single integer. This resolves various "TODO" comments that were in the source, but I don't think we have an issue open for this.

Aside from the performance benefit there is also a substantial reduction in memory usage for the default case of `return_distance=True, return_indices=False` as in this case the explicit arrays of indices never need to be created.

# Benchmarks

## Implementation from release 22.08.01

|      shape      | return dist. | return ind. | SciPy  | ITK | cuCIM  | accel. (CuPy vs SciPy) | accel. (CuPy vs ITK) |
|-----------------|--------------|-------------|--------|-----|--------|------------------------------|--------------------------- |
|256, 256       | False            | True           | 0.00379 | ---- | 0.00081 | 4.66                         | ---- |
|256, 256 | True | False | 0.00428 | 0.00383 | 0.00102 | 4.21 | 3.76 |
|256, 256 | True | True | 0.00429 | ---- | 0.00107 | 4.02 | ---- |
|512, 512 | False | True | 0.01724 | ---- | 0.00091 | 18.97 | ---- |
|512, 512 | True | False | 0.01866 | 0.00788 | 0.00104 | 17.94 | 7.58 |
|512, 512 | True | True | 0.01876 | ---- | 0.00107 | 17.48 | ---- |
|2048, 2048 | False | True | 0.35053 | ---- | 0.00256 | 137.03 | ---- |
|2048, 2048 | True | False | 0.39468 | 0.08220 | 0.00288 | 137.02 | 28.54 |
|2048, 2048 | True | True | 0.39356 | ---- | 0.00296 | 133.04 | ---- |
|4096, 4096 | False | True | 1.57430 | ---- | 0.00736 | 213.85 | ---- |
|4096, 4096 | True | False | 1.79513 | 0.54968 | 0.00858 | 209.34 | 64.10 |
|4096, 4096 | True | True | 1.81570 | ---- | 0.00889 | 204.25 | ---- |
|32, 32, 32 | False | True | 0.00319 | ---- | 0.00069 | 4.62 | ---- |
|32, 32, 32 | True | False | 0.00356 | 0.00481 | 0.00096 | 3.71 | 5.01 |
[...]

PR #407, opened 2022-09-06
Support `sampling` kwarg for `distance_transform_edt` (take pixel/voxel sizes into account)

grlee77:
Please review/merge #406 before this one as those changes are also present here.

This PR adds support for anisotropic pixel/voxel shape to `distance_transform_edt` via the `sampling` kwarg. The changes involve just making variants of the kernels involving distance computations that do floating-point instead of integer-valued computations so that scaling by the sample spacings can be taken into account.


gigony (approved):
Thanks @grlee77 ! Looks good to me.

grlee77:
@gigony, I added two commits here after your prior review. These were to support the remaining two keyword arguments allowing the user to specify a pre-allocated output array for `distances` and/or `indices`. With that change we now fully match the SciPy API

grlee77:
@gpucibot merge

PR #409, opened 2022-09-07
Correctly use dtype when computing shared memory requirements of separable convolution

### LINKED ISSUE #408: [BUG] shared memory error during separable kernel compilation edge cases
**Describe the bug**

For separable convolution, the newer kernels are used by default, but when shared memory requirements are too large, we fall back to the CuPy-based convolution. Unfortunately, the shared memory estimation is currently incorrect in cases where the `itemsize` of the dtype is not equal to 4 bytes per element. This is not a big deal when the `itemsize` is smaller, but when the dtype has a larger `itemsize`, such as for `cupy.float64` (itemsize=8) the shared memory estimation can fail for some range of larger kernel sizes that are between 1 and 2 times the true shared memory limit.

Here is an example with a large kernel size where shared memory limits are satisfied for float32, but not for float64 on my system. The 
specific range of sigma that causes the error will be GPU-dependent due to different shared memory limits across models.

```py
import cupy as cp
from cucim.skimage.filters import gaussian

gaussian(cp.ones((1024, 1024), dtype=cp.float64), sigma=30)
```
will result in an error such as:

```
ptxas error   : Entry function 'correlate_s241_2d_ax0_nearest_Tdouble_Wdouble_Ddouble_patch16_halo16' uses too much shared data (0x18000 bytes, 0xc000 max)
```

**Expected behavior**

occurs for cuCIM 22.08.00 or 22.08.01 only

**Additional context**

grlee77:
closes #408

An error in the keyword arguments passed to the shared memory calculation is fixed in the first commit here. In the second commit, a second fallback is added in case of any other anticipated compilation failure.

### DIFF
--- a/python/cucim/src/cucim/skimage/_vendored/_ndimage_filters.py
+++ b/python/cucim/src/cucim/skimage/_vendored/_ndimage_filters.py
@@ -190,7 +190,9 @@ def _correlate_or_convolve1d(input, weights, axis, output, mode, cval, origin,
                              convolution=False, algorithm=None):
     # Calls fast shared-memory convolution when possible, otherwise falls back
     # to the vendored elementwise _correlate_or_convolve
+    default_algorithm = False
     if algorithm is None:
+        default_algorithm = True
         if input.ndim == 2 and weights.size <= 256:
             algorithm = 'shared_memory'
         else:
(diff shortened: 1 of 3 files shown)


jakirkham:
rerun tests

jakirkham (approved):

jakirkham:
Thanks Greg! 🙏

Happy with merging. Only question would be if we can include a test for the error. Could happen in this PR or after

gigony (approved):
looks good to me. Thank you Greg!

grlee77:
> Only question would be if we can include a test for the error. Could happen in this PR or after

Test case has now been added (values aren't checked, it is just to verify that no compilation error occurs). I verified that without the fix in this MR, the dtype=float64 case fails for the parameterized sigmas 16, 21, 26 and 31 on my system.

jakirkham (approved):

gigony:
@gpucibot merge

Issue #419, opened 2022-10-18
[FEA] port new features and bug fixes from scikit-image 0.20


grlee77:
**Is your feature request related to a problem? Please describe.**
A scikit-image 0.20 release is planned in the near future. That release will add several new features and has a number of bug fixes. It also will add wheels for Python 3.11.

[Current draft of the release notes](https://github.com/scikit-image/scikit-image/pull/6556)

**Describe the solution you'd like**
We should port the relevant features here, ideally for inclusion in `22.12.00`. Fortunately, one of the larger ones (footprint decompositions for morphology) was already previously merged here. 

**Describe alternatives you've considered**
N/A

**Additional context**
N/A

grlee77:
All remaining features for 0.20 should be covered across #421, #422, #423, #424

grlee77:
closed by the 4 PRs mentioned above as well as #451, #455

There are a handful of deprecations mentioned in #451 that are delayed to a future release, but I will go ahead and close this as largely resolved.

PR #421, opened 2022-11-01
add isotropic binary morphology functions

grlee77:
related to #419

These are based on the distance transform and are a faster way of computing binary morphological operations for large diameter disk or ball footprints. Unlike the sequence footprint decomposition methods, the footprint is exactly circular (spherical).

One test case `test_isotropic_erosion_spacing` will require #407 to be merged first. 

#406 will also improve the performance of this implementation

### DIFF
--- a/benchmarks/skimage/cucim_morphology_bench.py
+++ b/benchmarks/skimage/cucim_morphology_bench.py
@@ -53,7 +53,7 @@ class BinaryMorphologyBench(ImageBench):
         )
 
     def set_args(self, dtype):
-        imaged = cp.random.standard_normal(self.shape).astype(dtype) > 0
+        imaged = (cp.random.standard_normal(self.shape) > 0).astype(dtype)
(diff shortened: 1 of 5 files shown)


gigony (approved):
Looks good to me! Thanks @grlee77 !

grlee77:
rerun tests

grlee77:
Not sure why, but use of the out kwarg causes a failure in only one of the CI test runs. I am trying with an explicit copy instead to see if that resolves it.

The test case with the issue appears to be:

```
FAILED src/cucim/skimage/morphology/tests/test_isotropic.py::test_out_argument - 
cupy_backends.cuda.api.driver.CUDADriverError: CUDA_ERROR_ILLEGAL_ADDRESS: 
an illegal memory access was encountered
```
and then every subsequent test also fails with this same error.

grlee77:
@gpucibot merge

PR #432, opened 2022-11-08
Fix rotate behavior for ndim > 2

### LINKED ISSUE #431: [BUG] image.ndim > 2 broken for cucim.skimage.transform.rotate when resize=True
**Describe the bug**

In scikit-image, rotate also works for color images (assuming colors are along axis=[-1]) by applying the 2D rotation in a loop over the last axis. 

cuCIM currently works for color images only if `resize=False`, otherwise extra dimensions are not accounted for properly and a 3D input will become 2D in the output. This can be fixed by updating the `output_shape` and affine matrix used for the `resize=True` code path.

For the following example, the assertions would pass for scikit-image, but currently fail in cuCIM.
```py
import cupy as cp
from skimage import data
from cucim.transform import rotate

image = cp.array(data.astronaut())  // example RGB image
image_rot = rotate(image, angle=15, resize=True)
assert image.ndim == image_rot.ndim
assert image.shape[-1] == image_rot.shape[-1]
```

grlee77:
closes #431

Fixes the `resize=True` code path to take additional dimensions into account. `rotate` will now rotate the first two axes, preserving any additional ones. 

New test cases were added to confirm that no dimensions are lost.

### DIFF
--- a/python/cucim/src/cucim/skimage/transform/_warps.py
+++ b/python/cucim/src/cucim/skimage/transform/_warps.py
@@ -423,6 +423,10 @@ def rotate(image, angle, resize=False, center=None, order=None,
     symmetric, the result would be [0, 1, 2, 2, 1, 0, 0], while for reflect it
     would be [0, 1, 2, 1, 0, 1, 2].
 
+    If ``image.ndim > 2``, the rotation occurs for the first two dimensions of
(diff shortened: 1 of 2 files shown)


grlee77:
rerun tests

gigony (approved):
Thanks Greg! Looks good to me!

PR #433, opened 2022-11-09
update incorrect argument and deprecated function for tifffile.TiffWriter

JoohyungLee0106:
1. Both ```tifffile.TiffWriter.write()```[[link](https://github.com/cgohlke/tifffile/blob/master/tifffile/tifffile.py#L1530)] and ```tifffile.TiffWriter.save()```[[link](https://github.com/cgohlke/tifffile/blob/master/tifffile/tifffile.py#L3284)] receive ```compression``` not ```compress```
->  Current version doesn't work and raises the following error msg:
```
tifffile TypeError: write() got an unexpected keyword argument 'compress'
```

3. ```tifffile.TiffWriter.write()``` instead of ```tifffile.TiffWriter.save()```, which has been deprecated [[link](https://github.com/cgohlke/tifffile/blob/master/tifffile/tifffile.py#L3287)]


grlee77:
Thanks for fixing the issue @JoohyungLee0106. 

I looked into the history of the API change and don't think we may not need to support a fallback code path for older `tifffile` versions. It looks like the suggestion here should work as long as we use `tifffile>=v2020.9.30`. Given that this version is from > 2 years ago, that seems okay to me. 

We already pinned to a newer version in our test requirements here: https://github.com/rapidsai/cucim/blob/branch-22.12/python/cucim/requirements-test.txt

However, I do see a few files pinned to a specific, older version:
https://github.com/rapidsai/cucim/blob/branch-22.12/docker/requirements-claratrain.txt
https://github.com/rapidsai/cucim/blob/branch-22.12/docker/requirements-jupyter.txt
https://github.com/rapidsai/cucim/blob/branch-22.12/docker/requirements-jupyter-dev.txt
@gbae, do you know of a reason those specific versions were pinned or can we just update them?

GPUtester:

Admins can comment `ok to test` to allow this one PR to run or `add to allowlist` to allow all future PRs from the same author to run.

GPUtester:

Admins can comment `ok to test` to allow this one PR to run or `add to allowlist` to allow all future PRs from the same author to run.

grlee77:
ok to test

gigony:
@grlee77 those requirement files are not used in anywhere (it was used when cuCIM was released as a separate package but those need to be removed in the future).

Updated this PR to add a test case and update the hard-coded version.

grlee77:
@gigony, looks like we have to add opencv-python to `requirements-test.txt` as `tiff.py` uses `cv2.resize`

grlee77 (approved):
This looks good, assuming CI passes this time around.

run
```diff
@@ -806,10 +806,9 @@ test() {
 
 install_python_test_deps_() {
     if [ -n "${CONDA_PREFIX}" ]; then
-        # https://github.com/rapidsai/cucim/pull/349#issuecomment-1203335731
-        # Do not update or change already-installed dependencies.
-        run_command conda install -c conda-forge -y --freeze-installed \
-            --file ${TOP}/python/cucim/requirements-test.txt
+        # Install test dependencies from pip because conda does not have all the dependencies (e.g., opencv-python).
```
grlee77, line 809: conda does have it, but the package is not called `opencv-python`. It was originally just `opencv` on `conda-forge` and included the python bindings, but this has apparently changed recently to `libopencv` with a `py-opencv` meta-package as described here:
https://github.com/conda-forge/opencv-feedstock/blob/4129446ade9cb1928fd40897960ed32a21c7f771/recipe/meta.yaml#L4-L21

That said, I am fine leaving this as a pip install here

grlee77: I see you later changed to `opencv-python-headless` for which I don't think there is a conda equivalent

grlee77 (approved):
Thanks Gigon, everything looks good now.

grlee77:
@gpucibot merge

grlee77:
Thanks again, @JoohyungLee0106. This fix will appear in the 2022.12.00 release (sometime around Dec 7)

ajschmidt8 (approved):

PR #451, opened 2022-11-18
Implement additional deprecations carried out for scikit-image 0.20

grlee77:
related to #419

This PR removes most of the deprecated parameters that are being removed in scikit-image 0.20.

There are a few exceptions (e.g. `selem`) where they will be removed in scikit-image 0.20, but we had previously advertised a specific removal date like `2022.03.02`. For those, I will delay the removal until the previously advertised release.

It will be nice to be rid of all the deprecated `multichannel` code! The first, large commit here is just the multichannel deprecation. The rest of the commits are much smaller.


grlee77:
These deprecations have been approved, but not yet merged upstream. See comment here: https://github.com/scikit-image/scikit-image/pull/6583#issuecomment-1331068074

I think it would be nice to not have all of these deprecated parameters around, but am also okay if we want to bump this until 2022.02.00.

gigony (approved):
Thanks @grlee77 ! It looks good to me!

grlee77:
@gpucibot merge

Issue #375, opened 2022-08-02
Refactoring kernel code into .cu/.c files


jakirkham:
In a few cases, we have needed to include kernel code in strings in Python. It would be good to start refactoring this out into separate .cu/.c files that we load instead. This has the benefit that .cu/.c files are more familiar to developers and various tooling (linters/compilers/etc.). Also it can potentially be used in C/C++ code directly. Making this change could help with general maintainability & usability of the codebase going forward.

_Originally posted by @jakirkham in https://github.com/rapidsai/cucim/pull/317#discussion_r934985904_

jakirkham:
As one example of this kind of change, please see commit ( https://github.com/rapidsai/cucim/pull/317/commits/2608b2e9771ea4143f10ca8c29660cc0a19150ca ). There are some other CUDA or C/C++ codes in Python `str`s. It would likely be simple to refactor these out into `.cu` or `.c`/`.cpp`.

A separate, but related question would be how we would like to use these on the C++ side.

PR #464, opened 2022-12-06
[WIP/RFC]: Refactor CUDA code out of cuCIM

jakirkham:
Addresses https://github.com/rapidsai/cucim/issues/375

This is a very rough attempt at starting to refactor CUDA code out of Python `str`s and into CUDA files. Starting with just one function to see how this goes and work through any issues that come up. If it is reasonable and proves viable, this could be expanded upon.


python/cucim/src/cucim/skimage/measure/_moments.py
```diff
@@ -519,20 +520,19 @@ def centroid(image, *, spacing=None):
...
-    F mu0, mxx, mxy, myy;
-    mu0 = mu[0];
-    mxx = mu[6];
-    myy = mu[2];
-    mxy = mu[4];
-
-    result[0] = myy / mu0;
-    result[1] = result[2] = -mxy / mu0;
-    result[3] = mxx / mu0;
+    inertia_tensor_2x2(&mu[0], &result[0]);
```
jakirkham, line 530: FWICT CuPy actually passes [`CArray`]( https://github.com/cupy/cupy/blob/aa12407137ae15ad014e14842c5e2a18338716cd/cupy/_core/include/cupy/carray.cuh#L379 ) objects into the kernel. These can be [indexed]( https://github.com/cupy/cupy/blob/aa12407137ae15ad014e14842c5e2a18338716cd/cupy/_core/include/cupy/carray.cuh#L588 ) though, which in [the C-contiguous case]( https://github.com/cupy/cupy/blob/aa12407137ae15ad014e14842c5e2a18338716cd/cupy/_core/include/cupy/carray.cuh#LL589 ) indexes [directly into the `data_` pointer]( https://github.com/cupy/cupy/blob/aa12407137ae15ad014e14842c5e2a18338716cd/cupy/_core/include/cupy/carray.cuh#L592 ). We leverage this here to back out the pointer to underlying array data, which the kernel then takes for normal operation.

python/cucim/src/cucim/skimage/measure/_moments.py
```diff
@@ -519,20 +520,19 @@ def centroid(image, *, spacing=None):
 
 
 def _get_inertia_tensor_2x2_kernel():
```
grlee77, line 522:
```suggestion
@cp.memoize(for_each_device=True)
def _get_inertia_tensor_2x2_kernel():
```

python/cucim/src/cucim/skimage/measure/_moments.py
grlee77: as long as we make sure to check that the inputs are C-contiguous, this seems okay

grlee77:
This seems reasonable for quite a few of the kernels we have.  Would you want to use this approach for even more trivial cases like the following two?:
https://github.com/rapidsai/cucim/blob/6c50a4edc94494bbb57dbd02749870d45533ee68/python/cucim/src/cucim/skimage/feature/corner.py#L357-L377

A slightly more involved example would be the following where there are two parameters that affect the generated code. The first part of the kernel is always the same, but the final portion can sort the eigenvalues in ascending or descending value, optionally via absolute values.
https://github.com/rapidsai/cucim/blob/6c50a4edc94494bbb57dbd02749870d45533ee68/python/cucim/src/cucim/skimage/feature/corner.py#L534-L615

The most complicated are in the `skimage/_vendored` folder such as this one used by transforms like `rotate`, `warp`, `resize`, etc.:
https://github.com/rapidsai/cucim/blob/6c50a4edc94494bbb57dbd02749870d45533ee68/python/cucim/src/cucim/skimage/_vendored/_ndimage_interp_kernels.py#L226-L510
There, the code is generated dynamically based on many attributes such as border mode, interpolation order and the number of dimensions which makes it very flexible, but quite hard to follow. I don't see a way to easily adapt those cases to this approach.

python/cucim/src/cucim/skimage/measure/_moments.py
jakirkham: There is a fallback for [the non-contiguous case]( https://github.com/cupy/cupy/blob/aa12407137ae15ad014e14842c5e2a18338716cd/cupy/_core/include/cupy/carray.cuh#L593 )

That said, the C code effectively assumes the data is contiguous somehow. We could require that before calling the kernel. Alternatively we could rewrite the kernel so it is less reliant on this requirement. Though not sure how often more complex cases are needed given this seems to be an internal function. Please feel free to correct me if I'm missing something

python/cucim/src/cucim/skimage/measure/_moments.py
```diff
@@ -519,20 +520,19 @@ def centroid(image, *, spacing=None):
...
+    kernel_directory = os.path.join(
+        os.path.normpath(os.path.dirname(__file__)), 'cuda'
+    )
+    with open(os.path.join(kernel_directory, "moments.h"), 'rt') as f:
+        preamble = f.read()
```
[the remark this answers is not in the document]
jakirkham: Agreed. Was hoping that we could write

```C++
#include "moments.h"
```

Not sure if there is a flag we need to provide to tell the kernel where to search for headers

jakirkham:
> This seems reasonable for quite a few of the kernels we have.

Indeed this was my hope and thinking. Glad to hear you agree :)

> Would you want to use this approach for even more trivial cases like the following two?

I'm not sure. That probably also depends whether this kernel is useful in CUDA itself. Think, as you hint, it is probably not worth it for the trivial cases, but could be wrong.

For more context picked this particular kernel since it was an easy one to try this idea out on, but still with enough meat on it to learn a few things along the way. It turned out to not be that difficult. So am encouraged we may be able to explore this strategy more.

> A slightly more involved example would be the following where there are two parameters that affect the generated code.

Thanks for the suggestion! Can take a closer look.

> The first part of the kernel is always the same, but the final portion can sort the eigenvalues in ascending or descending value, optionally via absolute values.

Interesting. Maybe the final portion could be a second templated CUDA function called by the first?

> There, the code is generated dynamically based on many attributes such as border mode, interpolation order and the number of dimensions which makes it very flexible, but quite hard to follow. I don't see a way to easily adapt those cases to this approach.

Indeed. It probably can be done. Though may need to resort to `#define` macros. That being said, as it is not low hanging fruit, it is probably best to save until the end or handle separately. Unless there is a compelling need to move it over to CUDA sooner.

jakirkham:
rerun tests

(one of the CI jobs hung)

python/cucim/src/cucim/skimage/measure/_moments.py
grlee77: I would be okay with removing it, given that it should now be <1 microsecond benefit to having it. No strong opinion either way in this case.

grlee77 (approved):
Thanks, this looks good. I agree the `#include` approach is preferable to parsing the file in Python.

PR #471, opened 2022-12-22
Add GitHub Actions Workflows.

bdice:
This PR adds GitHub Actions workflows to `cucim`.

### Task list

Coverage required for this PR:
- [x] C++ build
- [x] ~C++ tests~ (There are no C++ tests.)
- [x] Python build
- [x] Python tests
  - Some tests have been marked as `xfail` on ARM following discussion with @jakirkham and @grlee77. Those will be addressed in follow-up work.
- [x] Style checks


ci/test_notebooks.sh
```diff
@@ -0,0 +1,62 @@
+#!/bin/bash
```
ajschmidt8, line 1: @bdice, I see this `test_notebooks.sh` file, but I don't see a corresponding job in `pr.yaml`.

Also, it doesn't seem that `cucim` notebooks are currently tested in Jenkins PRs.

Therefore, should we punt notebooks testing for `cucim` for a separate PR to prevent scope creep?

bdice: I was going to make a quick attempt at running notebooks and defer to a later PR if it does not work.

In general, my strategy for all the tasks we want to do in this release is to try things aggressively, and reduce scope if an “optional” task fails after some small, fixed amount of effort. Some bets pay off, others do not, but I try to keep the downside loss (budget of effort for non-essential goals) small.

ajschmidt8:
Here's a summary of the latest `pytest` failures:

- on `amd64`
  - it seems like all of the `amd64` tests only have a single failure ([1](https://github.com/rapidsai/cucim/actions/runs/3818400587/attempts/1#summary-10386516675),[2](https://github.com/rapidsai/cucim/actions/runs/3818400587/attempts/1#summary-10386516722),[3](https://github.com/rapidsai/cucim/actions/runs/3818400587/attempts/1#summary-10386516751)) that says `ImportError: libGL.so.1: cannot open shared object file: No such file or directory`. I can't seem to figure out where this `.so` is available in our Jenkins CI images, so I'm not sure how to fix it for GHAs.
- on `arm64`
  - the `arm64` test failures ([link](https://github.com/rapidsai/cucim/actions/runs/3818400587/attempts/1#summary-10386516778)) have the same `.so` error as `amd64`, plus a few others. Just a reminder, we don't test `arm64` in PRs on Jenkins, so these failures could be legit

bdice:
I think the `libGL.so.1` errors are from attempting to run code that requires a display (or virtual GL output device?) on a headless machine without GL libraries.

On my WSL Ubuntu 22.04 system, I have `/usr/lib/x86_64-linux-gnu/libGL.so.1` which is provided by the apt package `libgl1`. I am not sure if that's sufficient here. need to do something like export an environment variable to tell the package to run in "headless" mode. I see there are some channels like `fastai` offering a package named `opencv-python-headless`, and a PyPI package: https://pypi.org/project/opencv-python-headless/ but no `conda-forge` package labeled "headless."

Some related StackOverflow answers are here: https://stackoverflow.com/questions/55313610/importerror-libgl-so-1-cannot-open-shared-object-file-no-such-file-or-directo

bdice:
I think we have a few potential courses of action:
- Install `libgl1` in Ubuntu images and `mesa-libGL` in CentOS/Rocky images. I am convinced that this is necessary to use the conda-forge packages because of [this comment](https://github.com/conda-forge/opencv-feedstock/issues/321#issuecomment-1165865099)
- Use the [`fastai::opencv-python-headless`](https://anaconda.org/fastai/opencv-python-headless) package and skip OpenCV tests on aarch64 (package not available for aarch64)
- Try the PyPI `opencv-python-headless` by adding it as a `pip` dependency in the conda environment -- not sure if this would work but it might be worth trying
- Skip these tests entirely (probably not a sufficient solution)

bdice:
@ajschmidt8 I implemented the first proposed solution here: https://github.com/rapidsai/ci-imgs/pull/31

If you're satisfied with that, we can update the images and retry this PR.

python/cucim/src/cucim/skimage/exposure/tests/test_exposure.py
```diff
@@ -14,6 +15,10 @@
 from cucim.skimage.exposure.exposure import intensity_range
 from cucim.skimage.util.dtype import dtype_range
 
+# TODO: Some tests fail unexpectedly on ARM.
+ON_AARCH64 = platform.machine() == "aarch64"
+ON_AARCH64_REASON = "TODO: Test fails unexpectedly on ARM."
```
jakirkham, line 20: We decided to skip these after discussing with @grlee77 offline. Noting this here for clarity. Also for reference in a follow up issue

ajschmidt8 (approved):

python/cucim/.pre-commit-config.yaml
```diff
@@ -1,20 +0,0 @@
-# To install the git pre-commit hook run:
```
jakirkham, line 1: Was this meant to be dropped?

conda/environments/all_cuda-115_arch-x86_64.yaml
```diff
@@ -0,0 +1,48 @@
...
+- pytest>=6.2.4
+- python>=3.8,<3.10
+- recommonmark
+- scikit-image >=0.19.0,<0.20.0a0
+- scipy
+- sphinx
+- sysroot_linux-64==2.17
+- tifffile>=2022.7.28
+- xz
+- zlib
```
jakirkham, line 44: It looks like `yasm` is missing from here (at least for `x86_64`). Should we be handling that some other way?

jakirkham (commented):
Thanks all! 🙏

Sorry spotted a couple things below that I had questions on. Just want to make sure I'm not missing something

python/cucim/.pre-commit-config.yaml
bdice: This shouldn't exist as far as I understand. The only config that should matter is in the repo root.

dependencies.yaml
```diff
@@ -0,0 +1,153 @@
...
+          - numpy
+          - scikit-image >=0.19.0,<0.20.0a0
+          - scipy
+          - xz
+          - zlib
+          - zstd
+          # Not sure where these go, if anywhere:
+          # - openslide
+          # - xorg-libxcb
+          # - yasm
```
bdice, line 132: @jakirkham Related to your question about yasm above -- what section do these go in? Build deps? Runtime deps? Testing deps?

conda/environments/all_cuda-115_arch-x86_64.yaml
bdice: Replied here: https://github.com/rapidsai/cucim/pull/471/files#r1063810045

I didn't know what section of the dependencies to put it in.

dependencies.yaml
jakirkham: YASM is an assembler. It is used for building

Am not aware of any runtime assembly occurring here. Though it is a fair question given CuPy does some runtime compilation, but it doesn't use YASM AFAIK

bdice: Aha. I didn't see `yasm` in the libcucim conda recipe until now. I'll add it here as a requirement on x86_64, like in the conda recipe (though it's weird that only one platform would need it...).

What about `openslide` and `xorg-libxcb`? Are those runtime requirements? Test requirements?

bdice: yasm updated in b91daf5e43a571567df7601e6446e06b37be800d.

jakirkham: Sounds good. Thanks Bradley! 🙏

YASM is an `x86` & `x86_64` assembler. It wouldn't really make sense to use on ARM since it would generate instructions that don't run there (or do and cause havoc 😟)

I don't recall exactly where YASM is used by `libcucim`, but would not be surprised if there is a fallback path that works on any architecture (like some C code). YASM really comes in when one writes their own assembly for a function or set of functions. This can come in handy for some video encoding/decoding libraries, which are very performance sensitive.

python/cucim/.pre-commit-config.yaml
jakirkham: Interestingly it seems we are including this one in [`MANIFEST.in`]( https://github.com/rapidsai/cucim/blob/dbbeaf92f7f5175b0eb47bb88aa2e3b74c88a5cb/python/cucim/MANIFEST.in#L20 ) though

In any event, agree we should only have this in one place. If we stick to dropping this one, the `MANIFEST.in` needs an update

bdice: Good catch! Removed from `MANIFEST.in` in 93c2c1ebbdb0e12768578eb851d2ef527227e774.

jakirkham (approved):

python/cucim/.pre-commit-config.yaml
jakirkham: Thanks Bradley! 🙏

Given the file is still here (just in another location), it is not really being dropped. If it turns out we need to move it for whatever reason, we can do that in a follow up.

PR #472, opened 2023-01-04
Fix random seed used in test_3d_similarity_estimation

grlee77:
This MR fixes the random seed used in `test_3d_similarity_estimation` to avoid a stochastic test failure observed both locally and on CI. While fixing the seed, the test case is also updated to use the newer style `RandomGenerator` API.

### DIFF
--- a/python/cucim/src/cucim/skimage/transform/tests/test_geometric.py
+++ b/python/cucim/src/cucim/skimage/transform/tests/test_geometric.py
@@ -147,13 +147,17 @@ def test_similarity_estimation():
 
 
 def test_3d_similarity_estimation():
-    src_points = cp.random.rand(1000, 3)
+    # Note: some choices of seed will produce poorly conditioned values,
+    #       leading to NaN in the estimate. We fix the seed here to avoid this.
+    seed = 10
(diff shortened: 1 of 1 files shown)


python/cucim/src/cucim/skimage/transform/tests/test_geometric.py
```diff
@@ -147,13 +147,17 @@ def test_similarity_estimation():
 
 
 def test_3d_similarity_estimation():
-    src_points = cp.random.rand(1000, 3)
+    # Note: some choices of seed will produce poorly conditioned values,
+    #       leading to NaN in the estimate. We fix the seed here to avoid this.
+    seed = 10
+    rng = cp.random.default_rng(seed)
+    src_points = rng.random(size=(1000, 3))
```
jakirkham, line 154: Maybe we should consider parameterizing these. Though that can be a follow up item separate from this PR

jakirkham (approved):

jakirkham:
Thanks Greg! 🙏

jakirkham:
@gpucibot merge

PR #474, opened 2023-01-10
Normalize whitespace.

bdice:
This PR re-enables pre-commit checks for normalizing whitespace (trimming trailing spaces) that I disabled in #471.

These pre-commit hooks existed prior to the GitHub Actions migration but they were not being enforced by CI style checks. I disabled them in #471 because enforcing them would have made the PR diff large.

The primary change is in `.pre-commit-config.yaml`, and everything else was automatically fixed by `pre-commit`.
```
# TODO: re-enable this after GitHub Actions migration, it currently makes a
# large number of changes that shouldn't go in the GitHub Actions PR.
#- repo: https://github.com/pre-commit/pre-commit-hooks
#  rev: v4.4.0
#  hooks:
#    - id: trailing-whitespace
#    - id: end-of-file-fixer
#    - id: debug-statements
```


ajschmidt8 (approved):

grlee77 (approved):
The changes look good. Looks like the CI failure is probably an unrelated network issue (failed download)

grlee77:
Thanks, @bdice

raydouglass (approved):

jakirkham:
Toggling for CI

jakirkham:
@gpucibot merge

cpp/cmake/deps/boost-header-only.patch
```diff
@@ -14,4 +14,3 @@ index 7da31f7..5816de7 100644
 +   //BOOST_ASSERT(0 == ((std::size_t)ret % ::boost::container::dtl::alignment_of<T>::value));
     return static_cast<T*>(ret);
  }
-
```
jakirkham, line 17: Think we want to keep this line. Otherwise the patch doesn't apply

jakirkham: Reverted in commit ( https://github.com/rapidsai/cucim/pull/474/commits/43606cf60afd5ce2a1693061d684e1da8b5f3687 )

jakirkham:
Looks like [the style check doesn't like that change]( https://github.com/rapidsai/cucim/actions/runs/4623027134/jobs/8176424920?pr=474#step:6:129 ) 😂 

```diff
If you are seeing this message in CI, reproduce locally with: `pre-commit run --all-files`.
To run `pre-commit` as part of git workflow, use `pre-commit install`.
All changes made by hooks:
diff --git a/cpp/cmake/deps/boost-header-only.patch b/cpp/cmake/deps/boost-header-only.patch
index e765790..96c0078 100644
--- a/cpp/cmake/deps/boost-header-only.patch
+++ b/cpp/cmake/deps/boost-header-only.patch
@@ -14,4 +14,3 @@ index 7da31f7..5816de7 100644
 +   //BOOST_ASSERT(0 == ((std::size_t)ret % ::boost::container::dtl::alignment_of<T>::value));
     return static_cast<T*>(ret);
  }
- 
```

Maybe we can skip this file during style checking?

.pre-commit-config.yaml
```diff
@@ -2,16 +2,14 @@
 #   pre-commit install
 # To update the pre-commit hooks run:
 #   pre-commit install-hooks
-exclude: '^(\.tox|ci/templates|\.bumpversion\.cfg)(/|$)'
+exclude: '^(\.tox|ci/templates|\.bumpversion\.cfg|\.patch)(/|$)'
```
jakirkham, line 5: Excluded `.patch` here. Not sure if we need to exclude this anywhere else

bdice: This excludes a single file at the root named `.patch`, I believe. I think you need something like:
```suggestion
exclude: '^(\.tox|ci/templates|\.bumpversion\.cfg|.*\.patch)(/|$)'
```

jakirkham: Thanks Bradley! 🙏

Committed this change

jakirkham (approved):

jakirkham:
Thanks all! 🙏

PR #476, opened 2023-01-11
Build CUDA `11.8` and Python `3.10` Packages

ajschmidt8:
This PR updates `cucim` to build against branch [cuda-118](https://github.com/rapidsai/shared-action-workflows/compare/cuda-118) of the `shared-action-workflow` repository.

That branch contains updates for CUDA `11.8` and Python `3.10` packages.

It also includes some minor file renames.

PR #480, opened 2023-01-16
Update `numpy` version specifier

ajschmidt8:
This PR updates the `numpy` version specifier for `cucim`.

It was previously updated in [this commit](https://github.com/rapidsai/cucim/pull/476/commits/043810aa6115b6b7183535037f6d9366769e6ece) to enable publishing Python `3.10` packages.

However, since there was no maximum version in that specifier, `conda` was picking up the latest version of `numpy` (`1.24.1`) which is incompatible with `cudf`'s version of `numba` ([`numba>=0.56.2`](https://github.com/rapidsai/cudf/blob/36b07f13480d8f7ba18274cb3a63f06855d848ff/conda/recipes/cudf/meta.yaml#L52)).

For instance, if you try to run this: `mamba create -n test 'numba>=0.56.2' 'numpy>=1.24.1'`, it will fail to solve.

Running with the new version specifier in this PR does solve: `mamba create -n test 'numba>=0.56.2' 'numpy>=1.21.3,<1.24'`.

Therefore, this maximum version limit seems necessary for `cucim` to be compatible with the rest of RAPIDS.

### DIFF
--- a/conda/recipes/cucim/meta.yaml
+++ b/conda/recipes/cucim/meta.yaml
@@ -43,7 +43,7 @@ requirements:
     - cudatoolkit ={{ cuda_version }}
     - cupy >=10,<12.0.0a0
     - libcucim ={{ version }}
-    - numpy >=1.21.3
+    - numpy >=1.21.3,<1.24
     - python
     - scikit-image >=0.19.0,<0.20.0a0
     - scipy
--- a/conda/environments/all_cuda-118_arch-x86_64.yaml
+++ b/conda/environments/all_cuda-118_arch-x86_64.yaml
@@ -21,7 +21,7 @@ dependencies:
 - libwebp-base
 - nbsphinx
 - ninja
-- numpy >=1.21.3
+- numpy >=1.21.3, <1.24
 - numpydoc
 - nvcc_linux-64=11.8
 - openslide-python>=1.1.2
--- a/dependencies.yaml
+++ b/dependencies.yaml
@@ -129,7 +129,7 @@ dependencies:
           - jbig
           - jpeg
           - libwebp-base
-          - numpy >=1.21.3
+          - numpy >=1.21.3, <1.24
           - scikit-image >=0.19.0,<0.20.0a0
           - scipy
           - xz


ajschmidt8:
This PR should unblock https://github.com/rapidsai/dask-cuda/pull/1080.

bdice:
I'm not sure if this is the right fix here. Only packages using `numba` should have a upper bound on `numpy`, and it doesn't look like this package depends on `numba`. Also, this upper bound pinning only applies until `numba 0.57.0` is released (which adds support for newer `numpy` versions).

Instead, shouldn't we be pinning only cudf (and not cucim)? https://github.com/rapidsai/cudf/blob/36b07f13480d8f7ba18274cb3a63f06855d848ff/conda/recipes/cudf/meta.yaml#L65

conda/recipes/cucim/meta.yaml
```diff
@@ -43,7 +43,7 @@ requirements:
     - cudatoolkit ={{ cuda_version }}
     - cupy >=10,<12.0.0a0
     - libcucim ={{ version }}
-    - numpy >=1.21.3
+    - numpy >=1.21.3,<1.24
```
bdice, line 46: Discussed with @ajschmidt8 offline. This was solving to `1.24.1` in the `host`, which meant that the runtime `pin_compatible` was something like `>=1.24` -- which is incompatible with cudf.

The solution is that `host` requirements should build against the oldest supported NumPy. Then the `pin_compatible` won't conflict with cudf's requirements.
```suggestion
    - numpy =1.21.3
```

conda/recipes/cucim/meta.yaml
```diff
@@ -43,7 +43,7 @@ requirements:
     - cudatoolkit ={{ cuda_version }}
     - cupy >=10,<12.0.0a0
     - libcucim ={{ version }}
-    - numpy >=1.21.3
+    - numpy =1.21.3
```
jakirkham, line 46: Wonder if we should relax the patch version (as that shouldn't matter as much)

bdice: The changelog for numpy 1.21.3 says it is the first with Python 3.10 support. I am fine with the relaxed pinning if that is not an issue.

raydouglass (approved):

conda/recipes/cucim/meta.yaml
ajschmidt8: after committing this change, it looks like the `3.10` packages built with `numpy` version `1.21.6`, which should be fine. 

merging!

jakirkham:
Thanks all! 🙏

PR #485, opened 2023-01-24
Improve rank filtering performance by removing use of footprint kernel when possible

grlee77:
In many cases such as rank filtering, if all elements of the kernel array are equal to one, then there is no reason to pay the overhead of allocating the footprint or checking its values during the filtering operation.

This MR is a small update to vendored CuPy ndimage code to avoid overhead of footprint creation and use when possible. 

Will update with benchmark results of acceleration vs. the kernel size. It is approximately a 2x improvement for small kernels (e.g. 3x3, 5x5, 3x3x3) approaching no acceleration for larger kernel sizes.


grlee77:
Here is a benchmark on a 2048x2048 2D image. (not sure why there is a spike specifically at size 13x13)

median_accel

grlee77:
This did require one API addition not present in scikit-image. Namely, one can provide an integer size instead of a footprint if a fully populated square footprint is desired. (as available in the API of `cupyx.scipy.median_filter`)

PR #486, opened 2023-01-25
Add disambiguation option to phase_cross_correlation (skimage 0.20 feature)

grlee77:
This MR ports a couple of recent features merged upstream for scikit-image 0.20

- a docstring update to `lab2rgb` and `rgb2lab`
- a new "disambiguation" flag to `phase_cross_correlation` (default of False matches the old behavior)

There is also one change that is technically breaking in that the shift is now always returned as a tuple of floats rather than as a `cupy.ndarray` of size `refernce_image.ndim`. I think this is preferable and working with small coordinate tuples rather than using array operations allows avoiding kernel launch overheads. Upstream scikit-image is a bit inconsistent in that it returns a tuple when `disambiguate` is True, but a numpy.ndarray otherwise. It should probably be changed to return a tuple in all cases.


gigony (approved):
Thanks @grlee77 !Looks good to me!

PR #491, opened 2023-01-30
[REVIEW] Closes #490 -- fixes bug in hue jitter

### LINKED ISSUE #490: [BUG] Hue color jitter on an all gray or all white image returns a black image
## Describe the bug

Hue jitter, via `cucim.core.operations.color.color_jitter`, on images with a lot (or all) gray seems to overflow and give either black pixels, or an all-black image.

I noticed this when using augmenting images of planes in, e.g., CIFAR: `color_jitter` with a non-zero hue parameter would render some sky pixels black. I found this issue only comes up when the hue parameter is non-zero -- other jitter parameters didn't seem to affect the behavior. Further experimenting showed that inputting an image of all white or all one shade of gray resulted in `color_jitter` outputting an image of all zeros. 

More specifically:
```python
import cupy as cp
import cucim.core.operations.color as ccl

x_cupy = cp.ones((3, 64, 64)).astype(cp.uint8)*255
x_jitter = ccl.color_jitter(x_cupy,0,0,0,0.1)

#Outputs all zeros, i.e.
(x_jitter == 0).all() # is True
```
The same behavior occurs for any gray image, e.g. if I make `x_cupy` a tensor of all intensity 128.

Perhaps this is some rounding/overflow issue since `color_jitter` seems to convert all inputs to `uint8`? 

## Expected behavior 

I would expect the output to not be all black but, if the image is all gray then hue jitter should have no effect -- it should output the same gray image. The torch implementation, for instance, behaves how I would expect:
```python
import torch
import numpy as np 
import torchvision
x_np = np.ones((3, 64, 64)).astype(np.uint8)*255
x_torch = torch.tensor(x_np)
x_torch_jitter = torchvision.transforms.ColorJitter(0, 0, 0, 0.1)(x_torch)

#Jittered output is same as input image
(x_torch_jitter == 255).all() # is True
```

## Environment

Running Ubuntu 20.04. cudatoolkit 11.6.2.

Installed cuCIM with pip. 
python==3.8.13
cupy==11.2.0
cucim==22.10.00

(torch version for the above example == 1.13.0)

## Other notes

* I came across this issue when using MONAI's `RandCuCIM` transform for data augmentation, as it calls cucim's `color_jitter`.
* I've tried a number of `hue` parameters -- anything non-zero seems to create the issue. Though the docstring seems to suggest anything between [-0.5, 0.5] is accepted, including negative values, negative values in fact return `ValueError: If hue is a single number, it must be non negative.` But I suppose that's a separate issue.

benlansdell:
This PR fixes a small bug in hue color jitter, as described in Issue #490. In images of all gray, some or all pixels are set to black when they shouldn't be. Looking at the CUDA kernel, it seems like the issue lies on line 208 of [cuda_kernel_source.py](https://github.com/rapidsai/cucim/blob/6c50a4edc94494bbb57dbd02749870d45533ee68/python/cucim/src/cucim/core/operations/color/kernel/cuda_kernel_source.py).

Currently in the kernel if `s == 0` then it sets the whole output to zero. But HSV is only black if `v == 0`, not s. So line 208 is here changed to:
```
if (v == 0) {
```

First PR here, so let me know what else is needed! Thanks.

### DIFF
--- a/python/cucim/src/cucim/core/operations/color/kernel/cuda_kernel_source.py
+++ b/python/cucim/src/cucim/core/operations/color/kernel/cuda_kernel_source.py
@@ -205,7 +205,7 @@ __global__ void huejitter_kernel(unsigned char *input_rgb, \
     float s = __uint2float_rn((unsigned int)s_char);
     float v = __uint2float_rn((unsigned int)v_char);
 
-    if (s == 0) {
+    if (v == 0) {
       // write zero and out
       output_rgb[idx] = 0;
       output_rgb[idx+pitch] = 0;


grlee77 (approved):
Thanks, @benlansdell! This looks good to me.

cc @shekhardw for visibility

jakirkham:
ok to test

jakirkham (approved):

ajschmidt8:
@jakirkham, @grlee77, there were some test failures for `arm` below. I don't know anything about `cucim` to determine if they're legit, but I restarted that job to see if they were transient regardless.

Just posting for awareness.


- https://github.com/rapidsai/cucim/actions/runs/4068397175/jobs/7007242415

jakirkham:
Thanks AJ! 🙏

Yeah we are discussing this offline

PR #494, opened 2023-02-01
apply bug fix to vendored ndimage code

grlee77:
This single line fix was made as part of a larger CUDA 12.0 support PR upstream in CuPy.

### DIFF
--- a/python/cucim/src/cucim/skimage/_vendored/_ndimage_util.py
+++ b/python/cucim/src/cucim/skimage/_vendored/_ndimage_util.py
@@ -168,7 +168,7 @@ def _generate_boundary_condition_ops(mode, ix, xsize, int_t="int",
             {ix} %= {xsize};
             '''
             ops_lower = ops_upper + f'''
-            if ({ix} < 0) {{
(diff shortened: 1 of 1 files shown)


jakirkham (approved):

python/cucim/src/cucim/skimage/_vendored/_ndimage_util.py
```diff
@@ -168,7 +168,7 @@ def _generate_boundary_condition_ops(mode, ix, xsize, int_t="int",
             {ix} %= {xsize};
             '''
             ops_lower = ops_upper + f'''
-            if ({ix} < 0) {{
```
grlee77, line 171: It doesn't seem like this fix should be necessary? I would think the modulus on line 168 makes it so that the xsize should not need to be added more than once.

jakirkham: Asked about this in comment ( https://github.com/cupy/cupy/pull/7235#discussion_r1093835536 )

PR #507, opened 2023-02-07
Fix `cupy.round` for newer version of cupy

drbeh:
`cupy.round` is used in stain normalization and it is removed from the newer version of cupy (although I didn't find the changelog related to that). This PR simply change `cp.round` to `cp.around`, which should be backward compatible.


drbeh:
> Retargeted to `branch-23.04`, which is in development. Though it looks like that made things a bit messy. Would it be possible to rebase @drbeh? Sorry for the trouble

Thanks, I saw from the checks that main is not the right place. Will make another PR based on  `branch-23.04` to have a clean merge.

jakirkham:
Ok sounds good. Thanks Behrooz! 🙏 Sorry again for the confusion

jakirkham:
Ah looks like @grlee77 may have already opened this ( https://github.com/rapidsai/cucim/pull/508 )?

PR #508, opened 2023-02-07
use cp.around instead of cp.round for CuPy 10.x compatiblity

grlee77:
`around` and `round` are equivalent, but `round` is only present in CuPy>=11. Switch to `around` here for compatibility with CuPy 10.x releases.

PR #509, opened 2023-02-07
Change `cp.round` `to cp.around`

drbeh:
cupy.round is used in stain normalization and it is removed from the newer version of cupy (although I didn't find the changelog related to that). This PR simply change cp.round to cp.around, which should be backward compatible.


drbeh:
closing in favor of https://github.com/rapidsai/cucim/pull/508

PR #516, opened 2023-02-21
Fix inefficiency in handling clipping of image range in `resize` and other transforms

grlee77:
There is substantial overhead when using the default `clip=True` in image transforms.  In this MR, there are two changes to improve this

- refactor `_clip_warp_output` to avoid recomputing min and max for cases like `resize` where the calling function has already computed them. In that case the (min, max) tuple can be directly passed to this utility instead of the image itself.
- change the default from `clip=True` to `clip=None` (for the default of None `clip` will be set to False for order 0 and 1 can True otherwise). The reason for the prior default of True was to deal with potentially ringing that can occur with higher order interpolation causing, e.g. uint8 values to fall outside of the original [0, 255] range.

### DIFF
--- a/python/cucim/src/cucim/skimage/transform/_warps.py
+++ b/python/cucim/src/cucim/skimage/transform/_warps.py
@@ -164,9 +164,7 @@ def resize(image, output_shape, order=None, mode='reflect', cval=0, clip=True,
         image = convert_to_float(image, preserve_range)
 
     # Save input value range for clip
-    img_bounds = (
-        cp.array([image.min(), image.max()]) if clip else None
-    )
+    img_bounds = (image.min(), image.max()) if clip else None
 
     # Translate modes used by np.pad to those used by scipy.ndimage
     ndi_mode = _to_ndimage_mode(mode)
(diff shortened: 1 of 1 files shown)


gigony (approved):

python/cucim/src/cucim/skimage/transform/_warps.py
```diff
@@ -774,21 +774,29 @@ def _clip_warp_output(input_image, output_image, mode, cval, clip):
         the image boundaries.
     clip : bool, optional
         Whether to clip the output to the range of values of the input image.
-        This is enabled by default, since higher order interpolation may
-        produce values outside the given input range.
+        If `order` > 1, this will be enabled by default, since higher order
+        interpolation may produce values outside the given input range.
 
     """
     if clip:
```
jakirkham, line 781: Should we be doing something to handle `clip is None` before this?

jakirkham (commented):
Thanks Greg! 🙏

Had one question below

python/cucim/src/cucim/skimage/transform/_warps.py
grlee77: I refactored slightly now to move the clip is None handling inside of this function and remove the `if clip:` indentation here. There is still one instance where `clip=None` also still needs to be checked externally outside of this function.

grlee77:
Rebased as I had accidentally merged a branch with a couple of stray commits.

PR #517, opened 2023-02-21
Binary morphology: omit weights array when possible

grlee77:
This is a follow-up to #485 allowing the overhead of creating a weights array of uniform values to be omitted when possible for all binary morphology operators. This gives a large performance benefit for small footprints (and small images) with the benefit becoming minimal for large footprint size.

As currently implemented, a shape tuple can be passed to `structure` to indicate the equivalent of an array of ones of that shape. This is a deviation from SciPy's API which does not currently allow this. I am not positive if this is the best API choice or if it might be better to instead add a `size` keyword-only argument like the grayscale morphology functions currently have.

The footprint generation methods `square`, `rectangle`, `cube` etc. use this new tuple format when possible to reduce overhead.

I will post benchmarks below.

### DIFF
--- a/python/cucim/src/cucim/skimage/morphology/footprints.py
+++ b/python/cucim/src/cucim/skimage/morphology/footprints.py
@@ -114,7 +114,7 @@ def footprint_from_sequence(footprints):
     return morphology.binary_dilation(imag, footprints)
 
 
-def square(width, dtype=cp.uint8, *, decomposition=None):
+def square(width, dtype=None, *, decomposition=None):
     """Generates a flat, square-shaped footprint.
 
     Every pixel along the perimeter has a chessboard distance
@@ -127,8 +127,11 @@ def square(width, dtype=cp.uint8, *, decomposition=None):
 
     Other Parameters
     ----------------
-    dtype : data-type, optional
-        The data type of the footprint.
+    dtype : data-type or None, optional
+        The data type of the footprint. When None, a tuple will be returned in
+        place of the actual footprint array. This can be be passed to grayscale
+        and binary morphology functions in place of an explicit array to avoid
+        array allocation overhead.
     decomposition : {None, 'separable', 'sequence'}, optional
         If None, a single array is returned. For 'sequence', a tuple of smaller
         footprints is returned. Applying this series of smaller footprints will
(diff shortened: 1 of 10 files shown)


grlee77:
# Benchmark results

Acceleration here is the ratio of the runtime when passing a boolean array of `footprint_shape` size vs. just passing the size tuple. 

These are worse case accelerations as they assume the footprint had already been created and don't include the initial overhead of allocating the footprint on the GPU.

Acceleration factors run from ~3x for small images approaching 1x as the image size becomes larger.

| shape | footprint_shape | acceleration |
|-------|-----------------|--------------|
| (256, 256) | (3, 3) | 3.04 |
| (256, 256) | (5, 5) | 3.06 |
| (256, 256) | (7, 7) | 2.90 |
| (256, 256) | (15, 15) | 2.91 |
| (256, 256) | (31, 31) | 2.88 |
| (512, 512) | (3, 3) | 3.03 |
| (512, 512) | (5, 5) | 2.77 |
| (512, 512) | (7, 7) | 2.88 |
| (512, 512) | (15, 15) | 2.89 |
| (512, 512) | (31, 31) | 2.82 |
| (1024, 1024) | (3, 3) | 2.57 |
| (1024, 1024) | (5, 5) | 2.22 |
| (1024, 1024) | (7, 7) | 2.73 |
| (1024, 1024) | (15, 15) | 2.58 |
| (1024, 1024) | (31, 31) | 2.08 |
| (2048, 2048) | (3, 3) | 1.83 |
| (2048, 2048) | (5, 5) | 1.55 |
| (2048, 2048) | (7, 7) | 1.66 |
| (2048, 2048) | (15, 15) | 1.47 |
| (2048, 2048) | (31, 31) | 1.31 |
| (4096, 4096) | (3, 3) | 1.25 |
| (4096, 4096) | (5, 5) | 1.14 |
| (4096, 4096) | (7, 7) | 1.18 |
| (4096, 4096) | (15, 15) | 1.12 |
| (4096, 4096) | (31, 31) | 1.08 |
| (32, 32, 32) | (3, 3, 3) | 2.99 |
| (32, 32, 32) | (5, 5, 5) | 2.87 |
| (32, 32, 32) | (9, 9, 9) | 2.88 |
| (32, 32, 32) | (13, 13, 13) | 2.86 |
| (64, 64, 64) | (3, 3, 3) | 2.63 |
| (64, 64, 64) | (5, 5, 5) | 2.85 |
| (64, 64, 64) | (9, 9, 9) | 2.89 |
| (64, 64, 64) | (13, 13, 13) | 2.87 |
| (128, 128, 128) | (3, 3, 3) | 1.63 |
| (128, 128, 128) | (5, 5, 5) | 2.17 |
| (128, 128, 128) | (9, 9, 9) | 1.94 |
| (128, 128, 128) | (13, 13, 13) | 1.83 |
| (256, 256, 256) | (3, 3, 3) | 1.09 |
| (256, 256, 256) | (5, 5, 5) | 1.12 |
| (256, 256, 256) | (9, 9, 9) | 1.10 |
| (256, 256, 256) | (13, 13, 13) | 1.09 |

python/cucim/src/cucim/skimage/morphology/footprints.py
```diff
@@ -144,19 +167,30 @@ def square(width, dtype=cp.uint8, *, decomposition=None):
     The 'sequence' decomposition mode only supports odd valued `width`. If
     `width` is even, the sequence used will be identical to the 'separable'
     mode.
+
```
gigony, line 170: unnecessary blank line?

gigony (approved):

grlee77:
rerun tests

grlee77:
@gigony, test failures in unrelated isotropic morphology functions have been fixed. I added checks to enforce properties of the provided output.

PR #528, opened 2023-03-28
Remove underscore in build string.

bdice:
This PR removes the underscore after "cuda" to align with other RAPIDS packages.


msadang (approved):
String change looks good.

bdice:
Needs a `/merge` from someone with write access. cc: @jakirkham

PR #540, opened 2023-04-13
protect against possible out of bounds memory access in 2D distance transform

### LINKED ISSUE #539: [BUG] CUDA_ERROR_ILLEGAL_ADDRESS with `morphology.isotropic_closing`
**Describe the bug**
I accidentally landed on a specific combination of conditions which causes illegal address access, resulting in a crash. I believe this has something to do with the time taken to write to memory, whilst another operation is happening. However, the timing required is incredibly inconsistent: sometimes it will happen repeatedly, then start working for a while, and go back to crashing.

Snippet 1 (MVP of my actual code):
```py
import cupy as cp
from cucim.skimage import morphology
x = cp.random.rand(1000, 1000)  # some large array simulating a 2D image
y = x.copy()
y = y >= 0.1
morphology.isotropic_closing(y, 3)
```
- all at once (in a function): crashes 5/5 times
- wait 10 seconds between `y = y >= 0.1` and `morphology.isotropic_closing(y, 3)`: crashes 5/5 times

Snippet 2:
```py
import cupy as cp
from cucim.skimage import morphology
x = cp.random.rand(1000, 1000)  # some large array simulating a 2D image
y = x.copy()
z = y >= 0.1
morphology.isotropic_closing(z, 3)
```
- doesn't crash

Snippet 3:
```py
import cupy as cp
from cucim.skimage import morphology
x = cp.random.rand(1000, 1000)  # some large array simulating a 2D image
y = x >= 0.1
morphology.isotropic_closing(y, 3)
```
- all at once (in a function): crashes 5/5 times
- wait 10 seconds between `y = x >= 0.1` and `morphology.isotropic_closing(y, 3)`: crashes 3/5 times

**Expected behavior**
All of the above snippets should run fine.

 - Environment location: HPC Cluster
 - Method of cuCIM install: conda, stable v23.02 and nightly v23.04a
 - GPU: Tesla V100-SXM2-32GB
 - CUDA: 11.4

**Additional context**
- This ONLY happens to `morphology.isotropic_closing`, `morphology.isotropic_opening` works fine
- `morphology.binary_closing` (i.e. `morphology.binary_closing(y, morphology.disk(3))`) works fine
-  When digging through the GitHub issues, I also found https://github.com/rapidsai/cucim/pull/421#issuecomment-1330140735, not sure if relevant.

grlee77:
closes #539

@accelyte, can you please check whether the following 1-line change resolves the illegal memory access issue you encountered? I could not reproduce it locally after making this change. It seems there was some location with negative coordinate access causing the out of range memory error.

I originally narrowed this down to `kernelColor` by putting manual device synchronization calls after each kernel in the `_pba_2d` function and the memory access issue reliably occurred for me in the synchronization after this kernel.


accelyte:
I can confirm that the fix works. Thanks for the quick resolution.

jakirkham (approved):

jakirkham:
Thanks Greg and Matthew! 🙏

Was there anything else we wanted to do here or should we merge?

grlee77:
> Was there anything else we wanted to do here or should we merge?

There was a second copy of this same kernel that was used when the user provides floating point "spacing" to account for pixel sizes. I didn't succeed in reproducing the same out of bounds issue for that version, but went ahead and added the same check there for consistency.

PR #545, opened 2023-04-18
add a 20 minute timeout for pytest runs on CI

grlee77:
Recently some CI runs seemed to hang for multiple hours with tests at 99% complete.

This PR adds a 20 minute timeout on the pytest run to avoid this issue. I also added a verbose flag so that if things get stuck we can identify which test case may be the cause.

### DIFF
--- a/ci/test_python.sh
+++ b/ci/test_python.sh
@@ -44,7 +44,7 @@ set +e
 
 rapids-logger "pytest cucim"
 pushd python/cucim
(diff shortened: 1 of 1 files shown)


ajschmidt8 (approved):

jakirkham (approved):

jakirkham:
Thanks Greg & AJ! 🙏

PR #547, opened 2023-04-22
Support Python 3.9 build/tests

shwina:


ajschmidt8:
For context, RAPIDS is dropping Python `3.8` support in favor of `3.9`.

This PR updates the `cucim` GHAs workflows to build from https://github.com/rapidsai/shared-action-workflows/tree/py-39, which contains updated build/test matrices for Python `3.9`.

This branch change will be reverted in a subsequent PR once all of RAPIDS is building and publishing `3.9` packages.

Issue #504, opened 2023-02-06
cuCIM: Need to canonicalize dlopen'd names


leofang:
Documenting an offline discussion with @jakirkham in preparation of CUDA 12 bring-up on conda-forge.

Currently, there're some places where `libXXXXX.so` without any SONAME being dlopen'd, for example,
https://github.com/rapidsai/cucim/blob/90d021eaf3cde68f73159db7b1b6b69d6cc27395/gds/src/cufile_stub.cpp#L64
There will become problematic because the `libXXXXX.so` symlink is supposed to exist only in the `libXXXXX-dev` package by the stand practice of Linux distros, not in `libXXXXX` which only contains `libXXXXX.so.{major}`. The dlopen'd names need to be canonicalized.

vyasr:
Please assign this issue to me.

jakirkham:
Done. Thanks Vyas! 🙏

jakirkham:
Submitted PR ( https://github.com/rapidsai/cucim/pull/548 ) in an attempt to address this borrowing heavily from cuDF and KvikIO

PR #548, opened 2023-04-26
Only load versioned `libcufile`

jakirkham:
Fixes the `dlopen` logic around `libcufile` to load either the SOVERSION library or the fully versioned library. This includes logic to handle the missing SOVERSION library on older CUDA versions. Drops loading the stub library, which typically only shows up in development environments.

<hr>

Fixes https://github.com/rapidsai/cucim/issues/504

Borrows from PR ( https://github.com/rapidsai/cudf/pull/13210 ) and PR ( https://github.com/rapidsai/kvikio/pull/203 ). Also adapts this logic to cuCIM ( https://github.com/rapidsai/kvikio/pull/141 )

### DIFF
--- a/gds/src/cufile_stub.cpp
+++ b/gds/src/cufile_stub.cpp
@@ -59,7 +59,7 @@ void CuFileStub::load()
 #if !CUCIM_STATIC_GDS
         if (handle_ == nullptr)
         {
-            // Note: Load the dynamic library with RTLD_NODELETE flag because libcufile.so uses therad_local which can
+            // Note: Load the dynamic library with RTLD_NODELETE flag because libcufile.so uses thread_local which can
             // cause a segmentation fault if the library is dynamically loaded/unloaded. (See #158)
             // CUDA versions before CUDA 11.7.1 did not ship libcufile.so.0, so this is
             // a workaround that adds support for all prior versions of libcufile.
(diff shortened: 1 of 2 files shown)


jakirkham:
@gigony could you please review?

gds/src/cufile_stub.cpp
```diff
@@ -61,7 +61,19 @@ void CuFileStub::load()
         {
             // Note: Load the dynamic library with RTLD_NODELETE flag because libcufile.so uses therad_local which can
```
bdice, line 62:
```suggestion
            // Note: Load the dynamic library with RTLD_NODELETE flag because libcufile.so uses thread_local which can
```

bdice (approved):
Changes look aligned with those in KvikIO and cuDF. 👍

grlee77 (approved):
Nice, thanks for handling this @jakirkham and @bdice

gigony (approved):
Thanks @jakirkham . Looks great to me! Thank you for adding this!

jakirkham:
Am seeing some test failures due to `test_disambiguate_2d` on Python 3.10 that appear unrelated to this change. @grlee77 could you please take a look at these?

grlee77:
Okay, I diagnosed the CI failure and it is due to a recent change in `imageio` that is causing the "eagle" image we are using to be loaded as RGB instead of grayscale.

The test will pass even with the most recent imageio if it is modified with an additional if statement guarding against the image being loaded as RGB instead of grayscale.
```py
    image = cp.array(eagle()[500:, 900:])  # use a highly textured image region
    # revert to grayscale if a recent imageio caused the image to be loaded as RGB
    if image.ndim == 3:
        image = image[..., 0]
```

grlee77:
I was looking into this failure in #549 and can clean that up to only have a necessary fix.

jakirkham:
Thanks Greg! 🙏

PR #553, opened 2023-05-02
Fix doc building via `run build_package`

grlee77:
Current pinnings in the requirements.txt for doc builds could not be satisfied (CuPy 9.3 appears to no longer be available on PyPI).

The key changes to those requirements are:
1.) use newer CuPy
2.) bump numpy to a version supported by CuPy
3.) bump sphinx to 4.5 so it is compatible with recent `jinja2`
4.) relax most other pinnings

A separate commit fixes several broken links discovered by the link checker during the doc builds.


gigony (approved):
Awesome! Thank you for updating links and typos!

PR #556, opened 2023-05-03
Handle space character in ./run download_testdata

gigony:
If the user's current path is having a space character (` `), `./run download_testdata` command doesn't work correctly.

This patch fixes the issue by wrapping paths with `"`.


grlee77 (approved):
Thanks, @gigony

PR #558, opened 2023-05-03
Update cupy dependency

vyasr:
Switch to requiring cupy 12


jakirkham:
Do we want to update the Conda recipe too?

https://github.com/rapidsai/cucim/blob/2207a690eff2bef2d084233ce20318ddc9547642/conda/recipes/cucim/meta.yaml#L48

AyodeAwe (approved):

jakirkham (approved):

jakirkham:
General question. In a few places (like in cuCIM) we had an upper bound. Do we want that going forward or are we wanting to drop that?

vyasr:
We decided to drop, see https://github.com/rapidsai/cugraph/pull/3539#pullrequestreview-1412087488

PR #563, opened 2023-05-26
Allow numpy 1.24.

bdice:
This removes the upper bound on `numpy<1.24` to allow newer NumPy versions.

PR #564, opened 2023-05-31
Fix SHA256 check failure in test suite

grlee77:
Due to a [recent change in the eagle image](https://gitlab.com/scikit-image/data/-/merge_requests/22) used in one of our test cases, skimage 0.19 fails to load it. As a solution, we can just restrict this one test to run only for `scikit-image>=0.20`.

Should resolve test failures seen in #563


bdice (approved):

jakirkham (approved):

jakirkham:
Thanks Greg! 🙏

PR #573, opened 2023-06-13
Sync cuCIM API with scikit-image 0.21

grlee77:
# Description

This MR ports several recent changes implemented upstream in scikit-image 0.21rc0. Some are just minor refactoring or documentation fixes, but I have commented on the more significant changes below so we can reuse that for the release note.

If preferred, I can split the items below into separate small PRs.

## API Changes / Deprecations
- The `random_state` argument of `medial_axis` and `unsupervised_wiener` is now deprecated and will be removed in the future. The new argument name, `seed`, should be used instead.

- The formerly private function `_invariant_denoise` has been renamed `denoise_invariant` and is now part of the public `cucim.skimage.restoration` API.

- The existing function `cucim.skimage.color.get_xyz_coords` has been renamed `cucim.skimage.color.xyz_tristimulus_values`. The former function name is deprecated and will be removed in the future.

## Enhancements
- A new `return_mapping` option was added to `join_segmentations`. This provides an additional output with a mapping between the labels in the joined segmentation and the original ones.

- Added support for y-axis shear to the [2D AffineTransform](https://github.com/scikit-image/scikit-image/pull/6752)

## Backwards Incompatible Changes
- A bug fix to the energy calculation in `chan_vese` may result in different output from previous versions.

## Bug Fixes
- Fixed [bug in energy calculation in Chan-Vese](https://github.com/scikit-image/scikit-image/pull/6902)

- Fixed a [bug in 2D shear calculations for `AffineTransform`](https://github.com/scikit-image/scikit-image/pull/6717)

### DIFF
--- a/python/cucim/src/cucim/skimage/_shared/utils.py
+++ b/python/cucim/src/cucim/skimage/_shared/utils.py
@@ -1,6 +1,7 @@
 import functools
 import inspect
 import numbers
+import sys
 import warnings
 from collections.abc import Iterable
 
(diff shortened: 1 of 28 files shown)


jakirkham:
Thanks Greg! 🙏

Does this depend on scikit-image 0.21 packages ( https://github.com/conda-forge/scikit-image-feedstock/pull/103 )?

gigony (approved):
Thanks @grlee77 This change looks good to me!

grlee77:
> Does this depend on scikit-image 0.21.0 packages ( https://github.com/conda-forge/scikit-image-feedstock/pull/103 )?

No, it should still work on older scikit-image as currently pinned

jakirkham:
Thank you both! 🙏

PR #574, opened 2023-06-19
Fix canny and butterworth (recent CuPy and NumPy compatibility)

grlee77:
It is possible in some cases for array scalar `low_threshold` to cause a failure to launch the non-maximum suppression kernel. This MR casts the array scalar to a float to avoid the issue.

I encountered this as a test failure locally with the existing `test_use_quantiles` test case for `canny` . I am unsure what underlying change caused this to suddenly start failing (possibly an underlying change in CuPy's kernel launching behavior).

A second issue with NumPy >= 1.25 compatibility for `butterworth` is also fixed in this MR. The `cupyx.scipy.fft` call raises an error in comparing an array fft_shape to a tuple shape internally. Casting `fft_shape` to a tuple avoids this.


jakirkham (approved):
LGTM. Thanks Greg! 🙏

PR #572, opened 2023-06-10
cuCIM: Build CUDA 12 packages

jakirkham:
Fixes https://github.com/rapidsai/cucim/issues/513
Fixes https://github.com/rapidsai/cucim/issues/582

Start building cuCIM for CUDA 12.

PR #578, opened 2023-07-14
Use linalg & inline `_get_manders_overlap_coeff`

jakirkham:
As most of the operations in `_get_manders_overlap_coeff` can be constructed from other linalg operations, use those directly to compute the result. Once this is done, there is less of a need to `fuse` it into a kernel. Instead those linalg functions can just be called as-is.

Note: This resolves a bug seen with `test_moc` in the CUDA 11.2 CI on the CUDA 12 PR  ( https://github.com/rapidsai/cucim/pull/572#issuecomment-1631610100 ) ( https://github.com/rapidsai/cucim/pull/572#discussion_r1264224233 )

### DIFF
--- a/python/cucim/src/cucim/skimage/measure/_colocalization.py
+++ b/python/cucim/src/cucim/skimage/measure/_colocalization.py
@@ -178,7 +178,7 @@ def manders_coloc_coeff(image0, image1_mask, mask=None):
 
 @cp.fuse()
 def _get_manders_overlap_coeff(image0, image1):
-    denom = cp.sqrt(cp.sum(cp.square(image0)) * (cp.sum(cp.square(image1))))
+    denom = cp.linalg.norm(image0) * cp.linalg.norm(image1)
     return cp.sum(cp.multiply(image0, image1)) / denom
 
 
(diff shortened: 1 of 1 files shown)


jakirkham:
@grlee77 would be curious to hear your thoughts on this approach 🙂

python/cucim/src/cucim/skimage/measure/_colocalization.py
```diff
@@ -261,7 +255,8 @@ def manders_overlap_coeff(image0, image1, mask=None):
     if image1.min() < 0:
         raise ValueError("image1 contains negative values")
 
-    return _get_manders_overlap_coeff(image0, image1)
+    denom = cp.linalg.norm(image0) * cp.linalg.norm(image1)
+    return cp.vdot(image0, image1) / denom
```
jakirkham, line 259: We could go further here and do

```python
1 - cupyx.scipy.spatial.distance.cosine(image0, image1)
```

This was added in PR ( https://github.com/cupy/cupy/pull/7063 ) and would require CuPy 12.0.0+ & pylibraft from RAPIDS

grlee77: Looks like we have to ravel to 1D to use the `cosine` solution and then do a subtraction to get the equivalent result of the linalg implementation:
```py
1 - cosine(image0.ravel(), image1.ravel())
```

If the RAFT solution is faster we can provide it with fallback to the linalg one here when unavailable.

grlee77 (approved):
Thanks, @jakirkham, looks great!

python/cucim/src/cucim/skimage/measure/_colocalization.py
jakirkham: Sounds good. Thanks Greg! 🙏

PR #579, opened 2023-07-17
Drop unneeded selector on `libwebp-base`

jakirkham:
This selector only allows `libwebp-base` on Linux & macOS builds. However the `libcucim` package is only built on Linux and there is no macOS support. So this selector is unneeded. Given this go ahead and drop it.


raydouglass (approved):

grlee77 (approved):

jakirkham:
Thanks all! 🙏

Issue #586, opened 2023-07-26
Disable cuFile / GDS on SBSA ( `aarch64` )


jakirkham:
As there are no cuFile / GDS binaries for SSBA ( `aarch64` ) currently, it would be good to disable downloading cuFile (or needing to build with it) in the SSBA ( `aarch64` ) case.

gigony:
Closing this because some portion of this issue is addressed by https://github.com/rapidsai/cucim/pull/590 .
Users will see an error message when cuFILE-related API is called while libcufile is not available.

PR #590, opened 2023-07-27
Do not use x86_64 GDS binaries for aarch64

gigony:
Although there are no cuFile / GDS binaries for SSBA ( aarch64 ) currently, it still requires 'cufile.h' file to create a libcufile_stub library that loads libcufile dynamically.

This patch still download GDS package but do not copy GDS libraries if the architecture is `aarch64`.

Address #586

### DIFF
--- a/run
+++ b/run
@@ -234,6 +234,30 @@ retry() {
     return 0
 }
 
+get_arch() {
+    local platform="${1:-$(uname -p)}"
+    local platform_str
+
+    case "${platform}" in
+        amd64|x86_64|x86|linux/amd64)
+            platform_str="x86_64"
+            ;;
+        arm64|aarch64|arm|linux/arm64)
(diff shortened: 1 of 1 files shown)


run
```diff
@@ -234,6 +234,30 @@ retry() {
     return 0
 }
 
+get_arch() {
+    local platform="${1:-$(uname -p)}"
```
grlee77, line 238: `uname -m` may be more portable?
I think it will mostly be `x86_64` or `aarch64`, but it looks like there are other arm-related possibilities listed here: https://stackoverflow.com/questions/45125516/possible-values-for-uname-m

grlee77 (approved):
It looks good to me, thanks Gigon!

jakirkham (approved):

jakirkham:
Thanks Gigon! 🙏

Issue #595, opened 2023-08-02
Flaky test: `test_read_region_cuda_memleak`


jakirkham:
Previously we saw ( https://github.com/rapidsai/cucim/pull/593#issuecomment-1662504461 ) this failure [on CI]( https://github.com/rapidsai/cucim/actions/runs/5740435207/job/15558989875?pr=593#step:7:20114 )

```python
=================================== FAILURES ===================================
________________________ test_read_region_cuda_memleak _________________________
[gw1] linux -- Python 3.10.12 /opt/conda/envs/test/bin/python3.10
tests/performance/clara/test_read_region_memory_usage.py:44: in test_read_region_cuda_memleak
    assert mem_usage_history[4] - mem_usage_history[1] < 180.0
E   assert (14623.0 - 14257.0) < 180.0
----------------------------- Captured stdout call -----------------------------
[12999.0, 14257.0, 14569.0, 14617.0, 14623.0, 14621.0]
```

In discussion offline it sounds like this test can be a bit flaky. Wondering if there are ways we can make it less flaky or if we should just `skip` it

gigony:
Addressed by #601 
Closing.

PR #601, opened 2023-08-03
Postponed the assessment of GPU memory for testing

gigony:
Since the GPUtil library does not allow us to measure the GPU memory usage for a specific process, we need to take a more cautious approach to gauge GPU usage.

There's a likelihood that the GPU memory usage may not stabilize during the initial iterations, as illustrated below:
```
[12999.0, 14257.0, 14569.0, 14617.0, 14623.0, 14621.0]
```
Rather than evaluating `mem_usage_history[4] - mem_usage_history[1] < 180.0`, we will now measure 10 points, and begin our examination from the midpoint of these points (`mem_usage_history[5] - mem_usage_history[9] < 180.0`).

This change addresses #595.

### DIFF
--- a/python/cucim/tests/performance/clara/test_read_region_memory_usage.py
+++ b/python/cucim/tests/performance/clara/test_read_region_memory_usage.py
@@ -30,7 +30,7 @@ def test_read_region_cuda_memleak(testimg_tiff_stripe_4096x4096_256_jpeg):
     gpu = gpus[0]
     mem_usage_history = [gpu.memoryUsed]
 
-    for i in range(5):
+    for i in range(10):
         _ = img.read_region(device='cuda')
         gpus = GPUtil.getGPUs()
         gpu = gpus[0]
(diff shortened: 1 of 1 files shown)


python/cucim/tests/performance/clara/test_read_region_memory_usage.py
```diff
@@ -40,8 +40,10 @@ def test_read_region_cuda_memleak(testimg_tiff_stripe_4096x4096_256_jpeg):
 
     # The difference in memory usage should be less than 180MB.
     # Note: Since we cannot measure GPU memory usage for a process,
-    #       we use a rough number (experimentally measured).
-    assert mem_usage_history[4] - mem_usage_history[1] < 180.0
+    #       we use a rough number.
+    #       (experimentally measured, assuming that each image load
+    #        consumes around 50MB of GPU memory).
+    assert mem_usage_history[5] - mem_usage_history[9] < 180.0
```
jakirkham, line 46: Were we wanting to increase the `180.0` limit as well?

jakirkham (commented):
Thanks Gigon! 🙏

Had a question below

python/cucim/tests/performance/clara/test_read_region_memory_usage.py
gigony: Hi @jakirkham ,

Since each method call is expected to increase ~50MB of GPU memory, it is possible that we may miss the actual memory leak if the value is more than 200MB (50MB x 4).

For this reason, instead of increasing the value, this chooses the 6th and 10th samples, instead of the 2nd and 5th samples for comparison.

jakirkham: Ok let's give this a try then and see how it goes

jakirkham (approved):

jakirkham:
Thanks Gigon! 🙏

PR #615, opened 2023-10-24
remove versioneer

grlee77:
remove versioneer to more closely follow RAPIDS conventions


python/cucim/src/cucim/__init__.pyi
```diff
@@ -27,9 +27,6 @@ try:
     _is_clara_available = True
     submodules += ['clara']
 except ImportError:
-    from ._version import get_versions
-    __version__ = get_versions()['version']
-    del get_versions
-    del _version
+    __version__ = "23.12.00"
```
gigony, line 30: Would be better if we could setup a mechanism to update this hard-corded value.

jakirkham: The [`update-version.sh`]( https://github.com/rapidsai/cucim/blob/branch-23.12/ci/release/update-version.sh ) is used to do this. Though we may need to update the logic in that script to perform replacement in the right files

jakirkham: For example [`rmm`'s `update-version.sh` has this line]( https://github.com/rapidsai/rmm/blob/39800d3d4aeff80de36d7872ed551243518e25a8/ci/release/update-version.sh#L40 ), which we might want to borrow

raydouglass (approved):

gigony (approved):
Looks good to me!

Issue #23, opened 2021-05-05
[FEA] Investigate using black


jakirkham:
Previously we tried to use black for formatting Python, but ran into some issues. It would be good to revisit this at some point

cc @grlee77 @quasiben

grlee77:
I think the two issues are:
1.) unless `fmt:off` / `fmt:on` blocks are used, it can make some small test matrices formatted in a less readable way.
2.) since upstream scikit-image is not formatted with black, it makes direct comparison with upstream code in diff tools a bit more difficult

jakirkham:
Yeah the matrix point makes sense. Disabling formatting on them seems fine

On 2 it might be worth asking scikit-image if they are willing to use black. We started using this a while back in Dask + Distributed as well as RAPIDS, which has helped keep things readable and consistent

grlee77:
> On 2 it might be worth asking scikit-image if they are willing to use black.

It did come up in the past in https://github.com/scikit-image/scikit-image/issues/4111 and wasn't received to favorably at the time. Maybe people are more familiar with it by now, though?

I am not too strongly against using it here even if upstream does not. We can always apply black to the scikit-image source prior to doing a diff if we need to compare files directly across the two. For specific files like `colorconv.py` where we added custom CUDA kernel code, the diff will already be pretty large in either case.

jakirkham:
Thanks for the context. Yeah I think black has also softened on some things, which helps, and more projects have started to use it, which also helps issue discovery & subsequent fixes

That said, if it is going to make it hard to coordinates changes between here and scikit-image. It's probably not worth the effort until scikit-image adopts it

PR #616, opened 2023-10-24
update linters and move their configurations from setup.cfg to pyproject.toml

grlee77:
The first commit here is from #615

This MR makes multiple linting-related changes to cuCIM
- moves existing `isort` config from `setup.cfg` to a new `pyproject.toml` file
- update `isort` conventions to match other RAPIDS projects
- add `black` and apply it (use `fmt::off`/`fmt::on` as needed to preserve formatting of look-up tables and test cases)
- replace use of `flake8` with `ruff`
- add config for `codespell` and apply recommended fixes across both C++ and Python code
- update CONTRIBUTING.md instructions to reflect these changes
- Pre-commit hooks for `black`, `ruff` and `codespell` were added. The pre-commit hook for `flake8` was removed.

The only non-linting change is
- move pytest configuration from `setup.cfg` to `pyproject.toml`

I plan to leave any build-system related changes to `pyproject.toml` to a follow-up MR. This MR focuses only on linting/testing.

It will probably be easiest to review the individual commits. Some, like applying `black` are quite large.

Fixes https://github.com/rapidsai/cucim/issues/23


grlee77:
Most changes here are automated. 

Please closely review the following manually updated files:

- `python/cucim/setup.cfg`
- `python/cucim/pyproject.toml`
- `.pre-commit-config.yaml`

grlee77:
Had to rename `pyproject.toml` to `pyproject_.toml` temporarily in this MR so it only gets used for linting steps, but isn't used during package build time.

grlee77:
After this MR `black`, `ruff`, `isort` and `codespell` would all be automaticaclly run locally on each commit via `pre-commit` if the developer has previously run the following in their repository:

```
pre-commit install
``` 


By default that only runs the checks on the files being modified/added in the commit. For testing purposes, I forced pre-commit to run on all files using:
```
pre-commit run --all-files
```
to verify that no problems are reported.

.pre-commit-config.yaml
```diff
@@ -15,10 +15,23 @@ repos:
...
+  - repo: https://github.com/psf/black
+    rev: 23.3.0
     hooks:
-      - id: flake8
-        args: ["--config=python/cucim/setup.cfg"]
-        files: ^python/cucim/.*
+      - id: black
+        files: (python|legate)/.*
+        args: ["--config", "python/cucim/pyproject_.toml"]
+  - repo: https://github.com/astral-sh/ruff-pre-commit
```
bdice, line 25: I won't advocate for changes in this PR due to its size, but for some potential follow-up work, you might want to update the linter versions (all RAPIDS repos' linters are independent, you don't have to use a specific isort/black/ruff version). Also you can use `ruff` in place of `isort` if you enable the `I` category, and you can enable auto-fixes as well, in `pyproject.toml`. Enabling auto-fixes plays very nicely with pre-commit.

```
[tool.ruff]
select = ["E", "F", "W", "I"]
fix = true
```

bdice (approved):

.pre-commit-config.yaml
grlee77: Sure. I also discussed with @jakirkham and @gigony that we can add a `.git-blame-ignore-revs` in a follow-up so all these linting changes [don't show up in the blame view on GitHub](https://github.blog/changelog/2022-03-24-ignore-commits-in-the-blame-view-beta/).

raydouglass (approved):
`ops-codeowners` changes look good

gigony (approved):

jakirkham (approved):

jakirkham:
Thanks Greg and everyone for the reviews! 🙏

.pre-commit-config.yaml
jakirkham: Greg opened a follow-up PR: https://github.com/rapidsai/cucim/pull/622

Please let us know if you have thoughts there 🙂

PR #617, opened 2023-10-24
Replace setup.py with pyproject toml

grlee77:
This MR is on top of #615 and #616 so those should be reviewed and merged first.

Currently `run build_python_package` will be in a broken state.


jakirkham:
Thanks Greg! 🙏

Would add `pyproject` to lines in `dependencies.yaml` like these that contain both `conda` & `requirements`

https://github.com/rapidsai/cucim/blob/5310870ae56ad5d07c58035526ed180383602fcf/dependencies.yaml#L41

grlee77:
> Would add pyproject to lines in dependencies.yaml like these that contain both conda & requirements

Thanks, @jakirkham! Please see new commit 727010a related to this. I now understand how the `pyproject.toml` sections can be generated based on that `dependencies.yaml` using `rapids-dependency-file-generator`. It is a pretty nice way of making sure all of these requirement lists (requirements, pyproject, conda) stay in sync! 

I also noticed that I must have mistakenly copied extra docs requirements from scikit-image in the earlier commits. The ones that were already in `dependencies.yaml` were sufficient for cuCIM.

## One point for discussion
Currently we have a bit of a non-standard build process where the `pyproject.toml` doesn't know anything about the C++ library and does not handle building it. Instead, it assumes that C++ library was previously built and the resulting `.so` file had already been copied into the `python/cucim/src/cucim/clara/` folder. Then that pre-compiled file just gets copied into the package during a pure Python build (i.e. as in this [build.sh](https://github.com/rapidsai/cucim/blob/pull-request/617/conda/recipes/cucim/build.sh) from the conda recipe). This behavior allows building the Python package even without ever copying the libcucim `.so`, in which case the user can still use the `cucim.core` and `cucim.skimage` modules, but `cucim.clara` would fail to import since it will try to load the missing libcucim `.so` file.

Supporting this non-standard pure Python build is why I moved `cucim.clara`-sepcific runtime dependencies under the "optional" dependencies for tests (added comments related to this in `dependencies.yaml`).

gigony (approved):
Looks good to me!

raydouglass (approved):

jakirkham (approved):

jakirkham:
Thanks all! 🙏

PR #618, opened 2023-10-25
Update packages (pybind11 and catch2) and do not use nvidia-docker command

gigony:
### Update Catch2 to v3.4.0

Without upgrading Catch2, The following error occurs when building on
Ubuntu 22.04 due to glibc:
```
cucim/build-debug/_deps/deps-catch2-src/single_include/catch2/catch.hpp:10830:58: error: call to non-‘constexpr’ function
‘long int sysconf(int)’
10830 |     static constexpr std::size_t sigStackSize = 32768 >=  MINSIGSTKSZ ? 32768 : MINSIGSTKSZ;
```
### Update pybind11 to v2.11.1

Even with the latest version of pybind11, we still have an issue
with `pybind11::array_t` when cuCIM is used in multithread without
importing numpy in the main thread.

See https://github.com/pybind/pybind11/pull/4877

Will need to wait for the next release of pybind11.

### Use runtime option instead of using nvidia-docker command

nvidia-docker binary is not available if user doesn't install
nvidia-docker2 package. This change uses runtime option instead
of using nvidia-docker command.

### Apply pybind11 patch to avoid deadlock (until new release is available)

This applies the following patches to pybind11:

- https://github.com/pybind/pybind11/pull/4857
- https://github.com/pybind/pybind11/pull/4877

to avoid deadlock when using pybind11 without importing numpy in
multi-threaded environment.


grlee77:
(@gigony, I sent you a slack message with info on how to comply with newer RAPIDS signing requirements, so adding "ok to test" comments won't be required for CI to run)

grlee77 (approved):
Thanks, @gigony. It looks good to me.

jakirkham (approved):

jakirkham:
Thanks all! 🙏

Issue #305, opened 2022-06-06
Updating cuCIM's build/CI infra to use `scikit-build`, `rapids-cmake`, `whl` builds, etc.


jakirkham:
Other RAPIDS libraries have been moving to [scikit-build]( https://scikit-build.readthedocs.io/en/latest/ ) for Python packages, this integrates nicely with CMake and can simplify the build process (as well as making other package builds like wheels easier). Would be good to investigate doing the same for cuCIM

cc @vyasr (who has some experience here)

vyasr:
If cuCIM moves forward with this transition I would recommend also making use of rapids-cmake, which has a lot of the helper functions I wrote to simplify the process.

jakirkham:
This PR ( https://github.com/rapidsai/cudf/pull/10919 ) making this change to cuDF could serve as a good example

jakirkham:
Maybe a simpler example case is this KvikIO PR ( https://github.com/rapidsai/kvikio/pull/32 )

bdice:
@jakirkham requested that I share additional info about wheel builds here:

cuCIM builds wheels in a different way than most of RAPIDS. The [guide for RAPIDS wheel builds](https://github.com/rapidsai/ops/issues/2572) says to remove versioneer and migrate to pyproject.toml. If migrating to a build system that is more similar to the rest of RAPIDS, cuCIM should remove its [usage of versioneer](https://github.com/rapidsai/cucim/blob/branch-23.08/python/cucim/versioneer.py). Another part of this migration would be to move from the [current setup.py](https://github.com/rapidsai/cucim/blob/branch-23.08/python/cucim/setup.py) to pyproject.toml.

bdice:
Additional context on rapids-cmake: I noticed a lot of CMake dependencies that are duplicating pinnings and features that are handled by rapids-cmake.

Examples of dependencies supplied by rapids-cmake: https://docs.rapids.ai/api/rapids-cmake/stable/api/#cpm-pre-configured-packages

This includes:
- fmt: https://github.com/rapidsai/cucim/blob/branch-23.10/cpp/cmake/deps/fmt.cmake
- benchmark: https://github.com/rapidsai/cucim/blob/branch-23.10/cpp/cmake/deps/googlebenchmark.cmake
- gtest: https://github.com/rapidsai/cucim/blob/branch-23.10/cpp/cmake/deps/googletest.cmake
[...]

PR #619, opened 2023-10-26
build wheels on CI

grlee77:
This MR includes the commits from #617, so that one should be merged first.

The purpose of this MR is to build and test wheels on RAPIDS CI infrastructure. Previously, cuCIM was using a custom manual wheel building process based on running:
```
./run build_python_package
```
which used `dockcross`/`cibuildwheel`. We would then manually upload the resulting wheel to PyPI using `twine`.

The changes in this MR are to enable automated wheel builds for future RAPIDS releases in a similar way to other RAPIDS projects.

### Misc issues/questions for discussion
- The `build_wheels.sh` script installs the YASM and the openslide libraries via system package managers. This is currently hard-coded in that script rather than using the `dependencies.yaml` mechanism being used elsewhere. Is this reasonable or is there a better way?
- Following cuML's example, I added a separate `ci/release/apply_wheel_modifications.sh` script that updates the pyproject.toml as needed (to fixup version string and the CUDA version strings in the cucim package name and CuPy dependency)
- I updated the cuCIM test suite to make a handful of tests that relied on `imagecodecs` to be skipped if it is not installed. This is because I did not find a pre-compiled wheel for it on aarch64 and when it tried to compile from source instead, it failed due to various missing system packages needed for the compilation.
- Unlike most RAPIDS projects, which seem to use scikit-build, this project performs a separate CMake build via `run build_local` commands and copies the C++ libraries into the Python source tree. The `pyproject.toml` then just does a standard pure-python build, but will package the previously compiled libraries along with the other files.
- The wheel tests are currently taking approximately 35-50 minutes to run. It should be possible to reduce that time by an order of magnitude if the CuPy kernels are cached across runs. Does any RAPIDS projects cache the compiled CuPy kernels in this way? (cached kernels end up in `~/.cupy/kernel_cache/` by default, although an alternative location can be specified via `CUPY_CACHE_DIR`)


ci/release/apply_wheel_modifications.sh
divyegala, line 1: This should be baked into `ci/build_wheel.sh`. See https://github.com/rapidsai/rmm/blob/5bdcc05683893c528287d84e4a804b96896541d6/ci/build_wheel.sh#L6-L26 for reference

divyegala (commented):
We are in the process of changing versioning strategy across other RAPIDS repos. Please see https://github.com/rapidsai/rmm/pull/1347 for reference, and we can get those changes in this PR itself

ci/build_wheel.sh
```diff
@@ -0,0 +1,65 @@
...
+RAPIDS_PY_CUDA_SUFFIX="$(rapids-wheel-ctk-name-gen ${RAPIDS_CUDA_VERSION})"
+
+ci/release/apply_wheel_modifications.sh ${version_override} "-${RAPIDS_PY_CUDA_SUFFIX}"
+echo "The package name and/or version was modified in the package source. The git diff is:"
+git diff
+
+pip install --upgrade pip
+
+#CMake version in the container is too old, install a new version in the python venv
+pip install "cmake>=3.26.4" ninja
```
bdice, line 27: Let's manage dependencies with `dependencies.yaml`, and list these as a build system requirement in `pyproject.toml`. See how cudf does this here:

This code...
https://github.com/rapidsai/cudf/blob/abc0d41d1d9033d581948ae19384e0aa0f33da77/dependencies.yaml#L173-L176

...generates this code:
https://github.com/rapidsai/cudf/blob/abc0d41d1d9033d581948ae19384e0aa0f33da77/python/cudf/pyproject.toml#L3-L16

grlee77: I installed it here because we are calling CMake manually via the `run build_local` commands below. `pyproject.toml` isn't being used until later for the `pip wheel` call, but we need the CMake update for the earlier `run` script steps to succeed.

We could look at refactoring to use scikit-build, but I'm not sure how much work it would be.

ci/release/apply_wheel_modifications.sh
grlee77: Thanks, I will take a look at adapting that approach tomorrow.

ci/build_wheel.sh
bdice: Okay. Instead, use rapids-dependency-file-generator to create a requirements.txt file containing the build dependencies and install from that. We just want a single source for version information.

bdice: Eventually I think migrating to scikit-build would be good for alignment with the rest of RAPIDS.

ci/release/apply_wheel_modifications.sh
grlee77: Pushed a couple of new commits related to version strings
removed separate script in: [2791228](https://github.com/rapidsai/cucim/pull/619/commits/27912280e8492c3ffd84165f4830d826bf755e5c)
switch to the proposed new-style version management in: [dff57f8](https://github.com/rapidsai/cucim/pull/619/commits/dff57f8bcc72c95e6b80c25759c5df79063d642d)

grlee77: @gigony: I didn't realize until today about how the VERSION files were being used by CMake and that CMake would not allow alpha-versions in the VERSION string (e.g. `23.12.00` is valid for CMake but not something like `23.12.00a45`). My changes in dff57f8 broke things and it took a number of iterations to get it working again. In summary, it seems there were a few separate version files C++ CMake was using
- The top-level `CMAKELISTS.txt` was using the top-level `VERSION` file
- The cuslide plugin build was using `cpp/plugins/cucim.kit.cumed/VERSION`
- The cumed plugin build was using `cpp/plugins/cucim.kit.cumed/VERSION`
- The pybind11 Python bindings code was also using the top-level `VERSION`

I have updated so all of these use the same top-level VERSION_CPP which is based on VERSION, but with any "alpha" details stripped from the version string so CMake is happy. I suppose, in theory, with the old setup, the plugins could be versioned independently of the main library, but it didn't seem we were currently going that?

ci/build_wheel.sh
jakirkham: This is tracked in issue: https://github.com/rapidsai/cucim/issues/305

grlee77:
I think this recent failure will be fixed by merging #623:
https://github.com/rapidsai/cucim/actions/runs/6711052263/job/18238149350?pr=619

ci/build_wheel.sh
bdice: I think you want to do this:
```suggestion
# Install pip build dependencies (not yet using pyproject.toml)
rapids-dependency-file-generator \
  --file_key "py_build" \
  --output "requirements" \
  --matrix "cuda=${RAPIDS_CUDA_VERSION%.*};arch=$(arch);py=${RAPIDS_PY_VERSION}" | tee build_requirements.txt
pip install -r build_requirements.txt
```

dependencies.yaml
```diff
@@ -245,32 +245,31 @@ dependencies:
...
-          - imagecodecs>=2021.6.8
-          - openslide-python>=1.1.2
           - psutil>=5.8.0
           - pytest-cov>=2.12.1
           - pytest-lazy-fixture>=0.6.3
           - pytest-xdist
           - pytest>=6.2.4
           - tifffile>=2022.7.28
+          - pooch>=1.6.0  # needed to download scikit-image sample data
+          - matplotlib
```
bdice, line 255: FYI: matplotlib has two conda package flavors, `matplotlib-base` and `matplotlib`. The `matplotlib` package includes a bunch of extra GUI bits which inflate the total size (with dependencies) quite a lot. You might only need `matplotlib-base` here, but I don't know how it's being used.

grlee77: It is only used in a single test case. I think it can probably work with `matplotlib-base`

ci/build_wheel.sh
grlee77: Thanks, it looks good!

dependencies.yaml
```diff
@@ -252,10 +252,10 @@ dependencies:
           - pytest>=6.2.4
           - tifffile>=2022.7.28
           - pooch>=1.6.0  # needed to download scikit-image sample data
-          - matplotlib
+          - matplotlib-base
```
bdice, line 255: `matplotlib-base` doesn't exist on PyPI so it won't work for `requirements` or `pyproject` outputs. You'll need to list `matplotlib-base` in a `conda` output and `matplotlib` for `[requirements, pyproject]`.
```suggestion
```

dependencies.yaml
```diff
@@ -252,10 +252,10 @@ dependencies:
...
       - output_types: [conda]
         packages:
-          - imagecodecs>=2021.6.8
+          # - imagecodecs>=2021.6.8
           - openslide-python>=1.1.2
```
bdice, line 259:
```suggestion
          - matplotlib-base
          - openslide-python>=1.1.2
```

dependencies.yaml
```diff
@@ -264,11 +264,12 @@ dependencies:
...
-          - openslide-python>=1.1.2; platform_machine=='x86_64'
+      # temporarily remove imagecodecs / openslide-python from wheel tests
+      # - output_types: [pyproject]
+      #   packages:
+      #     # skip packages on arm64 that don't provide a wheel
+      #     - imagecodecs>=2021.6.8; platform_machine=='x86_64'
+      #     - openslide-python>=1.1.2; platform_machine=='x86_64'
       - output_types: [requirements, pyproject]
         packages:
           - opencv-python-headless>=4.6
```
bdice, line 275:
```suggestion
          - matplotlib
          - opencv-python-headless>=4.6
```

grlee77: thanks, I pushed this fix

CONTRIBUTING.md
```diff
@@ -263,19 +263,21 @@ conda install -y -c ${CONDA_BLD_DIR} -c conda-forge \
...
 
+If you are using CUDA 12.x, please update pyproject.toml as follows before building the wheel
 ```bash
-./run build_package
+sed -i "s/cupy-cuda11x/cupy-cuda12x/g" python/cucim/pyproject.toml
 ```
+This will switch the CuPy dependency to one based on CUDA 12.x instead of 11.x.
 
-The command would use `./temp` folder as a local build folder and build a distribution package into `dist` folder using [dockcross](https://github.com/dockcross/dockcross)'s manylinux2014 docker image.
+The wheel can then be build using.
```
bdice, line 274:
```suggestion
The wheel can then be built using:
```

ci/wheel_smoke_test.py
bdice, line 1: It seems like this file is not covered by the `black` pre-commit hook because it's not in `python/` or `legate/`. I would consider enabling `black` for all Python files in the repo, and remove the reference to `legate/` because cucim doesn't have a `legate/` folder as far as I can tell. https://github.com/rapidsai/cucim/blob/5e25641bce2e52595eae160802d02ebd3e6de911/.pre-commit-config.yaml#L17

dependencies.yaml
```diff
@@ -245,32 +245,33 @@ dependencies:
...
           - psutil>=5.8.0
           - pytest-cov>=2.12.1
           - pytest-lazy-fixture>=0.6.3
           - pytest-xdist
           - pytest>=6.2.4
           - tifffile>=2022.7.28
+          - pooch>=1.6.0  # needed to download scikit-image sample data
       - output_types: [conda]
         packages:
+          - imagecodecs>=2021.6.8
```
bdice, line 257: It seems like `imagecodecs` is pinned the same way for `conda` and `requirements`. Do you want to list that above in the `[conda, requirements, pyproject]` section?

dependencies.yaml
```diff
@@ -245,32 +245,33 @@ dependencies:
...
+          - matplotlib-base
+          - openslide-python>=1.1.2
```
bdice, line 259: Like `imagecodecs` above, it seems like `openslide-python` is pinned the same way for `conda` and `requirements`. Do you want to list that above in the `[conda, requirements, pyproject]` section?

Note that in most cases you want the same dependencies in `requirements` and `pyproject` because they both install from PyPI, so I'm questioning why we have both `[requirements]` and `[requirements, pyproject]` sections. I expect we really only need three sections: `[conda, requirements, pyproject]` for dependencies pinned the same way in conda/pip, `[conda]` for conda-only dependencies, and `[requirements, pyproject]` for pip-only dependencies.

dependencies.yaml
```diff
@@ -245,32 +245,33 @@ dependencies:
...
           - pip
           - pip:
             - opencv-python-headless>=4.6
+      - output_types: [requirements]
+        packages:
+          - imagecodecs>=2021.6.8
+          - openslide-python>=1.1.2
+      # temporarily remove imagecodecs / openslide-python from wheel tests
```
bdice, line 267: Maybe this code comment affects the review comments I left above. I don't see manylinux ARM builds of imagecodecs. I'll leave it to you to figure out the best course of action here.

bdice (approved):
Approving. I have a few comments but this is the last pass of review I think I need to give.

ci/wheel_smoke_test.py
grlee77: Good catch. I just removed this files option altogether since it already has a selection for Python files in the `pyproject.toml`

```toml
include = '\.py?$'
```

dependencies.yaml
grlee77: Yeah, I need to open an issue to fix running tests when `imagecodecs` is installed (there is a single test case that fails on CI, but I couldn't reproduce using a wheel built on my local system). There was also an issue on arm64 where it tries to compile from source, but fails (only x86_64 wheels were available)

I went ahead and merged the requirements and pyproject sections and just comment it out in for now.

divyegala (approved):
`VERSION_CPP` is an interesting solution. It does seem to be more concise than what we are doing elsewhere

ci/release/update-version.sh
```diff
@@ -34,12 +35,12 @@ function sed_runner() {
...
-sed_runner "s/^version = .*/version = \"${NEXT_FULL_TAG}\"/g" python/cucim/pyproject.toml
-sed_runner "s/${CURRENT_LONG_TAG}/${NEXT_FULL_TAG}/g" VERSION
-sed_runner "s/${CURRENT_LONG_TAG}/${NEXT_FULL_TAG}/g" python/cucim/VERSION
-sed_runner "s/__version__ = .*/__version__ = \"${NEXT_FULL_TAG}\"/g" python/cucim/src/cucim/__init__.pyi
-sed_runner "s/${CURRENT_LONG_TAG}/${NEXT_FULL_TAG}/g" cpp/plugins/cucim.kit.cuslide/VERSION
-sed_runner "s/${CURRENT_LONG_TAG}/${NEXT_FULL_TAG}/g" cpp/plugins/cucim.kit.cumed/VERSION
+# Centralized version file update
+echo "${NEXT_FULL_TAG}" > VERSION
+# update VERSION files used by C++/CMake (TODO: avoid these duplicate copies)
+NEXT_CPP_LONG_TAG=${NEXT_MAJOR}.${NEXT_MINOR}.${NEXT_PATCH}  # no leading 'v' or trailing 'a'
```
raydouglass, line 41: This is the same thing as `NEXT_FULL_TAG`

.github/workflows/build.yaml
```diff
@@ -66,3 +66,23 @@ jobs:
       branch: ${{ inputs.branch }}
       date: ${{ inputs.date }}
       sha: ${{ inputs.sha }}
+  wheel-build:
+    needs: [cpp-build]
```
raydouglass, line 70: Does the `wheel-build` need to depend on the `cpp-build` (conda)?

ci/build_wheel.sh
```diff
@@ -0,0 +1,74 @@
...
+# Also need to install openslide dev libraries on the system
+if ! command -v apt &> /dev/null
+then
+    echo "apt package manager not found, attempting to use yum"
+    yum install yasm openslide-devel -y
+else
+    echo "apt package manager was found"
+    apt update
+    apt install yasm libopenslide-dev -y
+fi
```
raydouglass, line 61: These dependencies should be installed in the CI image instead: https://github.com/rapidsai/ci-imgs/blob/main/ci-wheel.Dockerfile

gigony: @raydouglass, I don't think these (yasm and libopenslide-dev) are common to all RAPIDS projects. Do you think we still need to update the umbrella ci-wheel.Dockerfile? In particular, `libopenslide-dev` is cucim-specific and only needed for test cases in C++ while building libcucim (not needed for Python wheel).

raydouglass: Yes. They are small packages and relatively independent. We already do this (albeit in a different image & layer) for things like `libnuma-dev`.

jakirkham: Adding with PR: https://github.com/rapidsai/ci-imgs/pull/90

jakirkham, line 61:
```suggestion
```

jakirkham: Next step would be to drop these lines: https://github.com/rapidsai/cucim/pull/619#discussion_r1382176217

.github/workflows/build.yaml
grlee77: Is this expected to work or would there be binary incompatibilities? Currently I am just rebuilding the C++ lib in the `build_wheel.sh` script via the `run build_local all` call.

ci/release/update-version.sh
grlee77: I thought I tried that at first but for some reason it still ended in `a75` or something like that. I see that the comment up top and the variable name would indicate that should not be the case, though. Let me retry...

ci/build_wheel.sh
grlee77: sorry, merged the suggestion before the CI img update was done. I will reintroduced it for now to check if using NEXT_FULL_TAG as in a comment above is working, but can remove them again later

ci/release/update-version.sh
grlee77: updated to use NEXT_FULL_TAG

.github/workflows/build.yaml
raydouglass: Just to clarify, I don't think the `wheel-build` stage needs to depend on the `cpp-build` stage since they are independent, so this line can be removed.

grlee77: sorry, I had misread and thought this was a suggestion to add it. I have removed it now

ci/build_wheel.sh
grlee77: Okay, now there should be no apt/yum calls in the script anymore

raydouglass (approved):

ci/build_python.sh
```diff
@@ -9,13 +9,27 @@ export CMAKE_GENERATOR=Ninja
 
 rapids-print-env
 
+package_name="cucim"
+package_dir="python/cucim/src"
+
+version=$(rapids-generate-version)
+# for CMake VERSION need to truncate any trailing 'a'
+version_cpp=${version%a*}
```
vyasr, line 17: suggestion (non-blocking): The alternative would be maintaining a single file and then stripping out the alpha in CMake. Now that it's done this way I don't know if it's worth changing, but it would let you carry around one less file.

ci/build_wheel.sh
```diff
@@ -0,0 +1,63 @@
+#!/bin/bash
+# Copyright (c) 2023, NVIDIA CORPORATION.
+
+set -euo pipefail
+
+package_name="cucim"
+package_dir="python/cucim"
+package_src_dir="${package_dir}/src/${package_name}"
+
+CMAKE_BUILD_TYPE="release"
```
vyasr, line 10: thought: I was going to ask if this was really necessary since it's CMake's default, but now I see that this is for the `build_local` script. It's out of scope for this PR, but maybe that script should support a default since CMake does?

ci/build_wheel.sh
```diff
@@ -0,0 +1,63 @@
...
+echo "${version_cpp}" > "${package_dir}/VERSION"
+sed -i "/^__git_commit__/ s/= .*/= \"${commit}\"/g" "${package_src_dir}/_version.py"
+
+if [[ ${PACKAGE_CUDA_SUFFIX} == "-cu12" ]]; then
+    # change pyproject.toml to use CUDA 12.x version of cupy
+    sed -i "s/cupy-cuda11x/cupy-cuda12x/g" ${pyproject_file}
+fi
+
+echo "The package name and/or version was modified in the package source. The git diff is:"
+git diff
```
vyasr, line 40: note: JFYI we've removed this from every other repo at this point but you're welcome to keep it if you find it useful for debugging. It's been a long time since we got much use out of it.

ci/build_wheel.sh
```diff
@@ -0,0 +1,63 @@
...
+
+pip install --upgrade pip
```
vyasr, line 42: suggestion (blocking): This shouldn't be necessary. Were you running into specific problems? If so, do we need to address them somewhere?

ci/build_wheel.sh
```diff
@@ -0,0 +1,63 @@
...
+
+# Install pip build dependencies (not yet using pyproject.toml)
+rapids-dependency-file-generator \
+  --file_key "py_build" \
+  --output "requirements" \
+  --matrix "cuda=${RAPIDS_CUDA_VERSION%.*};arch=$(arch);py=${RAPIDS_PY_VERSION}" | tee build_requirements.txt
+pip install -r build_requirements.txt
```
vyasr, line 49: suggestion (blocking): Is this still true? Now that #617 is merged you should be using pyproject.toml. That said, I don't know how the C++ build works when you do `build_local`. Are these dependencies somehow used there? That would be strange and require some gymnastics for library paths, but I suppose it's possible?

cpp/cmake/deps/openslide.cmake
```diff
@@ -22,7 +22,9 @@ if (NOT TARGET deps::openslide)
         set(OPENSLIDE_LIB_PATH "$ENV{CONDA_PREFIX}/lib/libopenslide.so")
     elseif (EXISTS /usr/lib/x86_64-linux-gnu/libopenslide.so)
         set(OPENSLIDE_LIB_PATH /usr/lib/x86_64-linux-gnu/libopenslide.so)
-    else () # CentOS 6
+    elseif (EXISTS /usr/lib/aarch64-linux-gnu/libopenslide.so)
+        set(OPENSLIDE_LIB_PATH /usr/lib/aarch64-linux-gnu/libopenslide.so)
```
vyasr, line 26: question: I can see why this change was needed in general, but it doesn't seem relevant to this PR. Was it just snuck in out of convenience?

vyasr (changes requested):
praise: Generally looks good! All the key changes are in place, there's just a little bit of cleanup to be done I think.

note: Looks like some formatting changes snuck in probably due to the change in the pre-commit config, but they're small and easily identified so not a big deal.

ci/build_wheel.sh
grlee77: No, I was originally also installing a newer CMake etc here and was updating pip before doing that. I can remove this line.

ci/build_wheel.sh
grlee77: Okay, I can remove it

ci/build_wheel.sh
grlee77: Yes, I think it still needs to be this way as from the `pyproject.toml` perspective this is a pure Python package. The pyproject build relies on an external build step (`build_local` script in this file) having already built and copied the corresponding C++ .so files into the `python/cucim/src/cucim/clara` folder.

I agree that this is a bit non-standard and we can potentially look into moving to a scikit-build based approach in the future. I don't have time to focus on that at the moment, though.

One benefit of the approach as it is now is that even Windows users can build the pure Python package using pip without having to have CMake and compile the C++ library (which supports linux only). They would be able to use all of `cucim.skimage` and part of `cucim.core`, but just would not have the I/O functions in `cucim.clara` available.

dependencies.yaml
grlee77: removed duplicate definitions. Install of `imagecodecs` and `openslide-python` is currently disabled at wheel test time, but we intend to fix that after this PR is merged (opened issue #626 to track)

ci/build_python.sh
grlee77: Agreed. I did it this way because I am not familiar with the proper CMake syntax to do it. Happy to update it if you have suggestions on how the CMake code would look.

ci/build_wheel.sh
grlee77: There is a default, but it is "debug" rather than release! Could potentially change it if there is no objection from @gigony 
```bash
    local subcommand="${1:-all}"
    local build_type="${2:-debug}"
```

cpp/cmake/deps/openslide.cmake
grlee77: It is needed here or the arm64 wheel builds fail to find libopenslide and fail during the `run build_local` call. The use of `libopenslide` is only needed for C++ lib benchmarks and test cases and it isn't directly used by the Python wheel, but it has to be found to run the C++ build stage (There is not currently a CMake flag to disable building the benchmark binary or tests).

ci/build_python.sh
vyasr: This would probably do it (after the existing `file(STRING...)` command):
```
string(REGEX REPLACE "a.*$" "" VERSION ${VERSION})                                                                                                                                                        
```

ci/build_wheel.sh
vyasr: I get that there's a separate `build_local` run, what I'm confused about is what dependencies are being installed by pip that are used in that step? Is it just for CMake/ninja? If so, are these really "py_build" requirements?

dependencies.yaml
grlee77: opened issue https://github.com/rapidsai/cucim/issues/626 to track

ci/build_python.sh
grlee77: Thanks, @vyasr, that seems to work! I pushed a commit to remove VERSION_CPP as well as simplify the CI scripts a bit (there was still writing of some unused VERSION file).

There should now be only a single top-level VERSION file (as well as  symlink to it from `python/cucim/src/cucim/VERSION`)

jakirkham:
In discussion with Vyas offline, we've agreed to go ahead and merge. We can follow up on anything else in subsequent PRs/issues

jakirkham (approved):

jakirkham:
Thanks Greg for the PR and everyone for the reviews! 🙏

ci/build_wheel.sh
jakirkham: Here's what I see [on CI]( https://github.com/rapidsai/cucim/actions/runs/6777581522/job/18423219369#step:9:21 ):

```requirements.txt
# To make changes, edit dependencies.yaml and run `rapids-dependency-file-generator`.
cmake>=3.23.1,!=3.25.0
ninja
setuptools>=24.2.0
wheel
```

jakirkham: Trying deleting these lines in PR: https://github.com/rapidsai/cucim/pull/628

ci/build_wheel.sh
jakirkham: Raised as issue: https://github.com/rapidsai/cucim/issues/629

PR #621, opened 2023-10-28
Standardize on `rng` over `seed` and fix miscellaneous deprecation warnings

grlee77:
This PR fixes the following warnings observed when running the test suite:

```
src/cucim/core/operations/morphology/tests/test_distance_transform.py: 90 warnings
  /pyenv/versions/3.9.18/lib/python3.9/site-packages/cucim/core/operations/morphology/_distance_transform.py:160: DeprecationWarning: Conversion of an array with ndim > 0 to a scalar is deprecated, and will error in future. Ensure you extract a single element from your array before performing this operation. (Deprecated NumPy 1.25.)
    scalar_sampling = float(unique_sampling)
```

```
 src/cucim/skimage/feature/tests/test_corner.py::test_custom_eigvals_kernels_vs_linalg_eigvalsh[float64-shape2]
  /__w/cucim/cucim/python/cucim/src/cucim/skimage/feature/tests/test_corner.py:345: FutureWarning: use_gaussian_derivatives currently defaults to False, but will change to True in a future version. Please specify this argument explicitly to maintain the current behavior
    H = hessian_matrix(img)
```

```
 src/cucim/skimage/morphology/tests/test_binary.py: 148 warnings
  /__w/cucim/cucim/python/cucim/src/cucim/skimage/morphology/tests/test_binary.py:68: FutureWarning: `seed` is a deprecated argument name for `binary_blobs`. It will be removed in version 0.23. Please use `rng` instead.
    img = cp.asarray(data.binary_blobs(32, n_dim=ndim, seed=1))
```

### DIFF
--- a/python/cucim/src/cucim/core/operations/morphology/_distance_transform.py
+++ b/python/cucim/src/cucim/core/operations/morphology/_distance_transform.py
@@ -157,7 +157,7 @@ def distance_transform_edt(
         if len(unique_sampling) == 1:
             # In the isotropic case, can use the kernels without sample scaling
             # and just adjust the final distance accordingly.
-            scalar_sampling = float(unique_sampling)
+            scalar_sampling = float(unique_sampling[0])
             sampling = None
 
     if image.ndim == 3:
--- a/python/cucim/src/cucim/core/operations/morphology/_distance_transform.py
+++ b/python/cucim/src/cucim/core/operations/morphology/_distance_transform.py
@@ -1,4 +1,4 @@
-import numpy as np
+import cupy as cp
 
 from ._pba_2d import _pba_2d
 from ._pba_3d import _pba_3d
@@ -153,11 +153,11 @@ def distance_transform_edt(
     """
     scalar_sampling = None
     if sampling is not None:
-        unique_sampling = np.unique(np.atleast_1d(sampling))
+        unique_sampling = cp.unique(cp.atleast_1d(sampling))
         if len(unique_sampling) == 1:
             # In the isotropic case, can use the kernels without sample scaling
             # and just adjust the final distance accordingly.
-            scalar_sampling = float(unique_sampling[0])
+            scalar_sampling = float(unique_sampling)
             sampling = None
 
     if image.ndim == 3:
--- a/python/cucim/src/cucim/skimage/_shared/utils.py
+++ b/python/cucim/src/cucim/skimage/_shared/utils.py
@@ -638,8 +638,7 @@ def check_random_state(seed):
     if isinstance(seed, cp.random.RandomState):
         return seed
     raise ValueError(
-        "%r cannot be used to seed a numpy.random.RandomState"
-        " instance" % seed
+        "%r cannot be used to seed a cupy.random.RandomState" " instance" % seed
     )
 
 
--- a/python/cucim/src/cucim/skimage/_shared/utils.py
+++ b/python/cucim/src/cucim/skimage/_shared/utils.py
@@ -638,7 +638,7 @@ def check_random_state(seed):
     if isinstance(seed, cp.random.RandomState):
         return seed
     raise ValueError(
-        "%r cannot be used to seed a cupy.random.RandomState" " instance" % seed
+        f"{seed} cannot be used to seed a cupy.random.RandomState instance"
     )
 
 
--- a/python/cucim/src/cucim/skimage/data/_binary_blobs.py
+++ b/python/cucim/src/cucim/skimage/data/_binary_blobs.py
@@ -1,8 +1,14 @@
 import cupy as cp
 
+from .._shared.filters import gaussian
+from .._shared.utils import deprecate_kwarg
 
+
+@deprecate_kwarg(
+    {"seed": "rng"}, deprecated_version="23.12.00", removed_version="24.12.00"
+)
 def binary_blobs(
-    length=512, blob_size_fraction=0.1, n_dim=2, volume_fraction=0.5, seed=None
+    length=512, blob_size_fraction=0.1, n_dim=2, volume_fraction=0.5, rng=None
 ):
     """
     Generate synthetic binary image with several rounded blob-like objects.
@@ -19,12 +25,10 @@ def binary_blobs(
     volume_fraction : float, default 0.5
         Fraction of image pixels covered by the blobs (where the output is 1).
         Should be in [0, 1].
-    seed : {None, int, `cupy.random.Generator`}, optional
-        If `seed` is None the `cupy.random.Generator` singleton is used.
-        If `seed` is an int, a new ``Generator`` instance is used,
-        seeded with `seed`.
-        If `seed` is already a ``Generator`` instance then that instance is
-        used.
+    rng : {`cupy.random.Generator`, int}, optional
+        Pseudo-random number generator.
+        By default, a PCG64 generator is used (see :func:`cupy.random.default_rng`).
+        If `rng` is an int, it is used to seed the generator.
 
     Returns
     -------
@@ -34,7 +38,7 @@ def binary_blobs(
     Notes
     -----
     Warning: CuPy does not give identical randomly generated numbers as NumPy,
-    so using a specific seed here will not give an identical pattern to the
+    so using a specific `rng` here will not give an identical pattern to the
     scikit-image implementation.
 
     The behavior for a given random seed may also change across CuPy major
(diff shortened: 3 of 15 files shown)


gigony (approved):
Thanks @grlee77 !
Looks good to me!

python/cucim/src/cucim/core/operations/morphology/_distance_transform.py
```diff
@@ -157,7 +157,7 @@ def distance_transform_edt(
         if len(unique_sampling) == 1:
             # In the isotropic case, can use the kernels without sample scaling
             # and just adjust the final distance accordingly.
-            scalar_sampling = float(unique_sampling)
+            scalar_sampling = float(unique_sampling[0])
```
jakirkham, line 160: If this is a scalar, should we use `.item()`?

```suggestion
            scalar_sampling = float(unique_sampling.item())
```

python/cucim/src/cucim/skimage/feature/tests/test_corner.py
```diff
@@ -342,7 +342,7 @@ def _reference_eigvals_computation(S_elems):
 def test_custom_eigvals_kernels_vs_linalg_eigvalsh(shape, dtype):
     rng = cp.random.default_rng(seed=5)
     img = rng.integers(0, 256, shape)
-    H = hessian_matrix(img)
+    H = hessian_matrix(img, use_gaussian_derivatives=False)
```
jakirkham, line 345: Will we need to change [the `hessian_matrix`'s use `use_gaussian_derivatives` default value]( https://github.com/rapidsai/cucim/blob/6d5b8741ee26e04716653f22602682010f7058df/python/cucim/src/cucim/skimage/feature/corner.py#L252 ) at some point?

python/cucim/src/cucim/skimage/morphology/tests/test_binary.py
```diff
@@ -65,7 +65,7 @@ def _get_decomp_test_data(function, ndim=2):
         img = cp.zeros((17,) * ndim, dtype=cp.uint8)
         img[(8,) * ndim] = 1
     else:
-        img = cp.asarray(data.binary_blobs(32, n_dim=ndim, seed=1))
+        img = cp.asarray(data.binary_blobs(32, n_dim=ndim, rng=1))
```
jakirkham, line 68: It looks like [our `binary_blobs` doesn't take `rng` currently]( https://github.com/rapidsai/cucim/blob/5e25641bce2e52595eae160802d02ebd3e6de911/python/cucim/src/cucim/skimage/data/_binary_blobs.py#L4-L6 ). Should we add it?

jakirkham (commented):
Thanks Greg! 🙏

Had a couple questions below

python/cucim/src/cucim/core/operations/morphology/_distance_transform.py
grlee77: It is a 1d array of length 1, not an array scalar. I thought `item()` was only for array scalars, but I checked just now and it also works for arrays of size 1.

grlee77: just `float(unique_sampling)` will work for size 1 arrays.

python/cucim/src/cucim/skimage/feature/tests/test_corner.py
grlee77: Not sure. It is not marked for change yet in upstream `scikit-image`.

python/cucim/src/cucim/core/operations/morphology/_distance_transform.py
grlee77: see [45fa113](https://github.com/rapidsai/cucim/pull/621/commits/45fa113c5689c1952bb5df34e66fd7a7d2bc9693)

python/cucim/src/cucim/skimage/morphology/tests/test_binary.py
grlee77: see [4f16f7a](https://github.com/rapidsai/cucim/pull/621/commits/4f16f7a6b4ce544e686fc93127165d77a1849a11) for the additional changes

python/cucim/src/cucim/skimage/data/_binary_blobs.py
```diff
@@ -1,8 +1,14 @@
...
+from .._shared.utils import deprecate_kwarg
 
+
+@deprecate_kwarg(
+    {"seed": "rng"}, deprecated_version="23.12.00", removed_version="24.12.00"
+)
 def binary_blobs(
-    length=512, blob_size_fraction=0.1, n_dim=2, volume_fraction=0.5, seed=None
+    length=512, blob_size_fraction=0.1, n_dim=2, volume_fraction=0.5, rng=None
 ):
```
jakirkham, line 12: Does this just forward the `seed` argument to `rng` and raise a depreciation warning?

python/cucim/src/cucim/core/operations/morphology/_distance_transform.py
```diff
@@ -153,7 +153,7 @@ def distance_transform_edt(
     """
     scalar_sampling = None
     if sampling is not None:
-        unique_sampling = np.unique(np.atleast_1d(sampling))
+        unique_sampling = cp.unique(cp.atleast_1d(sampling))
```
jakirkham, line 156: Glad to see we can use CuPy here. Was there some kind of change that made this possible?

python/cucim/src/cucim/skimage/_shared/utils.py
```diff
@@ -638,8 +638,7 @@ def check_random_state(seed):
     if isinstance(seed, cp.random.RandomState):
         return seed
     raise ValueError(
-        "%r cannot be used to seed a numpy.random.RandomState"
-        " instance" % seed
+        "%r cannot be used to seed a cupy.random.RandomState" " instance" % seed
```
jakirkham, line 641: Think we can combine the 2 strings. Also how about using f-strings?

```suggestion
        f"{seed} cannot be used to seed a cupy.random.RandomState instance"
```

python/cucim/src/cucim/skimage/data/_binary_blobs.py
```diff
@@ -19,12 +25,10 @@ def binary_blobs(
...
-    seed : {None, int, `cupy.random.Generator`}, optional
-        If `seed` is None the `cupy.random.Generator` singleton is used.
-        If `seed` is an int, a new ``Generator`` instance is used,
-        seeded with `seed`.
-        If `seed` is already a ``Generator`` instance then that instance is
-        used.
+    rng : {`cupy.random.Generator`, int}, optional
+        Pseudo-random number generator.
+        By default, a PCG64 generator is used (see :func:`cupy.random.default_rng`).
+        If `rng` is an int, it is used to seed the generator.
```
jakirkham, line 31: Do we still want to mention `seed` and just say it is deprecated (and to use `rng` instead)?

Note this comes up below as well. So any change here would be worth propagating

python/cucim/src/cucim/skimage/data/_binary_blobs.py
```diff
@@ -45,19 +49,16 @@ def binary_blobs(
     --------
     >>> from cucim.skimage import data
     >>> # tiny size (5, 5)
-    >>> blobs = data.binary_blobs(length=5, blob_size_fraction=0.2, seed=1)
+    >>> blobs = data.binary_blobs(length=5, blob_size_fraction=0.2)
```
jakirkham, line 52: Should we use `rng` here?

```suggestion
    >>> blobs = data.binary_blobs(length=5, blob_size_fraction=0.2, rng=1)
```

python/cucim/src/cucim/skimage/feature/tests/test_corner.py
jakirkham: Maybe we can write up an issue to track

jakirkham (commented):
Thanks Greg! 🙏

Had a few comments below

python/cucim/src/cucim/skimage/data/_binary_blobs.py
grlee77: Yes, it forwards and raises a `FutureWarning` (the decorator is the same one used by scikit-image itself).

I think scikit-image changed seed->rng before releasing 0.21, though so they wouldn't have the dual `deprecate_kwarg` decorators that are unfortunately present a couple places here.

python/cucim/src/cucim/core/operations/morphology/_distance_transform.py
grlee77: It was always using CuPy, it was just getting called indirectly before via [NEP-18](https://numpy.org/neps/nep-0018-array-function-protocol.html#nep-18-a-dispatch-mechanism-for-numpy-s-high-level-array-functions)! It was just an oversight not to do `np`->`cp` previously in this file.

e.g.
```Python
import cupy as cp
import numpy as np

x = cp.ones(256)
y = np.unique(x)
type(y)
```
returns `cupy.ndarray`

python/cucim/src/cucim/skimage/_shared/utils.py
grlee77: yeah, that is definitely better. not sure how this got formatted that way...

python/cucim/src/cucim/core/operations/morphology/_distance_transform.py
jakirkham: Ah ok. Well if NEP 18 makes it easier to reuse code, that seems like a plus

python/cucim/src/cucim/skimage/data/_binary_blobs.py
grlee77: it isn't necessary as we aren't doing any test on specific output values

python/cucim/src/cucim/skimage/_shared/utils.py
jakirkham: Probably `black` did something weird

python/cucim/src/cucim/core/operations/morphology/_distance_transform.py
jakirkham: That's even better

python/cucim/src/cucim/skimage/data/_binary_blobs.py
grlee77: Actually, the decorator also takes care of that by automatically appending Other Arguments with the deprecated parameter names...

it should look something like this
https://scikit-image.org/docs/stable/api/skimage.data.html#skimage.data.binary_blobs

jakirkham (approved):
Thanks Greg! 🙏

PR #620, opened 2023-10-27
Fix iterator-related memory issues

gigony:
MERGE after https://github.com/rapidsai/cucim/pull/618

### Fix memory corruption issue in ThreadBatchDataLoader

Fix #592    

- Return a shared pointer of CuImageIterator, taking ownership of CuImageIterator instance, instead of returning a copy of CuImageIterator
- Do not create new instance for CuImageIterator when `__iter__` is called. Instead, return self.
- Wait until all threads are finished in ThreadBatchDataLoader destructor

### Fix memory leak in ThreadBatchDataLoader

Fix #598
    
- Throw StopIteration exception early
- Do not set nullptr for `raster_data_` in `ThreadBatchDataLoader::next_data()`
  - Setting nullptr prevents freeing memory in the destructor of ThreadBatchDataLoader

### Fix fatal errors for cache object

- GIL is not acquired when using Python object.

PR #623, opened 2023-10-31
Improve memory leak check stability in test_read_region_cpu_memleak test

gigony:
There is a case where memory usage check is not stable enough:
https://github.com/rapidsai/cucim/pull/620#issuecomment-1786179731

Instead of choosing index 1 and 4 (among 5 samples), choose 5 and 9 (among 10 samples) for memory leak check.


grlee77 (approved):
LGTM, thanks, Gigon

PR #628, opened 2023-11-10
Try dropping `py_build` install step

jakirkham:
Follow up on this discussion: https://github.com/rapidsai/cucim/pull/619/files#r1383859352

cc @vyasr @grlee77

### DIFF
--- a/ci/build_wheel.sh
+++ b/ci/build_wheel.sh
@@ -31,13 +31,6 @@ if [[ ${PACKAGE_CUDA_SUFFIX} == "-cu12" ]]; then
     sed -i "s/cupy-cuda11x/cupy-cuda12x/g" ${pyproject_file}
 fi
 
-# Install pip build dependencies (not yet using pyproject.toml)
-rapids-dependency-file-generator \
-  --file_key "py_build" \
-  --output "requirements" \
-  --matrix "cuda=${RAPIDS_CUDA_VERSION%.*};arch=$(arch);py=${RAPIDS_PY_VERSION}" | tee build_requirements.txt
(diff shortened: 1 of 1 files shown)


grlee77:
I think this will not work as we need the CMake update (ninja may not strictly be necessary) from here
https://github.com/rapidsai/cucim/blob/c0ef41a0c3980aa5e7908671ab5074abc00e0ade/dependencies.yaml#L84-L87

in order to run the `./run build_local` commands for the wheel build. Those preceed building the python package with pip wheel, so the requirements would not have been installed yet from the `pyproject.toml` file.

jakirkham:
Am a little confused. Does that mean CMake is installed somewhere and just needs to be updated to a new version (maybe in RAPIDS Docker images instead)? Or does that mean something else?

bdice:
These lines cannot be removed. We must install these prior to running the package build. In most other RAPIDS packages, this is done by putting them in the `[build-system.requires]` section: https://github.com/rapidsai/cudf/blob/8deb3dd7573000e7d87f18a9e2bbe39cf2932e10/python/cudf/pyproject.toml#L3-L16

As I understand it, cuCIM doesn't build in a way that respects that field of `pyproject.toml` due to its unusual "run.sh" script, so we must install the build system requirements manually.

jakirkham:
After discussion offline, it sounds like we don't want to handle this in the Docker images. Most projects solve this by using scikit-build where CMake is listed as a dependency

Given cuCIM hasn't adopted scikit-build yet (tbd if/when that would happen), this workaround has been deemed acceptable for the time being

PR #631, opened 2023-11-13
carry out removals of previously deprecated functions/kwargs

grlee77:
Carry out a scheduled removal of the `inverse` and `compute_hessian_eigenvalues` functions. These were not part of the top level submodule APIs, but not technically private previously so were on a deprecation cycle before their eventual removal.

A couple other minor fixes were made:
- deprecated/removed versions were wrong in `unsupervised_wiener`
- updated a docstring for Gaussian which indicated a kwarg default change would have already taken place. I think we can just make the actual change in the next release rather than this one. (It was not yet changed in the latest scikit-image release, but has been changed on scikit-image main)


gigony (approved):

Issue #626, opened 2023-11-06
[BUG] fix isolated test failure in `test_cache_hit_miss`


grlee77:
**Describe the bug**

When working on wheel builds in !619, it was observed that a single test case, `test_cache_hit_miss` would fail when running the wheel tests. Temporarily, this test case is not being run because `imagecodecs` is not installed as part of the test dependencies during wheel testing.  Before release 23.12.00 we should enable install of `imagecodecs` and `openslide-python` again so that all `cucim.clara` submodule test cases get run.

grlee77:
#634 re-enables these additional test dependencies and can be used as a basis to start from when fixing this issue

jakirkham:
Moving to 24.02 as we are already in code freeze for 23.12

PR #634, opened 2023-11-17
install imagecodecs and openslide-python dependencies so additional tests will run

grlee77:
This MR may initially fail during wheel testing due to #626, as a few test cases relying on these packages were not being run at the time #619 was merged.


ci/test_wheel.sh
```diff
@@ -13,6 +13,21 @@ python -m pip install $(echo ./dist/cucim*.whl)[test]
 if [[ "$(arch)" == "aarch64" && ${RAPIDS_BUILD_TYPE} == "pull-request" ]]; then
     python ./ci/wheel_smoke_test.py
 else
+
+    DEBIAN_FRONTEND=noninteractive apt install -y --no-install-recommends libopenslide-dev
```
jakirkham, line 17: Do we need the dev package at test time or just the runtime library?

```suggestion
    DEBIAN_FRONTEND=noninteractive apt install -y --no-install-recommends libopenslide
```

grlee77: Thanks, we shouldn't need the dev lib. Apprently the package is called `libopenslide0`, though

grlee77, line 17:
```suggestion
    DEBIAN_FRONTEND=noninteractive apt install -y --no-install-recommends libopenslide0
```

grlee77:
Okay, libopenslide is now being found and the `test_converter` case passes. There is a single test failure corresponding to #626 and for some reason also a segfault upon pytest exit. This is the end of the log

```
FAILED python/cucim/tests/unit/clara/test_image_cache.py::test_cache_hit_miss - assert (0, 0) == (0, 1)
  At index 1 diff: 0 != 1
  Full diff:
  - (0, 1)
  ?     ^
  + (0, 0)
  ?     ^
==== 1 failed, 9665 passed, 42 skipped, 127 warnings in 2718.67s (0:45:18) =====
pure virtual method called
terminate called without an active exception
ci/test_wheel.sh: line 34:  1078 Aborted                 (core dumped) python -m pytest ./python/cucim
Error: Process completed with exit code 134.
```

jakirkham:
^ @gigony

ci/test_wheel.sh
```diff
@@ -13,6 +13,12 @@ python -m pip install $(echo ./dist/cucim*.whl)[test]
...
-    # TODO: revisit enabling imagecodecs package during testing
-    python -m pytest ./python/cucim
+
+    DEBIAN_FRONTEND=noninteractive apt update
+    DEBIAN_FRONTEND=noninteractive apt install -y --no-install-recommends libopenslide0
+
+    # TODO: fix and remove two test cases skipped here
+    #   test_cache_hit_miss currently fails
+    #   test_converter passes, but a segfault occurs on pytest exit when it is enabled
+    python -m pytest ./python/cucim -k "not test_converter and not test_cache_hit_miss"
```
jakirkham, line 23: Would it be possible to move these to PyTest decorators?

We could also use `skipif` to more carefully restrict when they are skipped

```Python
@pytest.mark.skip(reason="currently fails")
def test_cache_hit_miss():
    ...
 
@pytest.mark.skip(reason="passes, but a segfault occurs on pytest exit when it is enabled")
def test_converter():
    ...
```

jakirkham (commented):
Thanks Greg! 🙏

Had a follow up question below (assuming this change works)

jakirkham:
After a bit of effort Gigon was able to reproduce this when we met yesterday using a Ubuntu 18 Docker container and installing the wheel. With a bit of debugging it appears this goes back to an old cuFile issue that we thought had been fixed with PR ( https://github.com/rapidsai/cucim/pull/158 ). However it appears this issue is back

jakirkham:
After discussion with Greg offline, we decided to move this to 24.02 as we are already in code freeze for 23.12

jakirkham:
Style check failure is unrelated. It is coming from the version update script. Attempting to address in PR ( https://github.com/rapidsai/cucim/pull/648 )

raydouglass (approved):

gigony (approved):

jakirkham (approved):

jakirkham:
Thanks all! 🙏

Issue #641, opened 2023-11-29
23.12 nightly wheels missing Python minor version


jakirkham:
Currently the cuCIM 23.12 nightly wheels are showing up as if they are Python 3 universal wheels. Please see the screenshots below

<details>
<summary><code>cucim-cu12</code>:</summary>

<img width="1840" alt="Screenshot 2023-11-28 at 5 50 54 PM" src="https://github.com/rapidsai/cucim/assets/3019665/c20b2edf-e797-4f9a-9e7e-03979706108a">

</details>

<details>
<summary><code>cucim-cu11</code>:</summary>

<img width="1840" alt="Screenshot 2023-11-28 at 5 50 40 PM" src="https://github.com/rapidsai/cucim/assets/3019665/72b23731-a72a-430e-b653-bef5738d4c9f">

</details>

However they have strict Python version dependence, which is not being reflected

Basically this is a consequence of the wheels being treated as pure when they are not

Would be good to correct this somehow

jakirkham:
This [SO answer]( https://stackoverflow.com/a/75240449 ) gave a nice summary of potential options

PR #643, opened 2023-11-29
Inform `setuptools` that `Extensions` are included

jakirkham:
Fixes https://github.com/rapidsai/cucim/issues/641

As `setuptools`/`pip`/`wheel` will mark our package as pure (even though we put an extension module into it), make sure it doesn't by informing `setuptools` that `Extensions` are contained within. This should ensure that are wheels include the Python minor version they were built for.


jakirkham:
This works! 🎉 

Final name [from CI]( https://github.com/rapidsai/cucim/actions/runs/7027516742/job/19122056172?pr=643#step:9:3607 ):

```
./cucim_cu11-23.12.0a21-cp39-cp39-manylinux_2_17_x86_64.manylinux2014_x86_64.whl
```

grlee77 (approved):
Thanks, John. This looks good! I guess this may be preferable to renaming after the fact as in #644.

jakirkham:
Based on offline discussion, we decided to go with PR ( https://github.com/rapidsai/cucim/pull/644 ) as it is a publicly documented API that is most consistent with how the build works

Longer term we might want to use `scikit-build` ( https://github.com/rapidsai/cucim/issues/645 ), which would make this unnecessary

raydouglass (approved):

bdice (approved):

vyasr (approved):

vyasr (changes requested):
To unblock the release, can we also add (or symlink) the license into python/cucim?

jakirkham:
Good catch. Went ahead and symlinked both and updated `pyproject.toml` to list both

raydouglass (approved):

PR #644, opened 2023-11-29
Retag wheels to be `cpXY`

jakirkham:
Fixes https://github.com/rapidsai/cucim/issues/641

After building the wheels, this...

* Creates the tag string from `cp` (CPython) and the `python`'s version
* Uses the tag string to replace the `python_tag` and `abi_tag`
* Removes the previous wheels while performing this conversion
* Hands them to `auditwheel` for final steps

### DIFF
--- a/ci/build_wheel.sh
+++ b/ci/build_wheel.sh
@@ -45,6 +45,10 @@ cd "${package_dir}"
 
 python -m pip wheel . -w dist -vvv --no-deps --disable-pip-version-check
 
+# Fix Python wheel tag to be restricted to CPython and version dependent
+WHEEL_PYTHON_TAG=`python -c 'import sys; print("cp{0}{1}".format(*sys.version_info[:2]))'`
+python -m wheel tags --remove --python-tag="${WHEEL_PYTHON_TAG}" --abi-tag="${WHEEL_PYTHON_TAG}" dist/*
+
 mkdir -p final_dist
 python -m auditwheel repair -w final_dist dist/*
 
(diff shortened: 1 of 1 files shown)


ci/build_wheel.sh
```diff
@@ -45,6 +45,10 @@ cd "${package_dir}"
 
 python -m pip wheel . -w dist -vvv --no-deps --disable-pip-version-check
 
+# Fix Python wheel tag to be restricted to CPython and version dependent
+WHEEL_PYTHON_TAG=`python -c 'import sys; print("cp{0}{1}".format(*sys.version_info[:2]))'`
```
grlee77, line 49: This is an alternate way to achieve the same thing, but I am fine with it as-is
```suggestion
WHEEL_PYTHON_TAG=cp$(echo ${RAPIDS_PY_VERSION} | sed 's/\.//g')
```

grlee77 (approved):
Thanks, John. It looks good as is.

I left a comment with another possible way of setting WHEEL_PYTHON_TAG without launching Python.

ci/build_wheel.sh
```diff
@@ -45,6 +45,10 @@ cd "${package_dir}"
 
 python -m pip wheel . -w dist -vvv --no-deps --disable-pip-version-check
 
+# Fix Python wheel tag to be restricted to CPython and version dependent
```
grlee77, line 48: For future reference, we could be more explicit about why this change is needed.

```suggestion
# Because we built the library with the run script and manually copied the shared pybind11
# library, the subsequent py_project.toml wheel build was as a pure Python package and results in
# tags that incorrectly indicate it is a universal wheel. To fix this, we need to modify the wheel
# to have CPython ABI and Python tags matching the version of Python that the Python bindings were
# built with.
```

ci/build_wheel.sh
jakirkham: No strong feelings between the other approaches. Will leave for now and see what others think

vyasr (approved):

ci/build_wheel.sh
bdice: I'm in favor of using `RAPIDS_PY_VERSION` (accepting this suggestion).

raydouglass (approved):

jakirkham:
Thanks all! 🙏

PR #648, opened 2023-12-02
Fix style issue in `docs/source/conf.py`

jakirkham:
Currently the version update script is introducing a style error (using single quotes instead of double quotes) in the versions in `docs/source/conf.py`. This fixes that issue


grlee77 (approved):
Thanks, LGTM

jakirkham:
That Greg! 🙏

Also tested the `sed` command lines to confirm it was doing the right thing

Will check with OPS on reviewing next week

jakirkham:
Thanks all! 🙏

PR #650, opened 2023-12-06
pin to OpenSlide <4 during the conda build of libcucim

grlee77:
A conda package for openslide 4.0 was released yesterday, but it appears to not automatically pull in a required libdicom dependency. Pin to openslide<4 for now to avoid this issue and because we have not tested the library with 4.0.


gigony (approved):
Thanks for the triage and for fixing the issue!

PR #653, opened 2023-12-08
Relax `openslide` pin

jakirkham:
This was pinned to avoid `openslide` version `4` ( https://github.com/rapidsai/cucim/pull/650 ) due to some solver issues that cropped up with it. Namely it was missing `libdicom` (and an associated version constraint). However this has been fixed upstream by adding...

* A repodata patch ( https://github.com/conda-forge/conda-forge-repodata-patches-feedstock/pull/613 )
* A `run_exports` to `libdicom` (was missing) ( https://github.com/conda-forge/libdicom-feedstock/pull/5 )
* A `libdicom` associated test in `openslide` ( https://github.com/conda-forge/openslide-feedstock/pull/25 )

Given this we should be able to relax the pin here, which this does.


jakirkham:
Seeing this error on CI:

```
ERROR tests/unit/clara/converter/test_converter.py - OSError: libopenslide.so.0: cannot open shared object file: No such file or directory
```

The SOVERSION of `libopenslide` did go to `1`

<hr>

That said the usage of `libopenslide` in cuCIM does not appear to pin the SOVERSION:

https://github.com/rapidsai/cucim/blob/8951d792c1702b28a81afacfbbc92d499c9909b6/cpp/cmake/deps/openslide.cmake#L19-L29

https://github.com/rapidsai/cucim/blob/8951d792c1702b28a81afacfbbc92d499c9909b6/cpp/plugins/cucim.kit.cuslide/cmake/deps/openslide.cmake#L19-L29

<hr>

Given this, where does the expectation to have `libopenslide.so.0` come from?

jakirkham:
Looks like it comes from `openslide-python`. It appears we need a version of `openslide-python` with this fix ( https://github.com/openslide/openslide-python/pull/199 ), which means `1.3.0` or greater

ajschmidt8 (approved):

gigony (approved):
Thanks @jakirkham for handling openslide dependencies! Looks good to me!

jakirkham:
Thanks all! 🙏

PR #654, opened 2023-12-09
Add 3rd party license file in Conda packages

jakirkham:
Previously only cuCIM's own license file was included in Conda packages. However the 3rd party license file was not included in Conda packages. This adds that license file as well. Also goes ahead and adds the 3rd party license file to both `libcucim` and `cucim` for good measure (even though it is a little redundant to add it to `cucim` given it depends on `libcucim`, which already has it)


raydouglass (approved):

grlee77 (approved):
I guess we can potentially remove `dockcross` now that we are using RAPIDS containers for the wheel builds. We still have those files in the repository, though, so let's leave that for a separate follow-up.

jakirkham:
> I guess we can potentially remove `dockcross` now that we are using RAPIDS containers for the wheel builds. We still have those files in the repository, though, so let's leave that for a separate follow-up.

Yep filed issue ( https://github.com/rapidsai/cucim/issues/655 ) recently to discuss

PR #668, opened 2023-12-13
remove redundant notebook

### LINKED ISSUE #656: Cache Notebooks
It looks like we have 2 notebooks related to caching

One is in the top-level of the repo

https://github.com/rapidsai/cucim/blob/8951d792c1702b28a81afacfbbc92d499c9909b6/Cache%20Example.ipynb

The other is in the `notebooks` directory with the other notebooks

https://github.com/rapidsai/cucim/blob/8951d792c1702b28a81afacfbbc92d499c9909b6/notebooks/Using_Cache.ipynb

Do these serve different purposes? If so, should we consolidate into the `notebooks` directory

If not, should we drop the top-level notebook?

grlee77:
closes #656

We are retaining `notebooks/Using_Cache.ipynb` which covers use of the image cache. This top-level notebook is not needed.


jakirkham (approved):
Thanks Greg! 🙏

gigony (approved):

PR #670, opened 2024-01-06
minor updates/fixes for consistency with scikit-image 0.22

grlee77:
This MR does not add any new features. It just has minor fixes ported from scikit-image 0.22. Most changed lines are from moving additional modules to use `__init__.pyi` files as for upstream scikit-image.

Regarding introduction of `py.typed`, see [discussion on the scikit-image repo](https://github.com/scikit-image/scikit-image/pull/7073). The TLDR version is that it enables VS CODE (and potentially other IDEs) to use the `.pyi` files for intellisense capabilities. Those using mypy with strict type checking, may have to add an exclusion like the following to disable warnings about most of the cuCIM API which does not currently use type hints:

```toml
[tool.mypy]
strict = true
untyped_calls_exclude = [
    "cucim"
]
```


gigony (approved):
Thank you, @grlee77, for this update. It's helpful to learn about the method for making the document accessible in the IDE, as shown in [this pull request](https://github.com/scikit-image/scikit-image/pull/7073).
Looks good to me!

jakirkham:
It looks like the style checks are failing [on CI]( https://github.com/rapidsai/cucim/actions/runs/7576366509/job/20635013804?pr=670#step:6:127 ). The single quotes need to be replaced with double quotes. There are also some long lines that need to be wrapped

Could you please run `black` @grlee77 ?

jakirkham:
It looks like we are seeing some test failures [on CI]( https://github.com/rapidsai/cucim/actions/runs/7616429632/job/20743710237?pr=670 ):

```python
FAILED src/cucim/skimage/measure/tests/test_regionprops.py::test_extra_properties_table - IndexError: list index out of range
FAILED src/cucim/skimage/filters/tests/test_thresholding.py::test_triangle_uniform_images[uint8] - ValueError: zero-size array to reduction operation CUPY_CUB_ARGMAX which has no identity
FAILED src/cucim/skimage/filters/tests/test_thresholding.py::test_triangle_uniform_images[int16] - ValueError: zero-size array to reduction operation CUPY_CUB_ARGMAX which has no identity
FAILED src/cucim/skimage/filters/tests/test_thresholding.py::test_triangle_uniform_images[float16] - ValueError: zero-size array to reduction operation CUPY_CUB_ARGMAX which has no identity
FAILED src/cucim/skimage/filters/tests/test_thresholding.py::test_triangle_uniform_images[float32] - ValueError: zero-size array to reduction operation CUPY_CUB_ARGMAX which has no identity
```

Maybe this is related to CuPy 13 (or perhaps a NumPy change)

grlee77:
The regionprops test failure was a due to a couple missing entries in PROPS and should be fixed now.

I didn't see that jitify error related to `std::is_signed` when trying locally with pip packages, but conda seems to have a problem again with that now.

gigony (approved):
Thanks @grlee77 
The updated portion of the changes looks good to me!

python/cucim/src/cucim/skimage/_shared/filters.py
```diff
@@ -133,21 +112,7 @@ def gaussian(
     >>> from skimage.data import astronaut
     >>> image = cp.array(astronaut())
     >>> filtered_img = gaussian(image, sigma=1, channel_axis=-1)
-
     """
-    if channel_axis is ChannelAxisNotSet:
```
jakirkham, line 138: Do we still need to think about the `ChannelAxisNotSet` case?

python/cucim/src/cucim/skimage/_vendored/_internal.py
```diff
@@ -3,7 +3,14 @@
 from operator import mul
 
 import cupy
-import numpy
+
+# TODO: when minimum numpy dependency is 1.25 use:
+# np..exceptions.AxisError instead of AxisError
```
jakirkham, line 8:
```suggestion
# np.exceptions.AxisError instead of AxisError
```

python/cucim/src/cucim/skimage/color/colorconv.py
```diff
@@ -63,6 +63,14 @@
 )
 from ..util import dtype, dtype_limits
 
+# TODO: when minimum numpy dependency is 1.25 use:
+# np..exceptions.AxisError instead of AxisError
```
jakirkham, line 67:
```suggestion
# np.exceptions.AxisError instead of AxisError
```

python/cucim/src/cucim/skimage/color/tests/test_colorconv.py
```diff
@@ -63,6 +63,14 @@
 )
 from cucim.skimage.util import img_as_float, img_as_float32, img_as_ubyte
 
+# TODO: when minimum numpy dependency is 1.25 use:
+# np..exceptions.AxisError instead of AxisError
```
jakirkham, line 67:
```suggestion
# np.exceptions.AxisError instead of AxisError
```

python/cucim/src/cucim/skimage/filters/ridges.py
```diff
@@ -58,7 +58,7 @@ def meijering(
     wrinkles, rivers. It can be used to calculate the fraction of the
     whole image containing such objects.
 
-    Calculates the eigenvectors of the Hessian to compute the similarity of
+    Calculates the eigenvalues of the Hessian to compute the similarity of
```
jakirkham, line 61: Was this just a typo or was there some kind of behavior change?

python/cucim/src/cucim/skimage/filters/thresholding.py
```diff
@@ -910,12 +912,25 @@ def threshold_triangle(image, nbins=256):
...
+    if hist.size == 1:
+        # integer-valued image with constant intensity will have just 1 bin
+        return image.ravel()[0]
 
+    # In most cases, nbins is small so it is better to process hist on the CPU
+    if nbins > 100000:
+        xp = cp
+    else:
+        xp = np
+        hist = cp.asnumpy(hist)
```
jakirkham, line 924: Wonder if there is something we can use from [the Array API]( https://data-apis.org/array-api/latest/ ) to make this smoother

python/cucim/src/cucim/skimage/measure/_blur_effect.py
```diff
@@ -5,6 +5,14 @@
 from ..color import rgb2gray
 from ..util import img_as_float
 
+# TODO: when minimum numpy dependency is 1.25 use:
+# np..exceptions.AxisError instead of AxisError
```
jakirkham, line 9:
```suggestion
# np.exceptions.AxisError instead of AxisError
```

python/cucim/src/cucim/skimage/restoration/deconvolution.py
```diff
@@ -307,7 +307,7 @@ def unsupervised_wiener(
     prev_x_postmean = cp.zeros(trans_fct.shape, dtype=float_type)
 
     # Difference between two successive mean
-    delta = np.NAN
+    delta = np.nan
```
jakirkham, line 310: Wonder if we should use `float("nan")`

python/cucim/src/cucim/skimage/restoration/j_invariant.py
```diff
@@ -273,7 +283,7 @@ def calibrate_denoiser(
     ...                                          calibrate_denoiser)
     >>> img = color.rgb2gray(cp.array(data.astronaut()[:50, :50]))
     >>> noisy = img + 0.5 * img.std() * cp.random.randn(*img.shape)
-    >>> parameters = {'weight': cp.arange(0.01, 0.5, 0.05)}
+    >>> parameters = {'weight': cp.arange(0.01, 0.3, 0.02)}
```
jakirkham, line 286: Am curious about what motivated this change

python/cucim/src/cucim/skimage/transform/_geometric.py
```diff
@@ -1510,7 +1517,7 @@ class SimilarityTransform(EuclideanTransform):
 
     where ``s`` is a scale factor and the homogeneous transformation matrix is::
 
-        [[a0  b0  a1]
+        [[a0  -b0  a1]
```
jakirkham, line 1520: Was this a typo or was there some kind of behavior change?

python/cucim/src/cucim/skimage/util/dtype.py
```diff
@@ -279,7 +279,7 @@ def _convert(image, dtype, force_copy=False, uniform=False):
     #   is a subclass of that type (e.g. `cp.floating` will allow
     #   `float32` and `float64` arrays through)
 
-    if cp.issubdtype(dtype_in, cp.obj2sctype(dtype)):
+    if cp.issubdtype(dtype_in, np.core.numerictypes.obj2sctype(dtype)):
```
jakirkham, line 282: Does this still handle CuPy types smoothly? Does NumPy version matter here?

jakirkham (commented):
Thanks Greg! 🙏

Had a few comments below

python/cucim/src/cucim/skimage/_shared/filters.py
grlee77: No that was a temporary class used during the deprecation of old behavior that tried to guess if it was an RGB image based on the size of the last axis. The new default is that the user has to explicitly specify if that want an axis to be treated as channels.

python/cucim/src/cucim/skimage/color/tests/test_colorconv.py
grlee77: thanks

python/cucim/src/cucim/skimage/filters/ridges.py
grlee77: just a typo fix

python/cucim/src/cucim/skimage/filters/thresholding.py
grlee77: I think this is easy enough to follow. In practice, we will almost always take the numpy code path since default bins are 256. 

100,000 seemed a reasonable cutoff for when GPU is likely to be faster on my system but I don't think we need to try to tune this function too much as I don't think it is very heavily used.

python/cucim/src/cucim/skimage/restoration/deconvolution.py
grlee77: I don't think there is a strong reason to prefer one or the other in this case. It is only used in a stopping criterion check below

```py
delta < params["threshold"]
```
which should always be False for either

python/cucim/src/cucim/skimage/restoration/j_invariant.py
grlee77: The reason I had a different value was that originally scikit-image was using a different wavelet-based denoiser. Since we didn't have `denoise_wavelet` implemented in cuCIM I switched to `denoise_tv_chambolle` in our docstring and just chose a value that worked okay.

Later, scikit-image made PyWavelets optional and changed their example to use `denoise_tv_chambolle` but happened to choose a different range than what I originally had in cuCIM. I changed to the scikit-image values for consistency, but it doesn't really matter either way.

python/cucim/src/cucim/skimage/transform/_geometric.py
grlee77: it was just a typo fix

python/cucim/src/cucim/skimage/util/dtype.py
grlee77: [Yes, because cp.obj2sctype is just an alias for the NumPy function](https://github.com/cupy/cupy/blob/5d5ef648d1b1b3af4c969e2b7a47b0d63f8b57b0/cupy/__init__.py#L363-L380). The same is true for various dtypes like `cp.float32`, etc.

python/cucim/src/cucim/skimage/filters/thresholding.py
jakirkham: Yeah was more focused on maintainability and the potential to upstream more of these changes. Though need to solve that problem here. Just something to mull on

python/cucim/src/cucim/skimage/restoration/j_invariant.py
jakirkham: Ah ok. That context was all I was looking for. Thanks Greg! 🙏

jakirkham (approved):
Thanks Greg! 🙏

jakirkham:
As pointed out by Greg offline, we need to update this line and run [`rapids-dependency-file-generator`]( https://github.com/rapidsai/dependency-file-generator#usage )

https://github.com/rapidsai/cucim/blob/bf5a6f765de64b522ac060f92e24c52f69834c56/dependencies.yaml#L226

jakirkham (changes requested):
Need to update `scikit-image` upper bound

python/cucim/requirements-test.txt
```diff
@@ -1,5 +1,6 @@
 GPUtil>=1.4.0
 imagecodecs>=2021.6.8
+numpydoc>=1.5
```
jakirkham, line 3: Also think we can fold this into `dependencies.yaml`. Trying that in PR ( https://github.com/rapidsai/cucim/pull/683 )

jakirkham: This file was folded into `pyproject.toml`

So would also suggest adding this to `dependencies.yaml`

jakirkham:
Do we need PyWavelets as a (test) dependency?

Seeing these errors [on CI]( https://github.com/rapidsai/cucim/actions/runs/7672930942/job/20914817421?pr=670#step:8:826 ):

```python
FAILED python/cucim/src/cucim/skimage/restoration/tests/test_j_invariant.py::test_denoise_invariant_color[float16] - ImportError: PyWavelets is not installed. Please ensure it is installed in order to use this function.
FAILED python/cucim/src/cucim/skimage/restoration/tests/test_j_invariant.py::test_denoise_invariant_color[float32] - ImportError: PyWavelets is not installed. Please ensure it is installed in order to use this function.
FAILED python/cucim/src/cucim/skimage/restoration/tests/test_j_invariant.py::test_denoise_invariant_color[float64] - ImportError: PyWavelets is not installed. Please ensure it is installed in order to use this function.
FAILED python/cucim/src/cucim/skimage/restoration/tests/test_j_invariant.py::test_denoise_invariant_3d - ImportError: PyWavelets is not installed. Please ensure it is installed in order to use this function.
FAILED python/cucim/src/cucim/skimage/restoration/tests/test_j_invariant.py::test_calibrate_denoiser_extra_output - ImportError: PyWavelets is not installed. Please ensure it is installed in order to use this function.
FAILED python/cucim/src/cucim/skimage/restoration/tests/test_j_invariant.py::test_calibrate_denoiser - ImportError: PyWavelets is not installed. Please ensure it is installed in order to use this function.
FAILED python/cucim/src/cucim/skimage/restoration/tests/test_j_invariant.py::test_input_image_not_modified - ImportError: PyWavelets is not installed. Please ensure it is installed in order to use this function.
```

jakirkham:
Looks like [this is used as part of `denoise_wavelet`]( https://github.com/scikit-image/scikit-image/blob/94c288c3a31a40620ac0e7fed5c07b8bff02d886/skimage/restoration/_denoise.py#L712-L718 ), which we are using in [these tests]( https://github.com/rapidsai/cucim/blob/7d4373ba270547267121d1e1ae8f8ce587fdf57f/python/cucim/src/cucim/skimage/restoration/tests/test_j_invariant.py#L6-L7 )

As pointed out by Greg offline, upstream appears to have relaxed the PyWavelets dependency in 0.22.0 with PR ( https://github.com/scikit-image/scikit-image/pull/7156 ), which is why the dependency is now missing

python/cucim/src/cucim/skimage/restoration/tests/test_j_invariant.py
```diff
@@ -2,8 +2,6 @@
 import numpy as np
 import pytest
 from skimage.data import camera, chelsea
-
-# from cucim.skimage.restoration import denoise_wavelet
 from skimage.restoration import denoise_wavelet
```
jakirkham, line 5:
```suggestion
```

python/cucim/src/cucim/skimage/restoration/tests/test_j_invariant.py
```diff
@@ -20,6 +18,9 @@
 noisy_img_color = random_noise(test_img_color, mode="gaussian", var=0.01)
 noisy_img_3d = random_noise(test_img_3d, mode="gaussian", var=0.1)
 
+# skip tests if skimage.restoration module cannot be imported
+pytest.importorskip("skimage.restoration")
+
```
jakirkham, line 23:
```suggestion
skimage_restoration = pytest.importorskip("skimage.restoration")
denoise_wavelet = skimage_restoration.denoise_wavelet

```

jakirkham (commented):
Thanks Greg! 🙏

Think we might want to drop this top-level `import` and rewrite with something like this

python/cucim/src/cucim/skimage/restoration/tests/test_j_invariant.py
grlee77: thanks for fixing that!

jakirkham (approved):

raydouglass (approved):
`ops-codeowners` changes look good

jakirkham:
Thanks Greg! 🙏

PR #671, opened 2024-01-10
refactor CUDA versions in dependencies.yaml

jameslamb:
## Description

Contributes to https://github.com/rapidsai/build-planning/issues/7.

Proposes splitting the `cuda-version` dependency in `dependencies.yaml` out to its own thing, separate from the bits of the CUDA Toolkit this project needs.

### Benefits of this change

* prevents accidental inclusion of multiple `cuda-version` version in environments
* reduces update effort (via enabling more use of globs like `"12.*"`)
* improves the chance that errors like "`conda` recipe is missing a dependency" are caught in CI

### DIFF
--- a/dependencies.yaml
+++ b/dependencies.yaml
@@ -154,9 +154,9 @@ dependencies:
             packages:
               - cuda-version=11.4
           - matrix:
-              cuda: "11.6"
+              cuda: "11.5"
             packages:
-              - cuda-version=11.6
+              - cuda-version=11.5
           - matrix:
               cuda: "11.8"
             packages:
--- a/dependencies.yaml
+++ b/dependencies.yaml
@@ -7,7 +7,8 @@ files:
       arch: [x86_64]
     includes:
       - build
-      - cudatoolkit
+      - cuda
+      - cuda_version
       - checks
       - docs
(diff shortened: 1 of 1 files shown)


dependencies.yaml
```diff
@@ -140,34 +141,41 @@ dependencies:
...
             packages:
               - cuda-version=11.4
-              - cudatoolkit
           - matrix:
-              cuda: "11.5"
+              cuda: "11.6"
             packages:
-              - cuda-version=11.5
-              - cudatoolkit
+              - cuda-version=11.6
```
jakirkham, line 159: Just checking did want to change this from 11.5 to 11.6?

jakirkham (commented):
Thanks James! 🙏

Had one question below

dependencies.yaml
jameslamb: no sorry, that was an oversight. Reverted in https://github.com/rapidsai/cucim/pull/671/commits/1502e2039a64e993b27021d8e62f9ba88a32ffea.

jakirkham (approved):
Thanks James! 🙏

AyodeAwe (approved):

PR #676, opened 2024-01-19
Fix CI issues

jakirkham:
This definition causes problems with newer CCCL versions. As we don't actually use it ourselves, go ahead and drop it to fix CI


jakirkham:
Thanks @leofang for the tip! 🙏

@grlee77 could you please review & merge (if this looks acceptable)?

jakirkham:
Thanks Greg! 🙏

PR #682, opened 2024-01-24
add rapids-dependency-file-generator pre-commmit hook

jameslamb:
Proposes adding a pre-commit hook to run `rapids-dependency-file-generator`.

With that change, local `pre-commit` and CI will raise an error if changes have been made in `dependencies.yaml` that would affect any other files checked into the repo, like the conda envs in https://github.com/rapidsai/cucim/tree/branch-24.02/conda/environments.

### How I tested this

Changed a `cmake` version constraint in `dependencies.yaml`.

Ran `pre-commit run --all-files`.

Observed the expected failure.

image

Reverted that change, ran `pre-commit run --all-files` again, saw everything pass.


jakirkham (approved):

bdice (approved):
@jameslamb Also FYI: https://github.com/rapidsai/shared-workflows/issues/46#issuecomment-1568777611

There's a list of a few repos needing this update.

bdice (changes requested):
Also we can disable the dependency check from the `checks` job now that it's covered by style checks.

Like this in `.github/workflows/pr.yaml`:
```
  checks:
    secrets: inherit
    uses: rapidsai/shared-action-workflows/.github/workflows/checks.yaml@branch-24.04
    with:
      enable_check_generated_files: false
```

Example: https://github.com/rapidsai/cudf/pull/12819/files

jakirkham:
Regarding adding `enable_check_generated_files`. Would we actually just make that `false` by default (or drop that flag altogether) once pre-commit support for rdfg is added throughout RAPIDS?

bdice:
> Regarding adding `enable_check_generated_files`. Would we actually just make that `false` by default (or drop that flag altogether) once pre-commit support for rdfg is added throughout RAPIDS?

Yes. After we add pre-commit hooks we can change the default to false and later remove that option.

jameslamb:
> Also FYI: https://github.com/rapidsai/shared-workflows/issues/46#issuecomment-1568777611

There's a list of a few repos needing this update.

Thanks! Hadn't seen that before! I can make those changes in those other repos right now, shouldn't take that long.

> Also we can disable the dependency check from the checks job now that it's covered by style checks.

done in https://github.com/rapidsai/cucim/pull/682/commits/4cf17d3568e4246c6ae10f88787e65daacb73aaf

jameslamb:
> done in https://github.com/rapidsai/cucim/commit/4cf17d3568e4246c6ae10f88787e65daacb73aaf

On second thought... I just reverted that.

@bdice instead of checking in configuration like this in all the repos:

```yaml
with:
      enable_check_generated_files: false
```

I think it'd be preferable to just add those pre-commit hooks to the remaining repos, and then remove the `rapids-dependency-file-generator` functionality entirely from the configuration of `checks` in `shared-workflows`.

That'd avoid the need to come back and do another round of PRs removing the `with: enable_check_generated_files` from configs here. Re-reading it, I think that's what @jakirkham was implying in https://github.com/rapidsai/cucim/pull/682#issuecomment-1909082275 too.

What do you think?

bdice:
> That'd avoid the need to come back and do another round of PRs

That is smarter. Do that!

jakirkham:
Yep thinking the same thing. We can apply those cycles to some other goal

jameslamb:
> I can make those changes in those other repos right now

This didn't happen. I did search for which ones need it and summarized that at https://github.com/rapidsai/shared-workflows/issues/46#issuecomment-1912259754.

raydouglass (approved):

grlee77 (approved):

jakirkham:
Thanks all! 🙏

PR #683, opened 2024-01-26
Consolidate test requirements in `dependencies.yaml`

jakirkham:
Instead of having test requirements in `requirements-test.txt` and `dependencies.yaml`, consolidate test requirements into `dependencies.yaml` and rely on that (and where it propagates) to handle installing test dependencies

### DIFF
--- a/python/cucim/requirements-test.txt
+++ b/python/cucim/requirements-test.txt
@@ -1,9 +1,0 @@
-GPUtil>=1.4.0
-imagecodecs>=2021.6.8
-openslide-python>=1.3.0
-opencv-python-headless>=4.6
-psutil>=5.8.0
-pytest>=6.2.4
-pytest-cov>=2.12.1
-pytest-lazy-fixture>=0.6.3
-tifffile>=2022.7.28
--- a/run
+++ b/run
@@ -608,11 +608,13 @@ install_python_test_deps_() {
         # (https://github.com/rapidsai/cucim/pull/433)
         run_command pip install -r ${TOP}/python/cucim/requirements-test.txt
     else
+        pushd "${TOP}/python/cucim"
         if [ -n "${VIRTUAL_ENV}" ]; then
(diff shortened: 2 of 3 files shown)


python/cucim/requirements-test.txt
```diff
@@ -1,9 +0,0 @@
-GPUtil>=1.4.0
-imagecodecs>=2021.6.8
-openslide-python>=1.3.0
-opencv-python-headless>=4.6
-psutil>=5.8.0
-pytest>=6.2.4
-pytest-cov>=2.12.1
-pytest-lazy-fixture>=0.6.3
-tifffile>=2022.7.28
```
jakirkham, line 9: These are already contained in `pyproject.toml`'s optional `test` dependencies

https://github.com/rapidsai/cucim/blob/bf5a6f765de64b522ac060f92e24c52f69834c56/python/cucim/pyproject.toml#L61-L75

run
```diff
@@ -608,11 +608,13 @@ install_python_test_deps_() {
...
     else
+        pushd "${TOP}/python/cucim"
         if [ -n "${VIRTUAL_ENV}" ]; then
-            run_command pip3 install -r ${TOP}/python/cucim/requirements-test.txt
+            run_command pip3 install -e .[test]
         else
-            run_command pip3 install --user -r ${TOP}/python/cucim/requirements-test.txt
+            run_command pip3 install --user -e .[test]
         fi
+	popd
```
jakirkham, line 617: Hence we can just install `cucim` in development mode with its `test` dependencies

jakirkham (commented):
Provided a little more context on the change below

raydouglass (approved):

jakirkham:
Thanks all! 🙏

run
jakirkham: Realized the indentation here is wrong. Fixing with PR ( https://github.com/rapidsai/cucim/pull/693 )

PR #694, opened 2024-02-06
Update ops-bot.yaml

AyodeAwe:
This PR enables the ForwardMerger [ops-bot](https://github.com/rapidsai/ops-bot) plugin.


AyodeAwe:
> The `ops-bot.yaml` configuration file does not take affect unless it's committed to the default branch of a particular repository.
> 
> So this should be merged into `branch-24.04`.
> 
> Approved otherwise.

Oops, missed that detail. Thanks.

PR #699, opened 2024-02-09
Refactor CUDA libraries in dependencies.yaml

bdice:
Follow-up from #672. This fixes an issue where libcufile-dev could be included in aarch64 environments (this path was never called in CI so it wasn't a huge problem).

I also fixed some duplication in dependencies.yaml. The CUDA compilers (for 11 and 12) are now included in the `build` dependency list, and all CUDA libraries are included in the `cuda` dependency list. As before, the CUDA version is constrained by the `cuda_version` dependency list. This is more aligned with how cudf's dependency list is structured.

### DIFF
--- a/dependencies.yaml
+++ b/dependencies.yaml
@@ -115,24 +115,15 @@ dependencies:
               cuda: "11.8"
             packages:
               - nvcc_linux-64=11.8
-              - libcufile=1.4.0.31
-              - libcufile-dev=1.4.0.31
-              - libnvjpeg=11.6.0.55
-              - libnvjpeg-dev=11.6.0.55
           - matrix:
               arch: aarch64
               cuda: "11.8"
             packages:
               - nvcc_linux-aarch64=11.8
-              - libnvjpeg=11.6.0.55
-              - libnvjpeg-dev=11.6.0.55
           - matrix:
               cuda: "12.*"
             packages:
               - cuda-nvcc
-              - libcufile-dev
-              - libnvjpeg-dev
-              - libnvjpeg-static
   checks:
     common:
       - output_types: [conda, requirements]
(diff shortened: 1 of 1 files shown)


dependencies.yaml
```diff
@@ -115,24 +115,15 @@ dependencies:
...
               cuda: "11.8"
             packages:
               - nvcc_linux-aarch64=11.8
-              - libnvjpeg=11.6.0.55
-              - libnvjpeg-dev=11.6.0.55
           - matrix:
               cuda: "12.*"
             packages:
               - cuda-nvcc
-              - libcufile-dev
```
bdice, line 133: This matrix could be included on aarch64, which would pull in libcufile-dev. That isn't what we want. Thankfully we never ran into this in practice, to my knowledge.

dependencies.yaml
```diff
@@ -178,7 +169,40 @@ dependencies:
...
+              cuda: "11.8"
+            packages:
+              - libnvjpeg=11.6.0.55
+              - libnvjpeg-dev=11.6.0.55
+          - matrix:
+              arch: aarch64
+              cuda: "11.8"
+            packages:
+              - libnvjpeg=11.6.0.55
+              - libnvjpeg-dev=11.6.0.55
```
jakirkham, line 185: Maybe I'm missing it. Is there a difference between these two cases or can they be consolidated?

jakirkham (commented):
Thanks Bradley! 🙏

Had a question below

dependencies.yaml
bdice: These can be consolidated! I just copied from before, where they were split. Done in 057200c.

jakirkham: Ah ok. Sorry we were overly verbose before

AyodeAwe (approved):

jakirkham (approved):

jakirkham:
Thanks all! 🙏

PR #704, opened 2024-02-26
Add support for Python 3.11, require NumPy 1.23+

jameslamb:
## Description

Contributes to https://github.com/rapidsai/build-planning/issues/3

This PR adds support for Python 3.11.

It also bumps uses of `NumPy` to `numpy>=1.23`, see https://github.com/rapidsai/build-planning/issues/3#issuecomment-1967952280.

## Notes for Reviewers

This is part of ongoing work to add Python 3.11 support across RAPIDS.

The Python 3.11 CI workflows introduced in https://github.com/rapidsai/shared-workflows/pull/176 are *optional*... they are not yet required to run successfully for PRs to be merged.

This PR can be merged once all jobs are running successfully (including the non-required jobs for Python 3.11). The CI logs should be verified that the jobs are building and testing with Python 3.11.

See https://github.com/rapidsai/shared-workflows/pull/176 for more details.

### DIFF
--- a/conda/recipes/cucim/meta.yaml
+++ b/conda/recipes/cucim/meta.yaml
@@ -63,7 +63,7 @@ requirements:
     {% endif %}
     - cupy >=12.0.0
     - libcucim ={{ version }}
-    - numpy 1.21
+    - numpy 1.23
     - python
     - scikit-image >=0.19.0,<0.23.0a0
     - scipy
--- a/benchmarks/skimage/_image_bench.py
+++ b/benchmarks/skimage/_image_bench.py
@@ -12,7 +12,6 @@ import numpy as np
 import pandas as pd
 import scipy.ndimage
 import skimage.data
-
 from cucim.time import repeat
 
 
--- a/benchmarks/skimage/bench_convolve.py
+++ b/benchmarks/skimage/bench_convolve.py
@@ -3,8 +3,6 @@ Benchmark locally modified ndimage functions vs. their CuPy counterparts
 """
 import cupy as cp
 import cupyx.scipy.ndimage as ndi
-from cupyx.profiler import benchmark
-
 from cucim.skimage._vendored.ndimage import (  # noqa: F401
     convolve1d,
     correlate1d,
@@ -18,6 +16,7 @@ from cucim.skimage._vendored.ndimage import (  # noqa: F401
     uniform_filter,
     uniform_filter1d,
 )
+from cupyx.profiler import benchmark
 
 d = cp.cuda.Device()
 
--- a/benchmarks/skimage/cucim_color_bench.py
+++ b/benchmarks/skimage/cucim_color_bench.py
@@ -2,6 +2,8 @@ import argparse
 import os
 import pickle
 
+import cucim.skimage
+import cucim.skimage.color
 import cupy
 import cupy as cp
 import cupyx.scipy.ndimage
@@ -11,9 +13,6 @@ import scipy
 import skimage
 import skimage.color
 from _image_bench import ImageBench
-
-import cucim.skimage
-import cucim.skimage.color
 
 func_name_choices = [
     "convert_colorspace",
--- a/benchmarks/skimage/cucim_exposure_bench.py
+++ b/benchmarks/skimage/cucim_exposure_bench.py
@@ -2,6 +2,8 @@ import argparse
 import os
 import pickle
 
+import cucim.skimage
+import cucim.skimage.exposure
 import cupy
 import cupy as cp
 import numpy as np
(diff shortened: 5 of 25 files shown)


jakirkham:
cc @thewtex (for awareness)

jakirkham:
It looks like the `import`s were reordered in a bunch of places. Is this from `isort` or something?

Also are there any related updates needed to `.pre-commit-config.yaml`?

KyleFromNVIDIA:
https://github.com/rapidsai/cucim/actions/runs/8070438319/job/22048142083?pr=704 is failing because we're building against numpy 1.21.3, which doesn't support Python 3.11.

jameslamb:
> because we're building against numpy 1.21.3, which doesn't support Python 3.11

Thanks for finding that!

I just did a bit of research, sharing some notes that might help inform what we do about this.

From https://numpy.org/news/#numpy-1240-released, it looks like `numpy==1.24.*` (December 2022) was the first one to fully support Python 3.11.

It looks like `numpy==1.23.2` (August 2022) was the first to have CPython 3.11 wheels (https://pypi.org/project/numpy/1.23.2/#files).

`cudf` is allowing `numpy` to float, `numpy>=1.21` ([cudf/pyproject.toml](https://github.com/rapidsai/cudf/blob/ef20ef9af942d840effd8ac16f8b302a8e4b1dd4/python/cudf/pyproject.toml#L33)).

`cucim` is not pinning to a specific `numpy` in its wheel runtime dependencies

https://github.com/rapidsai/cucim/blob/47e0f74ce5299f5ac07ddfb4a506619f1026c734/python/cucim/pyproject.toml#L29

But is pinning to `numpy==1.21.*` in its conda runtime dependencies

https://github.com/rapidsai/cucim/blob/47e0f74ce5299f5ac07ddfb4a506619f1026c734/conda/recipes/cucim/meta.yaml#L66

KyleFromNVIDIA:
It was pinned specifically to 1.21 in #480, because that's the oldest version of numpy we support.

bdice:
Note the dependency on numpy is in `host`, so it's a build dependency here. We try to always build against the oldest supported numpy version, because numpy requires `run_version >= (build or host)_version`.

conda-forge is using `numpy 1.23` with `python 3.11`: https://github.com/conda-forge/conda-forge-pinning-feedstock/blob/bd0cc0a05ae8fc09f687ca52fcdc0cf4acbb5306/recipe/conda_build_config.yaml#L694-L699

We should be able to update this pin to `1.23` wherever numpy is used as a `host` dependency. This also applies to some other packages like RAFT.

jakirkham:
Yeah we [in Python 3.11 NumPy was bumped to 1.23]( https://github.com/conda-forge/conda-forge-pinning-feedstock/pull/3573/files#diff-779c41a6faa73aa6be302001fee6bf260b30769fd5e84a2ad206e96c2da9cf2dR30-R31 )

The simplest solution would be to bump to NumPy 1.23 here

If for some reason that does not work, we can use NumPy 1.21 for older Python versions and NumPy 1.23 for Python 3.11

conda/recipes/cucim/meta.yaml
```diff
@@ -63,7 +63,7 @@ requirements:
     {% endif %}
     - cupy >=12.0.0
     - libcucim ={{ version }}
-    - numpy 1.21
+    - numpy 1.23
```
jakirkham, line 66: Just for illustrative purposes, this is what it would look like to pin NumPy to 1.21 for Python pre-3.11 and to NumPy 1.23 starting with Python 3.11

```suggestion
    - numpy 1.21  # [py < 311]
    - numpy 1.23  # [py >= 311]
```

KyleFromNVIDIA: Is `311` a valid value? Perhaps the [documentation](https://docs.conda.io/projects/conda-build/en/stable/user-guide/environment-variables.html#environment-variables-that-affect-the-build-process) is outdated, but it seems to indicate that `37` is the last valid value for this variable.

jakirkham: Yep, here is [an example]( https://github.com/conda-forge/conda-feedstock/blob/010d41345593b8c0d98edc8fec4e8a6811c1839c/recipe/meta.yaml#L99-L100 )

That said, am ok with just using NumPy 1.23 everywhere

jameslamb:
> ⚠️ Note the dependency on numpy is in host, so it's a build dependency here, not a runtime dependency

Ah yeah sorry about that, I could have chosen a better link. It's build dependency **AND** a runtime dependency, via this

https://github.com/rapidsai/cucim/blob/47e0f74ce5299f5ac07ddfb4a506619f1026c734/conda/recipes/cucim/meta.yaml#L75

right?

jakirkham (approved):

jakirkham:
Thanks all! 🙏

PR #712, opened 2024-03-12
Add upper bound to prevent usage of NumPy 2

bdice:
NumPy 2 is expected to be released in the near future. For the RAPIDS 24.04 release, we will pin to `numpy>=1.23,<2.0a0`. This PR adds an upper bound to affected RAPIDS repositories.

xref: https://github.com/rapidsai/build-planning/issues/29


jakirkham:
Thanks Bradley! 🙏

Could we add this to wheels too? Looks like it is Conda only atm

bdice:
Wheels added in d7d067e.

jakirkham:
Looks good. Thanks Bradley! 🙏

raydouglass (approved):

python/cucim/pyproject.toml
```diff
@@ -27,6 +27,7 @@ dependencies = [
     "cupy-cuda11x>=12.0.0",
     "lazy_loader>=0.1",
     "numpy",
+    "numpy>=1.23.4,<2.0a0",
```
raydouglass, line 30: Probably doesn't matter, but something is causing numpy to appear twice.

grlee77 (approved):
Thanks, LGTM!

PR #713, opened 2024-03-18
prevent path conflict

AyodeAwe:
This will make builds fail when there are path conflicts


jolorunyomi (approved):

jakirkham:
Thanks Jake and Joseph! 🙏

PR #714, opened 2024-03-25
Fix docs upload directory

raydouglass:
Similar to https://github.com/rapidsai/cudf/pull/15291, fixes the upload directory for docs (`24.04.00a30` => `24.04`).

### DIFF
--- a/ci/build_docs.sh
+++ b/ci/build_docs.sh
@@ -2,7 +2,7 @@
 
 set -euo pipefail
 
-export RAPIDS_VERSION_NUMBER="$(rapids-generate-version)"
+export RAPIDS_VERSION_NUMBER="$(rapids-version-major-minor)"
 
 rapids-logger "Create test conda environment"
 . /opt/conda/etc/profile.d/conda.sh


ajschmidt8 (approved):

AyodeAwe (approved):

ajschmidt8 (approved):
Looks good. Though there is nothing currently preventing an accident like this from happening again.

I'm thinking we should add the following two checks in [this function](https://github.com/rapidsai/gha-tools/blob/a75d2b5dd1b01129490b478e38f18eb6e35170ac/tools/rapids-upload-docs#L21-L34):

- check to see if the `RAPIDS_VERSION_NUMBER` variable matches a `0.38` or `24.06` regex pattern
- check to see whether [each HTML directory](https://github.com/rapidsai/gha-tools/blob/a75d2b5dd1b01129490b478e38f18eb6e35170ac/tools/rapids-upload-docs#L3-L14) has an `index.html` file

That will help ensure that we don't upload any faulty doc versions or directories.

Issue #703, opened 2024-02-23
[BUG] `click` is listed only in test requirements but is a required in order to use `cucim.clara`


grlee77:
**Describe the bug**
`click` should not be required unless the CLI `convert` tool based on it is used.

**Steps/Code to reproduce bug**

@drbeh reported that `pip` installing the recent PyPI release of 24.02 and then trying
```py
from cucim import CuImage
```
results in 
```
ModuleNotFoundError: No module named 'click'
```

**Expected behavior**
It is possible to use `cucim.clara` classes other than the `convert` command-line utility if `click` is not installed.

**Environment details (please complete the following information):**

**Additional context**

jakirkham:
Thanks Greg! 🙏

We should also consider adding a test that tries `import`s without the test dependencies installed. This could flag similar issues for us in the future

jakirkham:
Currently this is just in test requirements

https://github.com/rapidsai/cucim/blob/eee3e4c3ae6521b532978f20e7083a5a3ea8287d/python/cucim/pyproject.toml#L57-L60

PR #719, opened 2024-04-02
Require `click` as a wheel dependency

jakirkham:
Fixes https://github.com/rapidsai/cucim/issues/703

Trying to `import cucim` results in `import click`, which fails if `click` is not installed. This was encountered recently by a user that tested out the cuCIM wheel packages.

The Conda packages already solve this by adding `click` to `requirements/run`. So this makes an equivalent change to the wheel packages.

Should fix user install issues.

### DIFF
--- a/dependencies.yaml
+++ b/dependencies.yaml
@@ -244,6 +244,7 @@ dependencies:
     common:
       - output_types: [conda, requirements, pyproject]
         packages:
+          - click
           - lazy_loader>=0.1
           - numpy>=1.23.4,<2.0a0
           - scikit-image>=0.19.0,<0.23.0a0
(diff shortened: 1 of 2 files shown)


grlee77 (approved):
Thanks, @jakirkham, LGTM!

gigony (approved):
Thanks @jakirkham! Looks good to me!

raydouglass (approved):

jakirkham:
Ran into a network error in one job. Mentioned in this upstream issue ( https://github.com/conda/infrastructure/issues/906#issuecomment-2033231631 ). Have also restarted it

jakirkham:
Thanks all! 🙏

PR #724, opened 2024-04-16
removal of deprecated functions/kwargs scheduled for release 24.06

grlee77:
This MR completes a few scheduled deprecations
- removes renamed `get_xyz_coords` function (it already wasn't accessible from the top-level `cucim.skimage.color` namespace)
- removes deprecated `return_error` kwarg from `phase_cross_correlation` (error is always returned)
- removes deprecated `random_state` kwarg from `medial_axis` (it was renamed to `rng` previously)


gigony (approved):
Thanks @grlee77 for this update!
Looks good to me!

PR #725, opened 2024-04-17
Remove code needed to support Python < 3.9 and apply ruff's pyupgrade rules

grlee77:
`math.prod` was added in Python 3.8
`math.lcm` was added in Python 3.9

`pyproject.toml` was updated to have `ruff` also check [pyupgrade rules](https://docs.astral.sh/ruff/rules/#pyupgrade-up)

PR #726, opened 2024-04-18
Update the test criteria for test_read_random_region_cpu_memleak

gigony:
We are seeing intermittent errors in CI/CD

See https://github.com/rapidsai/cucim/pull/725#issuecomment-2062326942

>Seeing one test failure in [this CI job]( https://github.com/rapidsai/cucim/actions/runs/8716002781/job/23909243290?pr=725#step:8:495 ):

```python
_ test_read_random_region_cpu_memleak[testimg_tiff_stripe_4096x4096_256_deflate] _
[gw6] linux -- Python 3.9.19 /pyenv/versions/3.9.19/bin/python
python/cucim/tests/performance/clara/test_read_region_memory_usage.py:100: in test_read_random_region_cpu_memleak
    assert mem_usage_history[-1] - mem_usage_history[1] < iteration * 100
E   assert (828035072 - 826982400) < (1000 * 100)
```

I couldn't determine which part of the code is increasing memory usage, but it's clear that there is no memory leak on the C++ side, as memory usage does not continuously increase per iteration.

This patch updates the validation logic to improve reliability:
- Increase the number of iterations by 10 times.
- Ensure that the memory leak per iteration does not exceed 768 bytes.
- Verify whether the increase in memory usage is continuous.

### DIFF
--- a/python/cucim/tests/performance/clara/test_read_region_memory_usage.py
+++ b/python/cucim/tests/performance/clara/test_read_region_memory_usage.py
@@ -72,6 +72,7 @@ def test_read_region_cpu_memleak(testimg_tiff_stripe_4096x4096_256):
 
 
 def test_read_random_region_cpu_memleak(testimg_tiff_stripe_4096x4096_256):
+    import gc
     import os
     import random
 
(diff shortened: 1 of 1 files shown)


grlee77 (approved):
looks good to me, the prior failure seems to have been fairly rare so probably just have to see over time whether any failures still occur.

jakirkham (approved):
Thanks Gigon and Greg! 🙏

PR #727, opened 2024-04-21
Implement upstream changes from scikit-image 0.23 (part 1 of 2)

grlee77:
This MR updates cuCIM for consistency with version scikit-image `0.23.2rc1`. There were a relatively large number of changes in the `morphology` module, so I am going to submit those in a subsequent MR. I will also bump the scikit-image version pinning as part of that second MR.

The commits in this MR can be reviewed individually. Much of the new code is private functions in `cucim.skimage._shared` that are copied from upstream and don't need to be reviewed in detail.

The highlights are:

- `cucim.skimage.measure.regionprops` (and `regionprops_table`) support one new region property: `intensity_std` 
- `cucim.skimage.segmentation.expand_labels` now supports a `spacing` keyword argument to take a pixel's physical dimensions into account.
- additional modules use `__init__.pyi` instead of just `__init__.py`
- various fixes to documentation strings (consistent shape notation, etc.)
- adopted the private `_shared/compat.py` for future NumPy 2.0/SciPy compatibility
- adopted private test utility `assert_stacklevel`
- updated the private utility functions for handling deprecations to match those upstream
  - update existing decorators to the new ones
  - new rename of output->out for `cucim.skimage.filters.gaussian`


gigony (approved):
Thanks @grlee77 for this change!
It looks good to me!

PR #728, opened 2024-04-21
Implement upstream changes from scikit-image 0.23 (part 2 of 2: morphology)

grlee77:
There were a number of non-trivial changes to the morphology module, so I broke those out from the other changes in #727. Please review and merge that MR first before reviewing this one.

Highlights from upstream are:
- binary morphology functions have a new `mode` argument that controls how values outside the image boundaries are interpreted
- grayscale morphology functions have new `mode` and `cval` arguments that control how boundaries are extended (these were already available in `scipy.ndimage`/`cupyx.scipy.ndimage`, they just weren't exposed via the `skimage`/`cuCIM` APIs)
- binary and grayscale morphology functions have bug fixes in the case of even-sized/non-symmetric footprints
  - additional corresponding test cases were added

Aside from the upstream changes, novel changes in this MR are:
- refactored utility functions to mirror and pad the footprints to allow use with the cuCIM-specific optimization of passing a tuple for rectangular footprints instead of explicitly allocating a GPU footprint array
- refactored some test cases to better use `pytest.mark.parametrize`
- some grayscale tests now compare directly to `skimage` CPU outputs instead fetching previously saved values
- bumped our version pinning for scikit-image to allow 0.23.x to be installed

I marked as "non-breaking" as the existing behavior has not changed except in the case of the bug fixes for even-sized footprints.


python/cucim/src/cucim/skimage/morphology/gray.py
```diff
@@ -716,12 +779,18 @@ def black_tophat(image, footprint=None, out=None):
 
     """
     if out is image:
-        original = image.copy()
-    else:
-        original = image
-    out = closing(image, footprint, out=out)
+        # We need a temporary image
+        # We need a temporary image
```
gigony, line 783: Looks like a redundant comment.

```suggestion
```

gigony (approved):
Thanks @grlee77 !
Looks good to me! (That's a lot of changes! ;) )

AyodeAwe (approved):

PR #734, opened 2024-05-01
Enable FutureWarnings/DeprecationWarnings as errors

mroeschke:
As part of https://github.com/rapidsai/build-planning/issues/26, enabling Python test failures for `FutureWarning`s and `DeprecationWarning`s

### DIFF
--- a/python/cucim/src/cucim/clara/converter/tiff.py
+++ b/python/cucim/src/cucim/clara/converter/tiff.py
@@ -73,11 +73,12 @@ def svs2tif(
     logger.info("      num_workers: %d", num_workers)
     logger.info("      compression: %s", compression)
     logger.info("  output filename: %s", output_filename)
-
+    compressionargs = None
     if compression is not None:
         # handles only jpeg or None (no compression)
         if compression.lower() == "jpeg":
-            compression = ("jpeg", 95)
+            compression = "jpeg"
+            compressionargs = {"level": 95}
         else:
             raise ValueError(
                 f"Unsupported compression: {compression}."
@@ -201,6 +202,7 @@ def svs2tif(
                             resolution_unit,
                         ),
                         compression=compression,  # requires imagecodecs
+                        compressionargs=compressionargs,
                         subfiletype=subfiletype,
                     )
                 logger.info("Done.")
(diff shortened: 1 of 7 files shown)


python/cucim/src/cucim/clara/converter/tiff.py
```diff
@@ -73,11 +73,12 @@ def svs2tif(
     logger.info("      num_workers: %d", num_workers)
     logger.info("      compression: %s", compression)
     logger.info("  output filename: %s", output_filename)
-
+    compressionargs = None
```
jakirkham, line 76: Do we want this to be user facing? Or do we want to keep this internal for now?

python/cucim/tests/util/gen_tiff.py
```diff
@@ -58,7 +58,11 @@ def save_image(
 
         compression = COMPRESSION_MAP.get(compression)
```
jakirkham, line 59: Maybe this can just be a default before

```suggestion
        compression = COMPRESSION_MAP.get(compression)
        compressionargs = None
```

python/cucim/tests/util/gen_tiff.py
```diff
@@ -58,7 +58,11 @@ def save_image(
...
         if not compression:
-            compression = ("jpeg", 95)
+            compression = "jpeg"
+        if compression == "jpeg":
+            compressionargs = {"level": 95}
+        else:
+            compressionargs = None
```
jakirkham, line 65:
```suggestion
```

jakirkham (commented):
Thanks Matthew! 🙏

Had a couple questions below

python/cucim/src/cucim/clara/converter/tiff.py
grlee77: If we make it user facing, we may need to add some additional validation on the input. I would be fine with just keeping it internal for now.

@gigony  do you have a preference?

grlee77 (commented):
Thanks for making this changes, I found one comment where bdtr wasn't updated to betainc. Otherwise it looks good to me.

python/cucim/src/cucim/clara/converter/tiff.py
gigony: @grlee77 @jakirkham 
It seems appropriate to update it now since this API change has been in effect since July 2022.
https://github.com/cgohlke/tifffile/blame/ec6be6289db1b8b7327bb98816a764c95b9b7299/README.rst#L519

jakirkham: @grlee77 were there particular thoughts you had on what input validation would look like?

Wonder if tifffile's own validation would work for us

jakirkham (approved):

PR #736, opened 2024-06-04
use rapids-build-backend

jameslamb:
Contributes to https://github.com/rapidsai/build-planning/issues/31
Contributes to https://github.com/rapidsai/dependency-file-generator/issues/89

Proposes introducing `rapids-build-backend` as this project's build backend, to reduce the complexity of various CI/build scripts.


bdice (approved):

dependencies.yaml
```diff
@@ -266,9 +279,18 @@ dependencies:
...
+        matrices:
+          - matrix: {cuda: "12.*"}
+            packages:
+              - cupy-cuda12x>=12.0.0
+          - matrix: {cuda: "11.*"}
+            packages:
+              - &cupy_cu11 cupy-cuda11x>=12.0.0
+          - matrix:
+            packages:
+              - *cupy_cu11
```
grlee77, line 293: unfamiliar with the notation, curious what the `&` and `*` in the package names do here?

grlee77 (approved):
Thanks, the Python code changes look good to me

dependencies.yaml
bdice: These are YAML anchors. We defined an anchor with & and reused its value with *.

jameslamb: oh sure, happy to explain! This is a YAML "alias", pointing to a YAML "anchor". When this document is loaded, `*cupy_cu11` will be replaced with whatever value is tagged with `&cupy_cu11`. It's a way to reduce a bit of duplication, to keep things that should be consistent consistent.

For example, this:

```yaml
limits:
  cpu: 1.0
requests:
  cpu: 1.0
```

Is identical to this:

```yaml
limits:
  cpu: &cpu_val 1.0
requests:
  cpu: *cpu_val
```

For reference, see:

* https://support.atlassian.com/bitbucket-cloud/docs/yaml-anchors/
* https://yaml.org/spec/1.2.2/#3222-anchors-and-aliases

raydouglass (approved):

PR #745, opened 2024-07-05
skip CMake 3.30.0, require CMake >=3.26.4

jameslamb:
## Description

Contributes to https://github.com/rapidsai/build-planning/issues/80

Adds constraints to avoid pulling in CMake 3.30.0, for the reasons described in that issue.

## Notes for Reviewers

Most of the rest of RAPIDS is using `cmake >=3.26.4`... but I didn't change the floor here to match that, because I didn't know if it was intentional that `cucim` allows a wider range of `cmake` versions.

Let me know if you'd prefer I bump the `cmake` minimum version to `3.26.4` as part of this.


bdice (changes requested):
Can we update this repo to use the same pinnings as the rest of RAPIDS? Something like `>=3.26.4,!=3.30.0`.

jameslamb:
> Can we update this repo to use the same pinnings as the rest of RAPIDS?

Sure! Like I mentioned in the description, I wasn't sure if this repo was intentionally using a more permissive range than others.

Just pushed commits moving the floor to `3.26.4` and updated the PR title.

bdice (approved):
Thanks. I think it was an oversight that we had not kept CMake pinnings aligned with the rest of RAPIDS in past CMake update migrations.

KyleFromNVIDIA (approved):

grlee77 (approved):
Thanks for updating the CMake pinning. I don't think we had any specific requirement to support older versions, so the update here seems fine.

Issue #742, opened 2024-06-25
Explore NumPy 2 compatibility


jakirkham:
Would be good to explore NumPy 2 compatibility in cuCIM and see what changes are needed to support NumPy 1 & 2.

* [NumPy Release Notes]( https://numpy.org/devdocs/release/2.0.0-notes.html )
* [NumPy 2 Migration Guide]( https://numpy.org/devdocs/numpy_2_0_migration_guide.html#numpy-2-migration-guide )
* [CuPy 13.2.0 Release Notes]( https://github.com/cupy/cupy/releases/tag/v13.2.0 ) (compatible with NumPy 2)
* RAPIDS NumPy 2 Tracking issue: https://github.com/rapidsai/build-planning/issues/38

jakirkham:
cc @grlee77 (for vis)

grlee77:
## Current issues running cuCIM with NumPy 2.0 and CuPy 13.2

I ran the cuCIM test suite locally with NumPy 2.0 and CuPy 13.2. The remaining compatibility issues encountered are documented here:

### 1.) use of `cupy.full` seems broken due to use of `numpy.can_cast`. Here is an example traceback for a function using it
```
measure/_moments.py:423: in moments_normalized
    nu = cp.full(mu.shape, cp.nan, dtype=mu.dtype)
/home/grelee/mambaforge/envs/np2/lib/python3.10/site-packages/cupy/_creation/basic.py:325: in full
    cupy.copyto(a, fill_value, casting='unsafe')
/home/grelee/mambaforge/envs/np2/lib/python3.10/site-packages/cupy/_manipulation/basic.py:38: in copyto
    can_cast = numpy.can_cast(src, dst.dtype, casting)
```

**Workaround**: modify this line to always set `can_cast = True` (this works for purposes of cuCIM, but isn't the right general solution)


### 2.) use of `cp.random.seed(1234)` with an integer seed also fails due to `numpy.can_cast`

```
metrics/tests/test_structural_similarity.py:19: in <module>
    cp.random.seed(1234)
/home/grelee/mambaforge/envs/np2/lib/python3.10/site-packages/cupy/random/_generator.py:1274: in seed
    get_random_state().seed(seed)
/home/grelee/mambaforge/envs/np2/lib/python3.10/site-packages/cupy/random/_generator.py:813: in seed
    numpy.asarray(seed).astype(numpy.uint64, casting='safe'))
[...]

PR #746, opened 2024-07-10
Minor fixes for NumPy 2.0 compatiblity

grlee77:
This MR resolves issues 3 and 4 reported [here](https://github.com/rapidsai/cucim/issues/742#issuecomment-2218492642) when testing with NumPy 2.0 and CuPy 13.2 locally (issues 1 and 2 will be addressed in CuPy itself)
https://github.com/rapidsai/cucim/issues/742#issuecomment-2218492642

For now, leave NumPy pinning as-is. Tests should continue to pass with NumPy 1.x

### DIFF
--- a/python/cucim/src/cucim/skimage/segmentation/_chan_vese.py
+++ b/python/cucim/src/cucim/skimage/segmentation/_chan_vese.py
@@ -429,6 +429,7 @@ def chan_vese(
         energies = []
     phivar = tol + 1
 
+    dt = cp.asarray(dt, dtype=float_dtype)
     while phivar > tol and i < max_num_iter:
         # Save old level set values
         oldphi = phi
--- a/python/cucim/tests/unit/core/test_stain_normalizer.py
+++ b/python/cucim/tests/unit/core/test_stain_normalizer.py
@@ -137,7 +137,7 @@ class TestStainExtractorMacenko:
                 stain_extraction_pca(image)
         else:
             result = stain_extraction_pca(image)
-            cp.testing.assert_allclose(result, expected)
+            cp.testing.assert_allclose(result, expected, rtol=1e-6)
 
 
 class TestStainNormalizerMacenko:


jakirkham (approved):
Thanks Greg! 🙏

python/cucim/src/cucim/skimage/segmentation/_chan_vese.py
```diff
@@ -429,6 +429,7 @@ def chan_vese(
         energies = []
     phivar = tol + 1
 
+    dt = cp.asarray(dt, dtype=float_dtype)
```
seberg, line 432: LGTM to me also.

Just FYI, I'll note that in some places using `float(dt)` may also make sense to ensure "weak" promotion handling (i.e. float scalars behave like they used to in most cases on both NumPy 1 and 2).
I don't think this needs a fix here!

jakirkham: Yeah I think here we realized that we were passing a Python `float` to [a CuPy-based custom kernel]( https://github.com/grlee77/cucim/blob/cc787fd64d3e5e97b96eddcc4c64c120dd910c50/python/cucim/src/cucim/skimage/segmentation/_chan_vese.py#L107-L109 ) and relying on whatever default promotion it was doing, which CuPy changed recently. There's more context in 3 under comment: https://github.com/rapidsai/cucim/issues/742#issuecomment-2218492642

So we decided it was better that we determine the type here instead. Hence the change

Hopefully that context is helpful

seberg: Ah, NumPy 2 + CuPy 13.2 interaction might make promotion a bit more tricky even, so I guess the Python float thing may not work there.

jakirkham: Yeah actually in the case where the error was seen, we are using the default value for `dt` (as it is unspecified)

https://github.com/rapidsai/cucim/blob/fbca7fa847f8bf9916ea013dd274d88be43c99fb/python/cucim/src/cucim/skimage/segmentation/tests/test_chan_vese.py#L46

Looking at the default value for `dt`, it appears to be `0.5`

https://github.com/rapidsai/cucim/blob/fbca7fa847f8bf9916ea013dd274d88be43c99fb/python/cucim/src/cucim/skimage/segmentation/_chan_vese.py#L282

So perhaps CuPy has decided to always promote `float` to `float64` before CUDA kernel calls

Greg did the investigation here. So likely knows more about the issue

seberg: All, good, I don't think it's worth to track this down exactly.  Eventually (probably with CuPy 14), this shouldn't be necessary for Python floats at least.  But it is definitely the pragmatic thing right now.

jakirkham:
Thanks Greg and Sebastian! 🙏

jakirkham:
Fixed a minor typo in the PR title and in the CHANGELOG. This was picked up by a style check in 24.10

xref: https://github.com/rapidsai/cucim/pull/759

PR #751, opened 2024-07-31
Drop NumPy build dependency

jakirkham:
Partially addresses issue: https://github.com/rapidsai/build-planning/issues/82

Even though cuCIM currently `#include`s `<pybind11/numpy.h>`, the actual C++ code appears not to use NumPy. So this attempts to drop the header and the NumPy build dependency.

### DIFF
--- a/conda/recipes/cucim/meta.yaml
+++ b/conda/recipes/cucim/meta.yaml
@@ -63,7 +63,6 @@ requirements:
     {% endif %}
     - cupy >=12.0.0
     - libcucim ={{ version }}
-    - numpy 1.23
     - python
     - rapids-build-backend >=0.3.0,<0.4.0.dev0
     - scikit-image >=0.19.0,<0.23.0a0
@@ -74,7 +73,7 @@ requirements:
     {% if cuda_major != "11" %}
     - cuda-cudart
     {% endif %}
-    - {{ pin_compatible('numpy') }}
+    - numpy >=1.23,<2.0a0
     - click
     - cupy >=12.0.0
     - lazy_loader >=0.1
--- a/python/pybind11/cucim_py.cpp
+++ b/python/pybind11/cucim_py.cpp
@@ -19,8 +19,10 @@
 
 #include <fmt/format.h>
 #include <fmt/ranges.h>
+#include <pybind11/buffer_info.h>
 #include <pybind11/operators.h>
 #include <pybind11/pybind11.h>
+#include <pybind11/pytypes.h>
 #include <pybind11/stl.h>
 
 #include <cucim/cuimage.h>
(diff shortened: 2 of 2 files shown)


jakirkham:
Looks like this code might need some tweaks if we proceed further

https://github.com/rapidsai/cucim/blob/0e75c7676dc3dd818ae735d660d0fead88e523ba/python/pybind11/cucim_py.cpp#L448

msarahan (approved):

gigony (approved):
Thanks @jakirkham ! It looks good to me.

I have a question: Did `pybind11/numpy.h` depends on numpy header or numpy library?

If this change is about for handling numpy 2, 
can upgrading pybind11 library to the latest version helps?
Looks like pybind11 is handling numpy 2 case - https://github.com/pybind/pybind11/blob/master/include/pybind11/numpy.h#L187

jakirkham:
> I have a question: Did `pybind11/numpy.h` depends on numpy header or numpy library?

A few things to unpack here

NumPy is atypical in its setup. When building against NumPy, one only `#include`s the NumPy header. There is not a NumPy library that one links against in the typical sense

However the symbols that the NumPy header names are in the Python shared objects that the NumPy package ships. Those symbols get loaded when calling `import numpy` (there is a similar operation that [NumPy supplies for use in C APIs]( https://numpy.org/doc/1.26/reference/c-api/array.html#importing-the-api )). So this is how the symbols get resolved at runtime

Regardless, from a developer's perspective, building against NumPy always means using the header and the libraries. There isn't a way to pick just one or the other

Using pybind11 for NumPy support is not unique in this regard

> If this change is about for handling numpy 2, can upgrading pybind11 library to the latest version helps? Looks like pybind11 is handling numpy 2 case - https://github.com/pybind/pybind11/blob/master/include/pybind11/numpy.h#L187

Yes, it is true that [pybind11 2.12.0 ships with NumPy 2 support]( https://pybind11.readthedocs.io/en/stable/upgrade.html#v2-12 ). Building against that would be sufficient for NumPy 1 & 2 support (without other changes)

That said, there are relatively few cases where the NumPy API is strictly needed. Especially after the introduction of [the Python Buffer Protocol]( https://docs.python.org/3/c-api/buffer.html ). Many use cases (and ours in cuCIM is one of these) simply need a way to access the underlying memory buffer of Python objects (NumPy arrays or otherwise). So in these cases, it is better to use the Python Buffer Protocol directly (as this code change does), which works not only with NumPy arrays, but any object that supports the Python Buffer Protocol. As a result this simplifies our dependencies. Plus this approach is more flexible and interoperable with other libraries

gigony:
> > I have a question: Did `pybind11/numpy.h` depends on numpy header or numpy library?
> 
> A few things to unpack here
> 
> NumPy is atypical in its setup. When building against NumPy, one only `#include`s the NumPy header. There is not a NumPy library that one links against in the typical sense
> 
> However the symbols that the NumPy header names are in the Python shared objects that the NumPy package ships. Those symbols get loaded when calling `import numpy` (there is a similar operation that [NumPy supplies for use in C APIs](https://numpy.org/doc/1.26/reference/c-api/array.html#importing-the-api)). So this is how the symbols get resolved at runtime
> 
> Regardless, from a developer's perspective, building against NumPy always means using the header and the libraries. There isn't a way to pick just one or the other
> 
> Using pybind11 for NumPy support is not unique in this regard
> 
> > If this change is about for handling numpy 2, can upgrading pybind11 library to the latest version helps? Looks like pybind11 is handling numpy 2 case - https://github.com/pybind/pybind11/blob/master/include/pybind11/numpy.h#L187
> 
> Yes, it is true that [pybind11 2.12.0 ships with NumPy 2 support](https://pybind11.readthedocs.io/en/stable/upgrade.html#v2-12). Building against that would be sufficient for NumPy 1 & 2 support (without other changes)
> 
> That said, there are relatively few cases where the NumPy API is strictly needed. Especially after the introduction of [the Python Buffer Protocol](https://docs.python.org/3/c-api/buffer.html). Many use cases (and ours in cuCIM is one of these) simply need a way to access the underlying memory buffer of Python objects (NumPy arrays or otherwise). So in these cases, it is better to use the Python Buffer Protocol directly (as this code change does), which works not only with NumPy arrays, but any object that supports the Python Buffer Protocol. As a result this simplifies our dependencies. Plus this approach is more flexible and interoperable with other libraries

Thanks @jakirkham for the comprehensive explanation!
It makes sense, and thank you for the update! 🙂

jakirkham:
Thanks all! 🙏

Issue #755, opened 2024-08-02
Replace or drop `pytest-lazy-fixture`


jakirkham:
After some recent unrelated testing work, we learned that `pytest-lazy-fixture` is incompatible with `pytest` version `8.0.0` or newer. This has been the case for 6 months now as this issue notes: https://github.com/TvoroG/pytest-lazy-fixture/issues/65

Given this we may want to drop or replace this dependency

One option on the replacement side is `pytest-lazy-fixtures`, which it looks straightforward to switch to

jakirkham:
cc @grlee77

grlee77:
It is not used for `cucim.skimage` or `cucim.core` test cases. but is used for some I/O related ones under `python/cucim/tests/`. It is imported [here](https://github.com/rapidsai/cucim/blob/branch-24.10/python/cucim/tests/fixtures/testimage.py) in a helper module loaded by the `python/cucim/tests/conftest.py`.

gigony:
Thanks @jakirkham and @grlee77 !
Let me check if I can replace with `pytest-lazy-fixtures`.

jakirkham:
This was done in PR: https://github.com/rapidsai/cucim/pull/756

Thanks Gigon! 🙏

PR #753, opened 2024-08-01
De-vendor CMake dependencies and use shared libraries

gigony:
This patch de-vendors CMake dependencies and use/link shared libraries as much as possible.

- https://github.com/rapidsai/cucim/issues/686

### Tasks

- [ ] Update CMake files to use the conda package (and fall back to the downloaded source if the `CONDA_PREFIX` environment variable is not available).
  - [ ] cpp/cmake/deps
    - [x] abseil.cmake
    - [x] boost-header-only.cmake
    - [x] catch2.cmake
    - [x] cli11.cmake
    - [x] fmt.cmake
    - [ ] googlebenchmark.cmake
    - [ ] googletest.cmake
    - [x] libcuckoo.cmake   => no conda package available
    - [x] nlohmann_json.cmake
    - [x] nvtx3.cmake
    - [x] taskflow.cmake
  - [ ] python/cmake/deps
    - [x] fmt.cmake
    - [x] nlohmann_json.cmake
    - [x] pybind11_json.cmake
    - [x] pybind11.cmake
  - [ ] cpp/plugins/cucim.kit.cuslide/cmake/deps
    - [x] catch2.cmake
    - [x] cli11.cmake
    - [x] fmt.cmake
    - [ ] googlebenchmark.cmake
    - [ ] googletest.cmake
    - [ ] libculibos.cmake
    - [ ] libdeflate.cmake
    - [ ] libjpeg-turbo.cmake
    - [ ] libopenjpeg.cmake
    - [ ] libtiff.cmake
    - [x] nlohmann_json.cmake
    - [ ] nvjpeg.cmake
    - [ ] openslide.cmake
    - [ ] pugixml.cmake
  - [ ] cpp/plugins/cucim.kit.cumed/cmake/deps
    - [x] catch2.cmake
    - [x] cli11.cmake
    - [x] fmt.cmake
    - [ ] googlebenchmark.cmake
    - [ ] googletest.cmake
- [ ] Make sure that required shared libraries are embedded in the xx.libs folder in the Python wheel file:
  - Related to [this file](https://github.com/rapidsai/cucim/blob/a6c4f59c7a1b4c9bc2c8e09697d1f0c3b0cc0311/ci/build_wheel.sh#L36)
[...]

PR #756, opened 2024-08-05
Switch to pytest-lazy-fixtures

gigony:
Remove the `pytest-lazy-fixture` plugin and use the replacement `pytest-lazy-fixtures` package instead.

Address  #755

### DIFF
--- a/3rdparty/LICENSE.pytest-lazy-fixtures
+++ b/3rdparty/LICENSE.pytest-lazy-fixtures
@@ -1,6 +1,6 @@
-The MIT License (MIT)
+MIT License
 
-Copyright (c) 2016 Marsel Zaripov
+Copyright (c) 2023 Anton Petrov
 
 Permission is hereby granted, free of charge, to any person obtaining a copy
 of this software and associated documentation files (the "Software"), to deal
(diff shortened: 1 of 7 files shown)


3rdparty/LICENSE.pytest-lazy-fixtures
```diff
@@ -1,6 +1,6 @@
-The MIT License (MIT)
+MIT License
```
jakirkham, line 1: Given `pytest-lazy-fixtures` bundles the license file in the wheel and conda packages, do we need to bundle the license file here or could it be dropped?

dependencies.yaml
```diff
@@ -303,7 +303,7 @@ dependencies:
           - psutil>=5.8.0
           - pytest>=6.2.4,<8.0.0a0
           - pytest-cov>=2.12.1
-          - pytest-lazy-fixture>=0.6.3
+          - pytest-lazy-fixtures
```
jakirkham, line 306: Do we want a lower bound here as well?

jakirkham (commented):
Thanks Gigon! 🙏

Had a couple questions below

python/cucim/tests/fixtures/testimage.py
```diff
@@ -16,7 +16,7 @@
 import shutil
 
 import pytest
-from pytest_lazyfixture import lazy_fixture
+from pytest_lazy_fixtures import lf as lazy_fixture
```
grlee77, line 19: nice to see that this can continue to be used as-is without further changes to the test cases using the fixture

3rdparty/LICENSE.pytest-lazy-fixtures
gigony: Thanks @jakirkham !

All files under '3rdparty' folder is actually optional and we added the folder for tracking licenses internally.

Since 'LICENSE-3rdparty.md' file on the root folder covers the list of 3rdparty libraries, do you think it is better to remove whole '3rdparty' folder instead of maintaining the folder?

dependencies.yaml
gigony: Thanks @jakirkham !
I can update it if needed.
I wonder how we can determine the lowest version to use here.

3rdparty/LICENSE.pytest-lazy-fixtures
jakirkham: That's a good idea. Let's follow up after we finish the devendoring work: https://github.com/rapidsai/cucim/pull/753

Since some of those dependencies are statically linked now, we do need to include their licenses. Though it is possible after devendoring that we can substantially reduce or drop licenses from this list

With Python only dependencies, we could clean those up before devendoring, but it might make sense to wait until we can do a more thorough cleanup

dependencies.yaml
jakirkham: Perhaps `1.0.0` is good enough to start?

```suggestion
          - pytest-lazy-fixtures>=1.0.0
```

gigony: Great! Thank you! Will update this.

jakirkham (approved):

PR #764, opened 2024-08-20
chan_vese: pass all constants to `_fused_variance_kernel2` as device scalars

grlee77:
This change avoids an issue with an internal `numpy.can_cast` call in CuPy's kernel fusion code when using NumPy 2.0. With this change all cuCIM tests passed using NumPy 2.0.1 and the CuPy 13.3dev branch.

Fixes https://github.com/rapidsai/cucim/issues/742

### DIFF
--- a/python/cucim/src/cucim/skimage/segmentation/_chan_vese.py
+++ b/python/cucim/src/cucim/skimage/segmentation/_chan_vese.py
@@ -7,6 +7,8 @@ from cucim.core.operations.morphology import distance_transform_edt
 from .._shared.utils import _supported_float_type
 from .._vendored import pad
 
+_one = cp.asarray(1.0, dtype=cp.float32)
+
 
 @cp.fuse()
 def _fused_variance_kernel1(eta, x_start, x_mid, x_end, y_start, y_mid, y_end):
@@ -53,7 +55,7 @@ def _fused_hphi_hinv(phi):
 
 @cp.fuse()
 def _fused_variance_kernel2(
-    image, c1, c2, lam1, lam2, phi, K, dt, mu, delta_phi, Csum
+    image, c1, c2, lam1, lam2, phi, K, dt, mu, delta_phi, Csum, one
 ):
     difference_term = image - c1
     difference_term *= difference_term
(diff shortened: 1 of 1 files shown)


python/cucim/src/cucim/skimage/segmentation/_chan_vese.py
```diff
@@ -7,6 +7,8 @@
 from .._shared.utils import _supported_float_type
 from .._vendored import pad
 
+_one = cp.asarray(1.0, dtype=cp.float32)
```
jakirkham, line 10: Do the functions always use `float32`? Or could this be other types?

python/cucim/src/cucim/skimage/segmentation/_chan_vese.py
```diff
@@ -65,7 +67,7 @@ def _fused_variance_kernel2(
     difference_term += term2
 
     new_phi = phi + (dt * delta_phi) * (mu * K + difference_term)
-    out = new_phi / (1 + mu * dt * delta_phi * Csum)
+    out = new_phi / (one + mu * dt * delta_phi * Csum)
```
jakirkham, line 70: If we wanted to generalize the type handling here, we could do something like this. That way we promote `1` to the same `dtype` of the result it is being added to

```suggestion
    out_denom = mu * dt * delta_phi * Csum
    out_denom += out_denom.dtype.type(1)
    out = new_phi / out_denom
```

jakirkham (commented):
Thanks Greg! 🙏

Had a couple thoughts below

python/cucim/src/cucim/skimage/segmentation/_chan_vese.py
grlee77: It can use float64 as well. We could set it based on `float_type` in the chan_vese function instead of doing it globally up here.

python/cucim/src/cucim/skimage/segmentation/_chan_vese.py
grlee77: Thanks, pushed a new commit following this suggestion

jakirkham (approved):

gigony:
Thanks @grlee77 It looks good to me!

gigony (approved):

jakirkham:
Thanks Greg and Gigon! 🙏

Issue #768, opened 2024-08-22
Upgrading to scikit-image 0.24.0


jakirkham:
In late June, [scikit-image released 0.24.0]( https://github.com/scikit-image/scikit-image/releases/tag/v0.24.0 ). Currently we only support scikit-image 0.23.x as latest

https://github.com/rapidsai/cucim/blob/c004b6f540b28fadce3bab78a290e25ea3308095/python/cucim/pyproject.toml#L31

Opening this to track adding support for scikit-image 0.24.0

PR #769, opened 2024-08-22
bump scikit-image upper bound (allow 0.24.x)

grlee77:
Fixes https://github.com/rapidsai/cucim/issues/768


jakirkham:
Thanks Greg! 🙏

Could you please also update these?

https://github.com/rapidsai/cucim/blob/de6bfbb444acca44ef1c35846be7f53fdd70de6a/conda/recipes/cucim/meta.yaml#L68

https://github.com/rapidsai/cucim/blob/de6bfbb444acca44ef1c35846be7f53fdd70de6a/conda/recipes/cucim/meta.yaml#L82

jakirkham:
Went ahead and pushed the suggested change above. Hope that is ok 🙂

jakirkham (approved):
Thanks Greg! 🙏

Confirmed this gets all of the scikit-image requirements up-to-date

Greg and I also discussed offline if code changes were needed as part of this update. We determined no we can just update the dependency without issues

So this is all we need for scikit-image 0.24.0

Approving so we can get this in

PR #770, opened 2024-08-26
Disable custom 2D separable filtering kernels on windows

grlee77:
We currently bundle some custom 2D convolution kernels not present in CuPy. We have additional testing of these internel kernels, but I found that on the Windows platform a subset of these tests currently fail.

This MR disables auto-selection of these kernels on the Windows platform (will always fall back to CuPy's `cupyx.scipy.ndimage` kernels in this case). The test cases that are currently known to fail are skipped for now on Windows. Longer term it would be good to resolve the issues with the kernels, restore the test cases and enable use of these kernels on Windows.

### DIFF
--- a/python/cucim/src/cucim/skimage/_vendored/_ndimage_filters.py
+++ b/python/cucim/src/cucim/skimage/_vendored/_ndimage_filters.py
@@ -1,5 +1,6 @@
 """A vendored subset of cupyx.scipy.ndimage._filters"""
 import math
+import platform
 import warnings
 
 import cupy
@@ -21,6 +22,8 @@ try:
     compile_errors = (ResourceLimitError, CompileException)
 except ImportError:
     compile_errors = (ResourceLimitError,)
+
+_is_not_windows = platform.system() != "Windows"
 
 
 def correlate(input, weights, output=None, mode="reflect", cval=0.0, origin=0):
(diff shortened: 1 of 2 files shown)


jakirkham:
Thanks Greg! 🙏

Can you please write up the test failure in a new issue so we have that for future reference?

grlee77:
uploaded the test failures in #771

python/cucim/src/cucim/skimage/_vendored/_ndimage_filters.py
```diff
@@ -231,7 +234,7 @@ def _correlate_or_convolve1d(
     default_algorithm = False
     if algorithm is None:
         default_algorithm = True
-        if input.ndim == 2 and weights.size <= 256:
+        if input.ndim == 2 and weights.size <= 256 and _is_not_windows:
```
jakirkham, line 237: Would check for Windows before the other stuff. This can shortcut evaluation on Windows

```suggestion
        if _is_not_windows and input.ndim == 2 and weights.size <= 256:
```

jakirkham (commented):
Thanks Greg! 🙏

Have a suggestion below

grlee77:
@jakirkham, please approve if there are no further comments

jakirkham (approved):

jakirkham:
Thanks for the ping Greg! 🙏

Missed the update

PR #774, opened 2024-09-05
Fix wheel tests for Rocky Linux 8.

bdice:
Recently RAPIDS added wheel tests on Rocky Linux 8. This requires a change to support installing dependencies with `yum` on Rocky Linux 8.

### DIFF
--- a/ci/test_wheel.sh
+++ b/ci/test_wheel.sh
@@ -11,8 +11,13 @@ python -m pip install $(echo ./dist/cucim*.whl)[test]
 
 CUDA_MAJOR_VERSION=${RAPIDS_CUDA_VERSION:0:2}
 
-DEBIAN_FRONTEND=noninteractive apt update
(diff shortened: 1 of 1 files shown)


bdice:
See failure here: https://github.com/rapidsai/cucim/actions/runs/10714414951/job/29708066437#step:8:227

```
ci/test_wheel.sh: line 14: apt: command not found
```

jameslamb (approved):

bdice:
@jameslamb I thought yesterday that I had missed something in this PR, but it turns out that we're fine. We need EPEL (Extra Packages for Enterprise Linux) to get openslide, but we're already enabling that in citestwheel: https://github.com/rapidsai/ci-imgs/blob/e0ffa9bae4a6cb998e594a1f9bc3916780516b21/citestwheel.Dockerfile#L47

So we're in the clear, and this PR was sufficient to make nightly tests pass.

jameslamb:
ah great, thanks for that!

Issue #775, opened 2024-09-06
[BUG] test_read_region_cuda_memleak failing (GPUtil does not support Python 3.12)


jameslamb:
**Describe the bug**

Python 3.12 support is rolling out across RAPIDS (https://github.com/rapidsai/build-planning/issues/40), and was added in this project in #773.

After merge, one nightly performance test failed like this:

```text
________________________ test_read_region_cuda_memleak _________________________
[gw6] linux -- Python 3.12.5 /pyenv/versions/3.12.5/bin/python
python/cucim/tests/performance/clara/test_read_region_memory_usage.py:25: in test_read_region_cuda_memleak
    import GPUtil
/pyenv/versions/3.12.5/lib/python3.12/site-packages/GPUtil/__init__.py:33: in <module>
    from .GPUtil import GPU, getGPUs, getAvailable, getAvailability, getFirstAvailable, showUtilization, __version__
/pyenv/versions/3.12.5/lib/python3.12/site-packages/GPUtil/GPUtil.py:34: in <module>
    from distutils import spawn
E   ModuleNotFoundError: No module named 'distutils'
```

([build link](https://github.com/rapidsai/cucim/actions/runs/10732925239/job/29765432400#step:8:529))

**Steps/Code to reproduce bug**

Run this test:

https://github.com/rapidsai/cucim/blob/1584885a0e6447ded28f78a59e3f303a25c7c0f6/python/cucim/tests/performance/clara/test_read_region_memory_usage.py#L24

**Expected behavior**

N/A

**Additional context**

`distutils` was removed from the Python standard library in Python 3.12: https://peps.python.org/pep-0632/

`GPUtil` has not been updated since August 2019 ([commit history](https://github.com/anderskm/gputil/commits/master/)), and it unconditionally runs `import distutils` ([code link](https://github.com/anderskm/gputil/blob/42ef071dfcb6469b7ab5ab824bde6ca9f6d0ab8a/GPUtil/GPUtil.py#L34)).

Options to fix this:

* remove that performance test
* refactor that performance test to not use `GPUtil`
* modify `GPUtil` to make it compatible with Python 3.12 (and either contribute that upstream or patch/vendor it here)

jameslamb:
I was wondering why, in the failing build linked above, the wheel tests for Python 3.12 were failing in this way but the conda tests were not.

Think I found it... `setuptools` provides `distutils`, and `setuptools` was making it into some but not all CI environments here.

```shell
docker run \
    --rm \
[...]

PR #776, opened 2024-09-06
add test-time dependency on setuptools

jameslamb:
## Summary

Contributes to #775, follow-up to #773.

Some nightly Python 3.12 tests are failing in this repo, as a result of the following:

* some tests here rely on `GPUtil`
* `GPUtil` was last updated 5+ years ago, and unconditionally imports `distutils` ([code link](https://github.com/anderskm/gputil/blob/42ef071dfcb6469b7ab5ab824bde6ca9f6d0ab8a/GPUtil/GPUtil.py#L34))
* `distutils` was removed from the Python standard library in Python 3.12

This proposes a **temporary fix**... adding a test-time dependency on `setuptools`. `import distutils` will work in an environment where `setuptools` is installed.

## Benefits of these changes

Fixes nightly tests, without imposing any new requirements of users of `cucim`.

## Notes for Reviewers

At some point, `setuptools` will likely remove `distutils` or the parts of it that `GPUtil` needs, and these tests will break in this way again. I've listed out some alternatives in #775, those should be explored at some point.

## How I tested this

On a machine with CUDA 12.2 and V100s, I can reproduce the test failures like this:

```shell
docker run \
    --rm \
    --gpus 1 \
    -v $(pwd):/opt/work \
    -w /opt/work \
    -it rapidsai/citestwheel:cuda12.5.1-ubuntu22.04-py3.12 \
    bash

python -m pip install \
    'cucim-cu12[test]==24.10.*,>=0.0.0a0'

python -m pytest \
    --numprocesses=8 \
    --dist=worksteal \
    ./python/cucim/tests/performance/clara
```

After running `pip install setuptools`, those tests pass 😁


dependencies.yaml
```diff
@@ -301,6 +301,7 @@ dependencies:
           - pytest-cov>=2.12.1
           - pytest-lazy-fixtures>=1.0.0
           - pytest-xdist
+          - setuptools>=60.0  # provides 'distutils', keeps 'GPUtil' working on Python 3.12+
```
jameslamb, line 304: `setuptools==60.0.0` is the first release where `disutils` was vendored. From https://setuptools.pypa.io/en/latest/deprecated/distutils-legacy.html

> *Since the 60.0.0 release, Setuptools includes a local, vendored copy of distutils (from late copies of CPython) that is enabled by default.*

That release is from December 2021 ([link](https://pypi.org/project/setuptools/60.0.0/)), so I'd be surprised if it causes any conflicts.

bdice:
We may be able to get rid of `GPUtil`. It's very outdated and we have other ways to get the information it provides.

It is only used in one test: https://github.com/rapidsai/cucim/blob/1584885a0e6447ded28f78a59e3f303a25c7c0f6/python/cucim/tests/performance/clara/test_read_region_memory_usage.py#L24

We can replace `GPUtil.get_GPUs()[0].memoryUsed` with this code from rmm:

```python
import rmm

def get_used_memory():
    free, total = rmm.mr.available_device_memory()
    memory_used = (total - free) / (2 ** 20)
    return memory_used
```

Or we can use cuda-python: https://nvidia.github.io/cuda-python/module/cudart.html#cuda.cudart.cudaMemGetInfo
```python
import cuda.cudart

def get_used_memory():
    status, free, total = cuda.cudart.cudaMemGetInfo()
    if status != cuda.cudart.cudaError_t.cudaSuccess:
        raise RuntimeError("Failed to get GPU memory info.")
    memory_used = (total - free) / (2 ** 20)
    return memory_used
```

Neither rmm or cuda-python are test dependencies at present, but I think it would be reasonable to use one of these stable, actively maintained packages instead of `GPUtil`. I lean towards the `cuda-python` solution because it's more minimal.

PR #777, opened 2024-09-06
Use cupy to measure memory leak

### LINKED ISSUE #775: [BUG] test_read_region_cuda_memleak failing (GPUtil does not support Python 3.12)

bdice:
While adding support for Python 3.12 in #773, we found a problem where `GPUtil` does not support Python 3.12 (#775).

This PR removes the dependency on `GPUtil` and replaces it with `cupy`, which is already a dependency.

Closes #775. Closes #776.

### DIFF
--- a/dependencies.yaml
+++ b/dependencies.yaml
@@ -295,7 +295,6 @@ dependencies:
     common:
       - output_types: [conda, requirements, pyproject]
         packages:
-          - GPUtil>=1.4.0
           - psutil>=5.8.0
           - pytest>=6.2.4,<8.0.0a0
           - pytest-cov>=2.12.1
@@ -323,3 +322,13 @@ dependencies:
         packages:
           # Already added to requirements via docs. This is for tests.
           - numpydoc>=1.5
+    specific:
+      - output_types: [conda, requirements, pyproject]
+        matrices:
+          - matrix: {cuda: "12.*"}
+            packages:
+              - cuda-python>=12.0,<13.0a0
+          - matrix: {cuda: "11.*"}
+            packages: &test_cuda_python_cu11
+              - cuda-python>=11.7.1,<12.0a0
+          - {matrix: null, packages: *test_cuda_python_cu11}
--- a/dependencies.yaml
+++ b/dependencies.yaml
@@ -322,13 +322,3 @@ dependencies:
         packages:
           # Already added to requirements via docs. This is for tests.
           - numpydoc>=1.5
-    specific:
-      - output_types: [conda, requirements, pyproject]
-        matrices:
-          - matrix: {cuda: "12.*"}
-            packages:
-              - cuda-python>=12.0,<13.0a0
-          - matrix: {cuda: "11.*"}
-            packages: &test_cuda_python_cu11
-              - cuda-python>=11.7.1,<12.0a0
-          - {matrix: null, packages: *test_cuda_python_cu11}
--- a/python/cucim/tests/performance/clara/test_read_region_memory_usage.py
+++ b/python/cucim/tests/performance/clara/test_read_region_memory_usage.py
@@ -23,7 +23,8 @@ pytest.importorskip("imagecodecs")
 
 
 def test_read_region_cuda_memleak(testimg_tiff_stripe_4096x4096_256_jpeg):
-    def get_used_gpu_memory():
+    def get_used_gpu_memory_mib():
+        """Get the used GPU memory in MiB."""
         status, free, total = cuda.cudart.cudaMemGetInfo()
         if status != cuda.cudart.cudaError_t.cudaSuccess:
             raise RuntimeError("Failed to get GPU memory info.")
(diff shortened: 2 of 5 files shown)


jameslamb (approved):
I'm much happier with this change than #776. Thank you!

dependencies.yaml
```diff
@@ -323,3 +322,13 @@ dependencies:
...
           # Already added to requirements via docs. This is for tests.
           - numpydoc>=1.5
+    specific:
+      - output_types: [conda, requirements, pyproject]
+        matrices:
+          - matrix: {cuda: "12.*"}
+            packages:
+              - cuda-python>=12.0,<13.0a0
+          - matrix: {cuda: "11.*"}
+            packages: &test_cuda_python_cu11
```
grlee77, line 332: out of curiosity, what does this `&test_cuda_python_cu11` notation do?

python/cucim/tests/performance/clara/test_read_region_memory_usage.py
```diff
@@ -22,23 +23,25 @@
...
+    def get_used_gpu_memory_mib():
+        """Get the used GPU memory in MiB."""
+        status, free, total = cuda.cudart.cudaMemGetInfo()
+        if status != cuda.cudart.cudaError_t.cudaSuccess:
+            raise RuntimeError("Failed to get GPU memory info.")
+        memory_used = (total - free) / (2**20)
+        return memory_used
+
+    status, num_gpus = cuda.cudart.cudaGetDeviceCount()
+    if status != cuda.cudart.cudaError_t.cudaSuccess or num_gpus == 0:
```
grlee77, line 35: We can also potentially just use CuPy for this and avoid having to add the cuda-python dependency:

add up top:
```py
import cupy as cp
```

then can use here:
```py
    def get_used_gpu_memory_mib():
        """Get the used GPU memory in MiB."""
        dev = cp.cuda.Device()
        free, total = dev.mem_info
        memory_used = (total - free) / (2**20)
        return memory_used

    num_gpus = cp.cuda.runtime.getDeviceCount()
    if num_gpus == 0:
```

grlee77 (changes requested):
Agree that using cuda-python is preferable to GPUtil. We can also get this info from CuPy without adding the dependency, though (see suggestion). Let me know what you think.

dependencies.yaml
bdice: This is a YAML anchor. We use this to define a list that we re-use in `*test_cuda_python_cu11`. We need a "fallback" list when the `cuda` selector is not defined.

python/cucim/tests/performance/clara/test_read_region_memory_usage.py
bdice: Great. I'll implement that and remove the cuda-python dependency.

jameslamb (approved):
re-approving, awesome that the net result here is -1 dependencies 😁

grlee77 (approved):
Looks great, thanks!

gigony:
Thanks @bdice for fixing the issue!

gigony (approved):

PR #782, opened 2024-09-23
Fix a couple of performance issues in `peak_local_max` (improve performance of blob detectors and `corner_peaks`)

grlee77:
It seems that we can get a large speedup in calls to `peak_local_max` when the default `min_distance=1` is specified. A bottleneck in that function can be the call to `ensure_spacing` which falls back to the CPU, but we don't need to call that function at all unless `min_distance>1` (no coordinates can be excluded if min_distance=1, so it is pointless to call it).

I proposed the same changes upstream in https://github.com/scikit-image/scikit-image/pull/7548. See additional comments and measurements made there.

There should be no change in behavior with this MR, it is purely a performance improvement.

### DIFF
--- a/python/cucim/src/cucim/skimage/_shared/coord.py
+++ b/python/cucim/src/cucim/skimage/_shared/coord.py
@@ -101,6 +101,8 @@ def ensure_spacing(
         A subset of coord where a minimum spacing is guaranteed.
 
     """
+    if spacing == 1:
+        return coords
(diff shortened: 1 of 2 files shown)


gigony (approved):
Thanks @grlee77 !

It looks great to me, and it's good that we can optimize the performance for cases where `min_distance` is 1!

Issue #763, opened 2024-08-20
[FEA] PyPI package (wheel) release for the cucim.skimage module on non-Linux platforms (Windows, etc.)


gigony:
**Is your feature request related to a problem? Please describe.**

There is an increasing number of requests from users who want to use the `cucim.skimage` module (which depends on CuPy) on other platforms (non-Linux).

- https://github.com/rapidsai/cucim/issues/748
- https://github.com/rapidsai/cucim/issues/454
  - https://github.com/rapidsai/cucim/issues/559
  - https://github.com/rapidsai/cucim/issues/441
  - https://github.com/rapidsai/cucim/issues/270
  - https://github.com/rapidsai/cucim/issues/86

It would be great if we could release the `cucim.skimage` portion of the package (e.g., cucim_cu12-24.8.0-py3-none-win_amd64.whl) in the PyPI registry without including the unnecessary/non-supported IO portion (shared libraries) of cuCIM.
- [cucim-cu11 · PyPI](https://pypi.org/project/cucim-cu11/#files)
- [cucim-cu12 · PyPI](https://pypi.org/project/cucim-cu12/#files)

**Describe the solution you'd like**

I believe there is a straightforward way to build and register a Windows Python wheel by updating [cucim/ci/build_wheel.sh at branch-24.10 · rapidsai/cucim](https://github.com/rapidsai/cucim/blob/branch-24.10/ci/build_wheel.sh) and register to Windows PyPI package along with the original wheel file (executing this only for a specific build configuration).

```sh
# The following will create cucim_cu12-24.6.0-py3-none-win_amd64.whl file.
wheel tags --python-tag=py3 --abi-tag=none --platform-tag=win_amd64 cucim_cu12-24.6.0-cp311-cp311-manylinux_2_28_aarch64.whl
```

jakirkham:
We discussed this in the cuCIM meeting today after feedback from RAPIDS OPS. Based on this we revised our plan, which is sketched below

![Image]( https://github.com/user-attachments/assets/2e27d4f1-7615-4e34-86e3-ee64693b2c9b )

This would split out the Python/C++ bits into `pylibcucim` and the pure Python bits into `pycucim`. The `cucim` package would become a meta package that pulls both in where supported. If the `pylibcucim` is not supported on a platform (like Windows currently), we would select it out. So users would still install `cucim`. It would just now work on other platforms not supported today (like Windows) by pulling in the pure Python bits only 

The code for `cucim-cuXY` would look like this 

[...]

Issue #645, opened 2023-11-29
Evaluate using `scikit-build`


jakirkham:
Currently cuCIM contains

* C++ library
* Python/C++ bindings using pybind11
* Pure Python code

When building wheels, the first two are built together as part of a CMake build. The result of which is bundled into a Python wheel with the last component, which is then shipped

The result is wheels are treated as pure, which caused some issues ( https://github.com/rapidsai/cucim/issues/641 ). Though those can be worked around

That said, it may be worth using scikit-build to streamline all of this and create a single wheel

One consideration is some users just use the Pure Python portion. Maybe we can add a flag to enable/disable the C++ build

jakirkham:
More history on build changes made in issue: https://github.com/rapidsai/cucim/issues/305

thewtex:
:+1: !

A few notes: scikit-build's core has been refactored / rebuilt into [scikit-build-core](https://github.com/scikit-build/scikit-build-core), which is a PEP517 build-backend that can be added to pyproject.toml.

There is outstanding pybind11 support.

Support for [optional built components with pure python fallback is planned](https://github.com/scikit-build/scikit-build-core/issues/112).

CC @henryiii @jcfr

thewtex:
Side note: the current pip install command is failing for me:

```
❯ pip install cucim-cu12
Collecting cucim-cu12
  Using cached cucim-cu12-24.2.0.tar.gz (6.9 kB)
  Preparing metadata (setup.py) ... error
  error: subprocess-exited-with-error
  
  × python setup.py egg_info did not run successfully.
  │ exit code: 1
  ╰─> [6 lines of output]
      Traceback (most recent call last):
        File "<string>", line 2, in <module>
        File "<pip-setuptools-caller>", line 34, in <module>
        File "/tmp/pip-install-65f7ajw3/cucim-cu12_b8221968d94c4da495894b499ea01754/setup.py", line 90, in <module>
          raise RuntimeError("Bad params")
      RuntimeError: Bad params
      [end of output]
  
  note: This error originates from a subprocess, and is likely not a problem with pip.
error: metadata-generation-failed

× Encountered error while generating package metadata.
╰─> See above for output.

note: This is an issue with the package mentioned above, not pip.
```

```
❯ pip --version
pip 24.0 from /home/matt/bin/micromamba/envs/itk-wasm/lib/python3.11/site-packages/pip (python 3.11)
```

jakirkham:
What Python version are you using?

[...]

PR #792, opened 2024-10-15
build wheels without build isolation, print sccache stats in builds

jameslamb:
## Description

Contributes to https://github.com/rapidsai/build-planning/issues/108
Contributes to https://github.com/rapidsai/build-planning/issues/111

Proposes some small packaging/CI changes, matching similar changes being made across RAPIDS.

* removes the use of build isolation in wheel builds
* printing `sccache` stats to CI logs
* reducing `pip`'s verbosity in wheel building scripts
* using the RAPIDS conventions `py_build_{project}` and `py_rapids_build_{project}` in `dependencies.yaml`
* updating to the latest `rapids-dependency-file-generator` (v1.16.0)

### DIFF
--- a/.github/workflows/pr.yaml
+++ b/.github/workflows/pr.yaml
@@ -16,7 +16,6 @@ jobs:
     secrets: inherit
     uses: rapidsai/shared-workflows/.github/workflows/pr-builder.yaml@branch-24.12
   wheel-build:
-    needs: checks
     secrets: inherit
     uses: rapidsai/shared-workflows/.github/workflows/wheels-build.yaml@branch-24.12
     with:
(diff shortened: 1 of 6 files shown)


jakirkham:
One thing we would like to do is move to scikit-build-core: https://github.com/rapidsai/cucim/issues/645

Though that hasn't happened yet

We discussed some package restructuring ( https://github.com/rapidsai/cucim/issues/763 ), which would unblock scikit-build-core

Imagine things get a lot simpler to manage here if we are able to make those changes

jameslamb:
> *Imagine things get a lot simpler to manage here if we are able to make those changes*

"things" in general, definitely. Those look like good changes that should be made in this project.

But to be clear... the work described in those issues you linked wouldn't materially change the way this PR looks. For example, `cuspatial` projects already use `scikit-build-core` and have the wheels split into C++ vs. Python, and the changeset in that repo for this effort was pretty similar: https://github.com/rapidsai/cuspatial/pull/1473

bdice (approved):

PR #806, opened 2024-12-16
Incorporate upstream changes from scikit-image 0.25

grlee77:
This MR updates for consistency with scikit-image v0.25 (released on Dec 13, 2024).

The main API change is deprecating footprint functions `square`, `rectangle` and `cube` in favor of a general `footprint_rectangle` method. 

The supported version range of various dependencies have also been bumped.


python/cucim/src/cucim/skimage/filters/_gabor.py
```diff
@@ -79,20 +79,19 @@ def gabor_kernel(
...
+    >>> ax.imshow(cp.asnumpy(gk.real))  # doctest: +SKIP
+    >>> plt.show()                      # doctest: +SKIP
 
     >>> # more ripples (equivalent to increasing the size of the
     >>> # Gaussian spread)
     >>> gk = gabor_kernel(frequency=0.2, bandwidth=0.1)
-    >>> plt.figure()                    # doctest: +SKIP
-    >>> io.imshow(cp.asnumpy(gk.real))  # doctest: +SKIP
-    >>> io.show()                       # doctest: +SKIP
+    >>> fig, ax = plt.suplots()         # doctest: +SKIP
```
gigony, line 92:
```suggestion
    >>> fig, ax = plt.subplots()         # doctest: +SKIP
```

python/cucim/src/cucim/skimage/morphology/tests/test_footprints.py
```diff
@@ -250,3 +230,49 @@ def test_ball_series_approximation_unavailable():
...
+
+
+def assert_decomposition_equal(actual, desired):
+    assert len(actual) == len(desired)
+    for a, d in zip(actual, desired):
+        assert_array_equal(a[0], d[0])
+        assert a[1] == d[1]
+
+
+class Test_footprint_rectangule:
```
gigony, line 242:
```suggestion
class Test_footprint_rectangle:
```

python/cucim/src/cucim/skimage/filters/_gabor.py
```diff
@@ -209,21 +208,21 @@ def gabor(
...
+    >>> from skimage import data
     >>> from matplotlib import pyplot as plt  # doctest: +SKIP
 
     >>> image = cp.array(data.coins())
     >>> # detecting edges in a coin image
     >>> filt_real, filt_imag = gabor(image, frequency=0.6)
-    >>> plt.figure()                        # doctest: +SKIP
-    >>> io.imshow(cp.asnumpy(filt_real))    # doctest: +SKIP
-    >>> io.show()                           # doctest: +SKIP
+    >>> fix, ax = plt.subplots()          # doctest: +SKIP
```
gigony, line 217:
```suggestion
    >>> fig, ax = plt.subplots()          # doctest: +SKIP
```

gigony (approved):
Thanks @grlee77 for the change!

Besides some typos, it looks good to me!

dependencies.yaml
```diff
@@ -258,10 +258,10 @@ dependencies:
       - output_types: [conda, requirements, pyproject]
         packages:
           - click
-          - lazy_loader>=0.1
-          - numpy>=1.23.4,<3.0a0
-          - scikit-image>=0.19.0,<0.25.0a0
-          - scipy>=1.6.0
+          - lazy-loader>=0.4
```
jakirkham, line 261: Since `-` and `_` are distinct in conda and not in pip, we should double check this naming is matched by the conda package

Otherwise we may want to split conda and pip into two different sections

bdice: `lazy-loader` is the new name, and this should be correct for both PyPI and conda-forge.

See: https://github.com/conda-forge/lazy_loader-feedstock/pull/11

Also https://pypi.org/project/lazy_loader/ redirects to https://pypi.org/project/lazy-loader/

dependencies.yaml
```diff
@@ -258,10 +258,10 @@ dependencies:
...
+          - numpy>=1.24,<3.0a0
```
bdice, line 262: Is there a reason for the new lower bound on `numpy`? We have `1.23.4` as the lower bound across all of RAPIDS right now. If things are still compatible with 1.23.4, we should continue using that for now.

bdice: I see scikit-image updated their pinning upstream. This is fine, it can be left as-is.

jakirkham (approved):

dependencies.yaml
grlee77: Let me put is back to 1.23.4 in that case. It is only for the most recent release of scikit-image that a newer version would need to be pulled in. Older numpy should work for scikit-image<=0.24

conda/environments/all_cuda-125_arch-x86_64.yaml
```diff
@@ -17,14 +17,14 @@ dependencies:
...
+- lazy-loader>=0.4
 - libcufile-dev
 - libnvjpeg-dev
 - libnvjpeg-static
-- matplotlib-base
+- matplotlib-base>=3.7
 - nbsphinx
 - ninja
-- numpy>=1.23.4,<3.0a0
+- numpy>=1.24,<3.0a0
```
grlee77, line 27:
```suggestion
- numpy>=1.23.4,<3.0a0
```

dependencies.yaml
grlee77, line 262:
```suggestion
          - numpy>=1.23.4,<3.0a0
```

conda/environments/all_cuda-118_arch-x86_64.yaml
```diff
@@ -16,15 +16,15 @@ dependencies:
...
 - libcufile-dev=1.4.0.31
 - libcufile=1.4.0.31
 - libnvjpeg-dev=11.6.0.55
 - libnvjpeg=11.6.0.55
-- matplotlib-base
+- matplotlib-base>=3.7
 - nbsphinx
 - ninja
-- numpy>=1.23.4,<3.0a0
+- numpy>=1.24,<3.0a0
```
grlee77, line 27:
```suggestion
- numpy>=1.23.4,<3.0a0
```

python/cucim/pyproject.toml
```diff
@@ -26,10 +26,10 @@ requires-python = ">=3.10"
 dependencies = [
     "click",
     "cupy-cuda11x>=12.0.0",
-    "lazy_loader>=0.1",
-    "numpy>=1.23.4,<3.0a0",
-    "scikit-image>=0.19.0,<0.25.0a0",
-    "scipy>=1.6.0",
+    "lazy-loader>=0.4",
+    "numpy>=1.24,<3.0a0",
```
grlee77, line 30:
```suggestion
    "numpy>=1.23.4,<3.0a0",
```

PR #808, opened 2025-01-09
CuPy 14.0 compatibility

grlee77:
The upcoming CuPy 14.0 release has some breaking change in dtype handling, etc. due to changes made to mirror NumPy 2.0 behavior. 

Fortunately most of cuCIM was still working as is when I tested it locally. The changes made here are what was required for all tests to pass when using the current `main` branch of CuPy (14.0 is not yet released)


gigony (approved):
It looks good to me. Thanks @grlee77 !

jakirkham:
Thanks Greg and Gigon! 🙏

PR #809, opened 2025-01-09
Improve performance of color distance calculations by kernel fusion

grlee77:
## Description

This MR is purely to fuse many separate kernels into a single elementwise kernel for each of the "color distance" functions. This is expected to result in substantial performance improvement by reduced kernel launch overhead and because a single pass through the image is much more memory efficient than many separate kernel calls.

There is not expected to be any change in behavior (existing tests must continue to pass)

## Benchmark Results

I added benchmarks for these functions and compared the results before and after this change.

The acceleration values are the relative speedup as compared to the scikit-image implementation

### For a pair of 32-bit LAB images of shape: (512, 512, 3)

function         | acceleration (old) | acceleration (new)
-----------------|--------------------|-------------------
deltaE_cie76     |   5.49             |   21.85
deltaE_ciede94   |   37.60            |   65.09
deltaE_ciede2000 |   12.54            |  109.71
deltaE_cmc       |   27.18            |   89.23

### For a pair of 32-bit LAB images of shape: (3840, 2160, 3)

function         | acceleration (old) | acceleration (new)
-----------------|--------------------|-------------------
deltaE_cie76     |   31.12            |  250.74
deltaE_ciede94   |   79.15            |  117.50
deltaE_ciede2000 |   71.36            |  132.73
deltaE_cmc       |   70.65            |  154.83

### DIFF
--- a/benchmarks/skimage/cucim_color_bench.py
+++ b/benchmarks/skimage/cucim_color_bench.py
@@ -1,4 +1,5 @@
 import argparse
+import math
 import os
 import pickle
 
@@ -25,6 +26,10 @@ func_name_choices = [
     "lab2xyz",
     "rgba2rgb",
     "label2rgb",
+    "deltaE_cie76",
+    "deltaE_ciede94",
+    "deltaE_ciede2000",
+    "deltaE_cmc",
 ]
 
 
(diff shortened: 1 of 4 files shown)


grlee77:
Not sure that there is a common use case for actually running these on full images or if it is most commonly used just to compare the perceptual difference of single pairs of colors. If the later, then the CPU will be better for running that case. 

I think perhaps there could be use cases like comparing perceptual differences among all pairs of colors in a categorical colormap, etc.

PR #814, opened 2025-01-22
Add shellcheck to pre-commit and fix warnings

gforsyth:
## Description                                                                                                
                                                                                                              
`shellcheck` is a fast, static analysis tool for shell scripts. It's good at
flagging up unused variables, unintentional glob expansions, and other potential
execution and security headaches that arise from the wonders of `bash` (and
other shlangs).                   
                                                                                                              
This PR adds a `pre-commit` hook to run `shellcheck` on all of the `sh-lang`
files in the `ci/` directory, and the changes requested by `shellcheck` to make
the existing files pass the check.                              
                                                                                                              
xref: rapidsai/build-planning#135

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -37,3 +37,9 @@ repos:
     hooks:
         - id: rapids-dependency-file-generator
           args: ["--clean"]
+  - repo: https://github.com/shellcheck-py/shellcheck-py
+    rev: v0.10.0.1
+    hooks:
+      - id: shellcheck
+        args: ["--severity=warning"]
+        files: ^ci/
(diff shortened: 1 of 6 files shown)


jameslamb (approved):
I'm good with this! Just double-check that `update-version.sh` change if you haven't already, I think that `sed_runner` thing can be really sensitive to quoting.

You can test like this, from the root of the repo:

```shell
# get all the tags (assumed you have rapidsai/cucim as the remote)
git fetch upstream --tags

# update versions
ci/release/update-version.sh '25.04.00'
```

If that worked, you should see most "25.02" changed to "25.04" and all "25.2" changed to "25.4".

gforsyth:
> If that worked, you should see most "25.02" changed to "25.04" and all "25.2" changed to "25.4".

Yep, that was definitely one of the gnarlier quoting conversions, but yes, it works!

PR #822, opened 2025-02-04
add telemetry

msarahan:
Enables telemetry during cucim CI runs. This is done by parsing GitHub Actions run log metadata and should have no impact on build or test times.

xref https://github.com/rapidsai/build-infra/issues/139


.github/workflows/pr.yaml
```diff
@@ -19,33 +19,48 @@ jobs:
...
     with:
       build_type: pull-request
   docs-build:
     needs: conda-python-build
     secrets: inherit
-    uses: rapidsai/shared-workflows/.github/workflows/custom-job.yaml@branch-25.02
+    uses: rapidsai/shared-workflows/.github/workflows/custom-job.yaml@nvks-runners
     with:
       build_type: pull-request
       node_type: "gpu-v100-latest-1"
```
gforsyth, line 66:
```suggestion
      node_type: "gpu-l4-latest-1"
```

gforsyth (approved):
I think the node types need to be changed if we're requesting a specific node?  Otherwise this looks good!

PR #836, opened 2025-02-16
update vendored binary morphology code

grlee77:
Incorporate a couple of fixes/improvements proposed upstream
-  (cupy/cupy#8954)
-  (cupy/cupy#8956)


python/cucim/src/cucim/skimage/_vendored/_ndimage_morphology.py
```diff
@@ -480,6 +481,7 @@ def binary_dilation(
 
     .. seealso:: :func:`scipy.ndimage.binary_dilation`
     """
+    axes = _util._check_axes(axes, input.ndim)
     structure, structure_shape, symmetric = _prep_structure(
         structure, input.ndim
```
gigony, line 486: I wonder why this statement doesn't use `len(axes)` while others are using it.
```suggestion
        structure, len(axes)
```

gigony (commented):
Thanks @grlee77  for the update!
It looks good to me! I left one comment on this.

PR #837, opened 2025-02-16
vendor CUDA-accelerated `find objects`

grlee77:
Include a copy of `cupyx.scipy.ndimage.find_objects` that was contributed upstream.

A couple of cuCIM functions that used `find_objects` are also updated to use this implementation instead of copying to host to call the SciPy version.

See: https://github.com/cupy/cupy/pull/8916

The version vendored here is modified to exclude the Cython-based slice list comparison as we don't support using Cython code from cuCIM. This version will still provide good acceleration and we preferentially import the version from CuPy instead if it is available.

### DIFF
--- a/python/cucim/src/cucim/skimage/_vendored/_ndimage_measurements.py
+++ b/python/cucim/src/cucim/skimage/_vendored/_ndimage_measurements.py
@@ -1,0 +1,212 @@
+import math
+
+import cupy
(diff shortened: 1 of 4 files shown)


gigony (approved):
Thanks @grlee77 for adding accelerated find_objects!
It looks good to me!

PR #838, opened 2025-02-16
add utility for memory efficient maximum pairwise distance computation with GPU support

grlee77:
The maximum Feret diameter computations compute pairwise distances between all points which can lead to out of memory errors if the number of points on the object boundary is large. (The memory used is quadratic in the number of points).

This MR implements a block-wise version that retains efficiency, but has much lower memory requirements. It also handles checking for optional GPU acceleration via CuPy's optional cuVS/pylibraft dependencies. CPU fallback is done if those are not available.


python/cucim/src/cucim/skimage/_shared/distance.py
```diff
@@ -0,0 +1,177 @@
...
+      └────────┴────────┴────────┘
+
+    where due to the symmetry of the pdist matrix, we don't compute the lower
+    triangular blocks.
+    """
+
+    if _distance_on_cpu:
+        warnings.warn(
+            "cuVS >= 25.02 or pylibraft < 24.12 must be installed to use "
+            "GPU-accelerated pairwaise distance computations. Falling back "
```
gigony, line 100:
```suggestion
            "GPU-accelerated pairwise distance computations. Falling back "
```

python/cucim/src/cucim/skimage/_shared/distance.py
```diff
@@ -0,0 +1,177 @@
...
+    coords : np.ndarray (num_points, ndim)
+        The coordinates to process.
+    metric : str, optional
+        Can be any metric supported by `scipy.spatial.distance.cdist`. The
+        default is the squared Euclidean distance (sqeuclidean).
+    coords_per_block : bool, optional
+        Internally, calls to cdist will be made with subsets of coords where
+        the subset size is (coords_per_block, ndim).
+    compute_argmax : bool, optional
+        If True, the value of the coordate indices corresponding to the maxima
```
gigony, line 57:
```suggestion
        If True, the value of the cooridate indices corresponding to the maxima
```

python/cucim/src/cucim/skimage/_shared/distance.py
```diff
@@ -0,0 +1,177 @@
...
+    compute_argmax=False,
+    cdist_kwargs={},
+):
+    """Find maximum pointwise distance.
+
+    Computes by processing blocks of coordinates to reduce overall memory
+    requirement. The memory used at runtime will be proportional to
+    ``coords_per_block**2``.
+
+    A block size of >= 2000 is recommended to overhead poor GPU resource usage
```
gigony, line 43:
```suggestion
    A block size of >= 2000 is recommended to avoid poor GPU resource usage
```

python/cucim/src/cucim/skimage/_shared/distance.py
```diff
@@ -0,0 +1,177 @@
...
+                distance.cdist(
+                    coords_block1,
+                    coords_block2,
+                    metric=metric,
+                    out=temp,
+                    **cdist_kwargs,
+                )
+                current_output = temp
+            else:
+                # omit out= for the last block as size may be
```
gigony, line 154: This comment sentence doesn't seem to be complete.

python/cucim/src/cucim/skimage/_shared/distance.py
```diff
@@ -0,0 +1,177 @@
...
+    requirement. The memory used at runtime will be proportional to
+    ``coords_per_block**2``.
+
+    A block size of >= 2000 is recommended to overhead poor GPU resource usage
+    and to reduce kernel launch overhead.
+
+    Parameters
+    ----------
+    coords : np.ndarray (num_points, ndim)
+        The coordinates to process.
```
gigony, line 49: The code converts input coordinates to float32 by default, which could lead to precision loss. I think it would be good to note that this `coords` would be converted to `float32` data type.

python/cucim/src/cucim/skimage/_shared/distance.py
```diff
@@ -0,0 +1,177 @@
...
+            "GPU-accelerated pairwaise distance computations. Falling back "
+            "to SciPy-based CPU implementation."
+        )
+        xp = np
+        coords = cp.asnumpy(coords)
+    else:
+        xp = cp
+        coords = cp.asarray(coords)
+
+    num_coords, _ = coords.shape
```
gigony, line 109: Input validation might be needed here.

```suggestion

    if not isinstance(coords, (np.ndarray, cp.ndarray)):
        raise TypeError("coords must be a numpy or cupy array")
    
    if coords.ndim != 2:
        raise ValueError(
            f"coords must be a 2-dimensional array, got shape {coords.shape}"
        )
    
    num_coords, _ = coords.shape
```

python/cucim/src/cucim/skimage/_shared/distance.py
```diff
@@ -0,0 +1,177 @@
...
+    """
+
+    if _distance_on_cpu:
+        warnings.warn(
+            "cuVS >= 25.02 or pylibraft < 24.12 must be installed to use "
+            "GPU-accelerated pairwaise distance computations. Falling back "
+            "to SciPy-based CPU implementation."
+        )
+        xp = np
+        coords = cp.asnumpy(coords)
```
gigony, line 104: We don't need to use numpy for here?

```suggestion
        coords = np.asnumpy(coords)
```

python/cucim/src/cucim/skimage/_shared/distance.py
```diff
@@ -0,0 +1,177 @@
...
+    Computes by processing blocks of coordinates to reduce overall memory
+    requirement. The memory used at runtime will be proportional to
+    ``coords_per_block**2``.
+
+    A block size of >= 2000 is recommended to overhead poor GPU resource usage
+    and to reduce kernel launch overhead.
+
+    Parameters
+    ----------
+    coords : np.ndarray (num_points, ndim)
```
gigony, line 48:
```suggestion
    coords : numpy.ndarray or cupy.ndarray of shape (num_points, ndim)
```

gigony (commented):
Thanks @grlee77 for this update! I left some minor comments on this.

python/cucim/src/cucim/skimage/_shared/distance.py
grlee77: There is no `np.asnumpy`. For CuPy, if coords is already a NumPy array the following would pass:
```py
assert cp.asnumpy(coords) is coords
```
but if `coords` was a CuPy array it would be copied to a new host NumPy array.

python/cucim/src/cucim/skimage/_shared/distance.py
grlee77: Thanks, it is good to check.

I put this function in the private `_shared` module to indicate it is not part of the public API and only intended for internal use.

python/cucim/src/cucim/skimage/_shared/distance.py
grlee77: right, it is supposed to say "size may be different". I will update it

python/cucim/src/cucim/skimage/_shared/distance.py
grlee77: I added a note that distances are always computed in 32-bit precision. Currently this utility is only used internally for the "feret_distance_max" region property. Given that the object is already discretized to pixels, the distance measurements based on pixel coordinates will already be a pretty rough measurement that is only accurate to ~1 pixel spacing, so high precision should not be needed.

gigony (approved):
Thanks @grlee77 !
It looks good to me!

PR #839, opened 2025-02-17
Improve performance of `label2rgb`

grlee77:
This MR substantially improves the performance of `label2rgb` by using custom elementwise kernels.

A few fixes in the benchmark scripts were also made
* use int32 dtype for the labels during benchmarks
* Fix label2rgb benchmark bug where 3rd dimension was truncated.
* Use n-d blobs image as a more realistic label image.

Accelerations vs skimage before vs. after this change (when `image` is used it has `dtype=np.float32`)

shape (no image, mode='overlay')   | previous | new
-----------------------------------|----------|---------
(512, 512)                         |   7.6    |   27.0
(3840, 2160)                       |  25.4    |   35.9
(192, 192, 192)                    |  76.1    |  365.6

shape (with image, mode='overlay') | previous | new
-----------------------------------|----------|---------
(512, 512)                         |   51.7   |   110.6
(3840, 2160)                       |  262.5   |   309.8
(192, 192, 192)                    |  693.1   |  1334.7

shape (with image, mode='avg')     | previous | new
-----------------------------------|----------|---------
(512, 512)                         |   2.5    |   322.5
(3840, 2160)                       |  66.7    |  1330.6
(192, 192, 192)                    |  63.7    |  1374.0


gigony (commented):
Thanks @grlee77 for improving the `label2rgb` performance!
It looks good to me but I found a typo which may lead to the incorrect result.

gigony (approved):

PR #842, opened 2025-03-03
update vendored binary_fill_holes

grlee77:
updated `binary_fill_holes` to:
- allow fast code path to be used when `structure` is provided by the user
- fix handling of case like `output=bool` where a dtype instead of array is provided

These are also in the upstream CuPy MR that is still in review, but were needed for regionprops work being submitted here this week.


gigony (approved):
Thanks @grlee77  for this change!
It looks good to me!

PR #844, opened 2025-03-03
Region Properties Performance Overhaul - Part 2: Intensity Image Properties

grlee77:

## Overview

This MR implements the following intensity-based region properties
- intensity_mean
- intensity_std
- intensity_min
- intensity_max

regionprops_intensity_vs_size

regionprops_3d_intensity_vs_size

Note that "multi-intensity" is the time to compute the following list of region properties rather than just a single property
```python
properties = ["intensity_mean", "intensity_std", "intensity_min", "intensity_max"]
```

regionprops_intensity_vs_object_size

regionprops_3d_intensity_vs_object_size

### Benchmark conclusions

**Note:** The results for the older GPU-based regionprops from cuCIM are not shown here. However, for many properties that implementation became much slower as the number of regions increased. We can see that for the new implementation proposed here, performance is roughly independent of the number of regions present. As expected, the relative acceleration that can be achieved becomes larger as image sizes increase, eventually starting to plateau at the larger sizes shown.


jameslamb (approved):
`packaging-codeowners` was just added because you changed the `codespell` config, approving this on that basis.

Will defer to other reviewers on the substance of this PR.

gigony (approved):
Thanks @grlee77 !
It looks good to me!

PR #828, opened 2025-02-07
Implement `convex_hull_image` and `convex_hull_object`

grlee77:
This MR implements a hybrid CPU + GPU implementation of `convex_hull_image` and `convex_hull_object`. These functions can be used standalone, but `convex_hull_image` is also used during computation of some regionprops: 
- `area_convex`
- `image_convex`
- `feret_diameter_max`
- `solidity`

The implementation here still uses QHull via SciPy on the CPU, but all preprocessing computations as well as the post-processing of which image pixels fall within the convex hull polygon are done on the GPU. Fortunately, QHull was not the bottleneck, as seen in the benchmark results below.

The image below shows an example of a non-convex labeled region (yellow) with the convex region computed by `convex_hull_image` shown in red
[
convex_hull_image
](url)

## Benchmark results

The last two columns labeled "accel" are the ratio of (`cpu_duration / gpu_duration`). "32" indicates the GPU case was run with an extra kwarg that allows 32-bit floating point when computing the final convex image mask from the convex hull coordinates. All durations are in seconds.

All except the smallest sizes tested show performance benefit relative to scikit-image. The relative performance benefit can be very large, especially for large images and in cases where the image is highly concave and if `offset_coords=False` as this reduces the run time of the QHull algorithm on the CPU.

| shape | im_type | +offset coords | dur. (CPU) | dur. (GPU64) | dur. (GPU32) | accel. 64 | accel. 32 |
|-------|---------|----------------|------------|--------------|--------------|-----------|-----------|
| (64, 64) | skeleton | False | 0.00028 | 0.00040 | 0.00041 | 0.70 | 0.70 |
| (64, 64) | skeleton | True | 0.00042 | 0.00054 | 0.00054 | 0.78 | 0.78 |
| (64, 64) | convex | False | 0.00034 | 0.00046 | 0.00046 | 0.74 | 0.74 |
| (64, 64) | convex | True | 0.00052 | 0.00063 | 0.00063 | 0.82 | 0.82 |
| (128, 128) | skeleton | False | 0.00068 | 0.00083 | 0.00083 | 0.83 | 0.82 |
| (128, 128) | skeleton | True | 0.00097 | 0.00110 | 0.00111 | 0.88 | 0.87 |
| (128, 128) | convex | False | 0.00084 | 0.00096 | 0.00096 | 0.87 | 0.87 |
| (128, 128) | convex | True | 0.00115 | 0.00129 | 0.00130 | 0.89 | 0.88 |
[...]

PR #845, opened 2025-03-03
Region Properties Performance Overhaul - Part 3: Convex Image Properties

grlee77:

## Overview

This MR implements the following properties based on first computing a convex object for each region.
- area_convex
- solidity
- feret_diameter_max

## Benchmarks

Note that unlike in Part1 and Part2, computing the convex hull image via `convex_hull_image` introduced in #828, can only operate on a single labeled region at a time. Thus, these properties are expected to show good acceleration in the case of relatively few, larger labeled regions but will become slow when there are a very large number of small regions.

regionprops_convex_vs_size

regionprops_3d_convex_vs_size

Note that scikit-image results were not computed at larger 3D size due to excessive computation time

regionprops_convex_vs_object_size

Due to very slow performance of scikit-image's convex hull on large 3D regions, no comparison is shown here for that case.

### Benchmark conclusions

We can see that there is substantial acceleration across image sizes when the number of labeled regions is small (top 2 plots). However, because we have to launch the hybrid convex hull computation for each region in isolation, performance suffers at fixed image size as the individual labeled regions become smaller and smaller. For the case with thousands of labeled regions performance can even become slower than on the CPU.


gigony (approved):
Thanks @grlee77 !
It looks good to me!

PR #846, opened 2025-03-03
Region Properties Performance Overhaul - Part 4: Moment-Based Properties

grlee77:

## Overview

This MR implements many properties based on image moments. These include centroids, normalized/central moments and measurements based on the inertia-tensor. These come in both unweighted and "weighted" variants where the weighted versions rely on values in a corresponding `intensity_image`.

The primary kernels that run on all image pixels are the computation of `moments` or `moments_weighted`. Other properties are typically derived from these raw moments and are very fast to compute once the moments of the required order have been computed.

Unweighted properties implemented are:
- axis_major_length
- axis_minor_length
- centroid
- centroid_local
- moments
- moments_central
- moments_normalized
- moments_hu
- inertia_tensor
- inertia_tensor_eigvals

Weighted properties implemented are:
- centroid_weighted
- centroid_weighted_local
- moments_weighted
- moments_weighted_central
- moments_weighted_normalized
- moments_weighted_hu

And a couple of properties that are useful, but no currently in scikit-image
- axis_lengths
    - retrieves lengths of all ellipsoid axes not just the largest and smallest (`axis_major_length` is axis_lengths[:, 0] while -`axis_minor_length` is axis_lengths[:, -1])
- inertia_tensor_eigenvectors
    - need this to know the spatial direction the ellipsoid axes correspond to

regionprops_moment_vs_size

regionprops_3d_moment_vs_size

Note that "multi-moment" is the time to compute the following list of region properties rather than just a single property
```python
properties = ["moments", "moments_central", "moments_normalized", "inertia_tensor", "inertia_tensor_eigvals", "axis_minor_length", "axis_major_length", "centroid", "centroid_local"]
```
and `"multi-moment-weighted" corresponds to
```python
properties = ["moments_weighted", "moments_weighted_central", "moments_weighted_normalized", "centroid_weighted", "centroid_weighted_local"]
```

regionprops_moment_vs_object_size

regionprops_3d_moment_vs_object_size

### Benchmark conclusions

**Note:** The results for the older GPU-based regionprops from cuCIM are not shown here. However, for many properties that implementation became much slower as the number of regions increased. We can see that for the new implementation proposed here, performance does not continuously decline once a certain number of objects are reached. There is still better acceleration for a small number of objects, but acceleration holds even for very large numbers of objects.


gforsyth (approved):
Approving for packaging codeowners

gigony (approved):
Thanks @grlee77 ! It looks good to me!

PR #847, opened 2025-03-04
Region Properties Performance Overhaul - Part 5: Perimeter and Euler Characteristic

grlee77:

## Overview

This MR implements the final few region properties supported by `regionprops_table`
- perimeter
- perimeter_crofton
- euler_number

The batch implementation here is not applicable to tightly packed regions (i.e. when regions share a border with other regions). The default `robust_perimeter=True` will maintain accuracy by computing the values for individual regions in isolation at cost of performance. 

If accurate perimeter values are not required, but only relative perimeter size, then "num_perimeter_pixels" as implemented in #843 just counts the number of pixels/voxels on the object perimeter is much faster to compute. That number of pixels measurement does not suffer from this region adjacency issue and can always safely be applied even to tightly packed regions.

regionprops_perimeter_vs_size

regionprops_3d_perimeter_vs_size

regionprops_perimeter_vs_object_size

regionprops_3d_perimeter_vs_object_size

### Benchmark conclusions

**Note:** We can see that performance improves as a function of image size (top 2 figures) and remains good across many labeled region sizes (lower 2 figures). 

**CAVEAT:**: The benchmarks shown above are for the case with `robust_perimeter=False` set. In this case the perimeter will not be accurate when labeled regions are immediately adjacent to each other (share a border with spacing of <= 1 background voxel). When `robust_perimeter=True`, performance will highly depend on how many labeled regions are within 1 pixel proximity of another label. Any such detected labels have to be processed in isolation to give accurate results and this will slow down the algorithm as we have to launch multiple CUDA kernels per each such labeled region.

If accurate perimeter values are not required, but only relative perimeter size, then "num_perimeter_pixels" which just counts the number of pixels/voxels on the object perimeter will be much faster to compute and does not suffer from this region adjacency issue.


python/cucim/src/cucim/skimage/measure/_regionprops_gpu_misc_kernels.py
```diff
@@ -0,0 +1,721 @@
...
+        border_image,
+        perimeter_weights,
+        mode="constant",
+        cval=0,
+        output=cp.uint8,
+    )
+
+    min_integer_type = _get_min_integer_dtype(max_label, signed=False)
+    # if labels.dtype != min_integer_type:
+    #     labels = labels_dilated.astype(min_integer_type)
```
gigony, line 219: We may want to remove this or move these lines after line 224 (after `labels_dilated` is defined).

python/cucim/src/cucim/skimage/measure/_regionprops_gpu_misc_kernels.py
```diff
@@ -0,0 +1,721 @@
...
+    >>> import cupy as cp
+    >>> from skimage import data
+    >>> from cucim.skimage import util
+    >>> from cucim.skimage.measure import label
+    >>> # coins image (binary)
+    >>> img_coins = cp.array(data.coins() > 110)
+    >>> # total perimeter of all objects in the image
+    >>> perimeter(img_coins, neighborhood=4)  # doctest: +ELLIPSIS
+    array(7796.86799644)
+    >>> perimeter(img_coins, neighborhood=8)  # doctest: +ELLIPSIS
```
gigony, line 649: It looks like the examples section incorrectly shows `perimeter` function calls instead of `euler_number`.

python/cucim/src/cucim/skimage/measure/_regionprops_gpu_misc_kernels.py
```diff
@@ -0,0 +1,721 @@
...
+    .. [1] S. Rivollier. Analyse d’image geometrique et morphometrique par
+           diagrammes de forme et voisinages adaptatifs generaux. PhD thesis,
+           2010. Ecole Nationale Superieure des Mines de Saint-Etienne.
+           https://tel.archives-ouvertes.fr/tel-00560838
+    .. [2] Ohser J., Nagel W., Schladitz K. (2002) The Euler Number of
+           Discretized Sets - On the Choice of Adjacency in Homogeneous
+           Lattices. In: Mecke K., Stoyan D. (eds) Morphology of Condensed
+           Matter. Lecture Notes in Physics, vol 600. Springer, Berlin,
+           Heidelberg.
+    --------
```
gigony, line 635:
```suggestion

    See Also
    --------
```

python/cucim/src/cucim/skimage/measure/_regionprops_gpu_misc_kernels.py
```diff
@@ -0,0 +1,721 @@
...
+        label_image,
+        label_image.size,
+        coefs_image,
+        output,
+        size=math.ceil(label_image.size / pixels_per_thread),
+    )
+    return output
+
+
+@cp.memoize(for_each_device=True)
```
gigony, line 85: Could not specifying the cache size here lead to memory issues with many different parameter combinations?

python/cucim/src/cucim/skimage/measure/_regionprops_gpu_misc_kernels.py
```diff
@@ -0,0 +1,721 @@
...
+            0,
+            0,
+            0,
+        ]
+    else:
+        sq2 = math.sqrt(2)
+        coefs = [
+            0,
+            np.pi / 4 * (1 + 1 / sq2),
+            np.pi / (4 * sq2),
```
gigony, line 303: I’m wondering if hardcoded floating-point values (e.g., np.pi / 4 * (1 + 1 / sq2)), which may introduce minor numerical errors, could be precomputed as high-precision constants instead.

gigony (approved):
Thanks @grlee77 for this update!
It looks good to me. Left minor comments.

python/cucim/src/cucim/skimage/measure/_regionprops_gpu_misc_kernels.py
grlee77: In this case the only items are the floating point data type and the GPU device ID so it should be only a handful of combinations. If you memoize over the exact image shape or something then that could be more problematic.

python/cucim/src/cucim/skimage/measure/_regionprops_gpu_misc_kernels.py
grlee77: We could, but I don't think it will make any practical difference. Python's double precision float should be accurate enough. There will be much larger error from the discrete pixel size than numerical precision loss in these values.

python/cucim/src/cucim/skimage/measure/_regionprops_gpu_misc_kernels.py
grlee77: thanks!

python/cucim/src/cucim/skimage/measure/_regionprops_gpu_misc_kernels.py
```diff
@@ -0,0 +1,721 @@
...
+
+    References
+    ----------
+    .. [1] K. Benkrid, D. Crookes. Design and FPGA Implementation of
+           a Perimeter Estimator. The Queen's University of Belfast.
+           http://www.cs.qub.ac.uk/~d.crookes/webpubs/papers/perimeter.doc
+
+    See Also
+    --------
+    perimeter_crofton
```
grlee77, line 175:
```suggestion
    regionprops_perimeter_crofton
```

python/cucim/src/cucim/skimage/measure/_regionprops_gpu_misc_kernels.py
```diff
@@ -0,0 +1,721 @@
...
+    of the field of view, so an accurate perimeter cannot be measured for such
+    objects.
+
+    If the labeled regions have holes, the hole edges will be included in this
+    measurement. If this is not desired, use regionprops_label_filled to fill
+    the holes and then pass the filled labels image to this function.
+
+    See Also
+    --------
+    perimeter
```
grlee77, line 412:
```suggestion
    regionprops_perimeter
```

python/cucim/src/cucim/skimage/measure/_regionprops_gpu_misc_kernels.py
```diff
@@ -0,0 +1,721 @@
...
+    >>> from skimage import data
+    >>> from cucim.skimage import util
+    >>> from cucim.skimage.measure import label
+    >>> # coins image (binary)
+    >>> img_coins = cp.array(data.coins() > 110)
+    >>> # total perimeter of all objects in the image
+    >>> perimeter(img_coins, neighborhood=4)  # doctest: +ELLIPSIS
+    array(7796.86799644)
+    >>> perimeter(img_coins, neighborhood=8)  # doctest: +ELLIPSIS
+    array(8806.26807333)
```
grlee77, line 650:
```suggestion
```

python/cucim/src/cucim/skimage/measure/_regionprops_gpu_misc_kernels.py
grlee77: I should just remove these "see also" and examples as we are not exposing these versions directly as public API. It is just for internal use within `regionprops_table`

python/cucim/src/cucim/skimage/measure/_regionprops_gpu_misc_kernels.py
grlee77, line 412:
```suggestion
```

python/cucim/src/cucim/skimage/measure/_regionprops_gpu_misc_kernels.py
```diff
@@ -0,0 +1,721 @@
...
```
grlee77, line 189:
```suggestion
```

PR #848, opened 2025-03-04
Region Properties Performance Overhaul - Part 6: Public API (regionprops_table)

grlee77:

## Overview

All of the available region properties were implemented for `regionprops_dict` in parts 1-5. 

This MR's purpose is just to add ability for the existing `cucim.skimage.measure.regionprops_table` function to call `regionprops_dict` by default under the hood. This will allow existing code based on `regionprops_table` to see an immediate performance benefit.

The user can set kwarg `batch_processing=False` to keep using the old implementation based on the `RegionProperties` class. I am not sure if we should provide that option or just remove it. I think the only case it may be useful is that memory requirements could be lower for the non-batch approach as properties are computed on individual regions in isolation. Due to caching behavior of `RegionProperties`, though, the memory usage might not actually be all that much lower.

For compatibility with `regionprops_table` API, the `regionprops_dict` implementation is updated to support user-provided callables for `extra_properties`. For such "extra" properties there will not be a performance benefit to `regionprops_dict` vs. legacy `regionprops_table` as it will also loop over individual regions and apply this function on each.


python/cucim/src/cucim/skimage/measure/_regionprops_gpu.py
```diff
@@ -0,0 +1,882 @@
...
+        n_args = _infer_number_of_required_args(func)
+        images = out["image"]
+        # determine whether func requires intensity image
+        results = []
+        if n_args == 2:
+            if intensity_image is not None:
+                images_intensity = out["image_intensity"]
+                for image, image_intensity in zip(images, images_intensity):
+                    multichannel = label_image.shape < intensity_image.shape
+                    multichannel = label_image.shape < intensity_image.shape
```
gigony, line 769: Please remove the repeated line.

python/cucim/src/cucim/skimage/measure/_regionprops_gpu.py
```diff
@@ -0,0 +1,882 @@
...
+                    if multichannel:
+                        multichannel_list = [
+                            func(images, images_intensity[..., i])
```
gigony, line 772: Just curious if `images` is intentional here, given that this statement is inside a for-loop using `image`.

python/cucim/src/cucim/skimage/measure/_regionprops_gpu.py
```diff
@@ -0,0 +1,882 @@
...
+    copy_to_host=False,
+):
+    out = {}
+    for prop in properties:
+        # Copy the original property name so the output will have the
+        # user-provided property name in the case of deprecated names.
+        orig_prop = prop
+        # determine the current property name for any deprecated property.
+        prop = PROPS_GPU.get(prop, prop)
+        # is_0dim_array = isinstance(rp, cp.ndarray) and rp.ndim == 0
```
gigony, line 827: could we remove this line?

python/cucim/src/cucim/skimage/measure/_regionprops_gpu.py
```diff
@@ -0,0 +1,882 @@
...
+        from each GPU thread. The number of adjacent pixels processed
+        corresponds to `pixels_per_thread` and can be used as a performance
+        tuning parameter.
+    robust_perimeter : bool, optional
+        Batch computation of perimeter and euler characteristics can give
+        incorrect results for perimeter pixels that are not more than 1 pixel
+        spacing from another label. If True, a check for this condition is
+        performed and any labels close to another label have their perimeter
+        recomputed in isolation. Doing this check results in performance
+        overhead so can optionally be disabled. This parameter effects the
```
gigony, line 285:
```suggestion
        overhead so can optionally be disabled. This parameter affects the
```

python/cucim/src/cucim/skimage/measure/_regionprops_gpu.py
```diff
@@ -0,0 +1,882 @@
...
+        optionally an extra dimension for multichannel data. The channel
+        dimension, if present, must be the last axis. Default is None.
+    properties : tuple or list of str, optional
+        Properties that will be included in the resulting dictionary
+        For a list of available properties, please see :func:`regionprops`.
+        Users should remember to add "label" to keep track of region
+        identities.
+    spacing : tuple of float, shape (ndim,), optional
+        The pixel spacing along each axis of the image.
+    extra_properties : Iterable of callables
```
gigony, line 254:
```suggestion
    extra_properties : Iterable of callables, optional
```

gigony (approved):
Thanks @grlee77 !
It looks good to me!

python/cucim/src/cucim/skimage/measure/_regionprops_gpu.py
grlee77: good catch! maybe there is no test case for this multi-channel intensity image case with a custom property. Let me test and push a fix

python/cucim/src/cucim/skimage/measure/_regionprops_gpu.py
grlee77, line 827:
```suggestion
```

python/cucim/src/cucim/skimage/measure/_regionprops_gpu.py
grlee77: Fixed as suggested and added a new test case to cover this code path

PR #856, opened 2025-03-25
CuPy 13.4.1 compatibility: Fix dtype handling in fused chan-vese kernels

grlee77:
Fix for recent test failures after CuPy 13.4.1 was released as seen here:
https://github.com/rapidsai/cucim/actions/runs/14052308871

There is a data type casting change in fused kernels used by `cucim.skimage.segmentation.chan_vese`. Likely this is related to the changes made in https://github.com/cupy/cupy/pull/9012. We can ensure the correct output dtype from the fused kernel by making sure all float scalars used as input to the fused kernels are device scalars of the desired precision.

### DIFF
--- a/python/cucim/src/cucim/skimage/segmentation/_chan_vese.py
+++ b/python/cucim/src/cucim/skimage/segmentation/_chan_vese.py
@@ -71,7 +71,7 @@ def _fused_variance_kernel2(
     return out
 
 
-def _cv_calculate_variation(image, phi, mu, lambda1, lambda2, dt):
+def _cv_calculate_variation(image, phi, mu, lambda1, lambda2, dt, eta):
     """Returns the variation of level set 'phi' based on algorithm parameters.
 
     This corresponds to equation (22) of the paper by Pascal Getreuer,
@@ -89,7 +89,6 @@ def _cv_calculate_variation(image, phi, mu, lambda1, lambda2, dt):
     similar approach is used by Rami Cohen, and it is from there that the
     C1-4 notation is taken.
     """
-    eta = cp.asarray(1e-16, dtype=phi.dtype)
     P = pad(phi, 1, mode="edge")
 
     x_end = P[1:-1, 2:]
@@ -413,6 +412,15 @@ def chan_vese(
         raise ValueError("Input image should be a 2D array.")
 
     float_dtype = _supported_float_type(image.dtype)
+
+    # cast Python float constants to device scalars of matching precision
+    mu = cp.asarray(mu, dtype=float_dtype)
+    lambda1 = cp.asarray(lambda1, dtype=float_dtype)
+    lambda2 = cp.asarray(lambda2, dtype=float_dtype)
+    dt = cp.asarray(dt, dtype=float_dtype)
+    tol = cp.asarray(tol, dtype=float_dtype)
+    eta = cp.asarray(1e-16, dtype=float_dtype)
+
     phi = _cv_init_level_set(init_level_set, image.shape, dtype=float_dtype)
     if type(phi) != cp.ndarray or phi.shape != image.shape:
         raise ValueError(
@@ -426,21 +434,17 @@ def chan_vese(
         image = image / cp.max(image)
 
     i = 0
-    mu = cp.asarray(mu, dtype=float_dtype)
-    lambda1 = cp.asarray(lambda1, dtype=float_dtype)
-    lambda2 = cp.asarray(lambda2, dtype=float_dtype)
     if extended_output:
         old_energy = _cv_energy(image, phi, mu, lambda1, lambda2)
         energies = []
 
     phivar = tol + 1
-    dt = cp.asarray(dt, dtype=float_dtype)
     while phivar > tol and i < max_num_iter:
         # Save old level set values
         oldphi = phi
 
         # Calculate new level set
-        phi = _cv_calculate_variation(image, phi, mu, lambda1, lambda2, dt)
+        phi = _cv_calculate_variation(image, phi, mu, lambda1, lambda2, dt, eta)
         phivar = phi - oldphi
         phivar *= phivar
         phivar = cp.sqrt(phivar.mean())
--- a/python/cucim/src/cucim/skimage/segmentation/_chan_vese.py
+++ b/python/cucim/src/cucim/skimage/segmentation/_chan_vese.py
@@ -89,7 +89,7 @@ def _cv_calculate_variation(image, phi, mu, lambda1, lambda2, dt):
     similar approach is used by Rami Cohen, and it is from there that the
     C1-4 notation is taken.
     """
-    eta = 1e-16
+    eta = cp.asarray(1e-16, dtype=phi.dtype)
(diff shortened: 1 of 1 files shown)


python/cucim/src/cucim/skimage/segmentation/_chan_vese.py
```diff
@@ -426,15 +426,15 @@ def chan_vese(
...
     i = 0
+    mu = cp.asarray(mu, dtype=float_dtype)
+    lambda1 = cp.asarray(lambda1, dtype=float_dtype)
+    lambda2 = cp.asarray(lambda2, dtype=float_dtype)
     if extended_output:
         old_energy = _cv_energy(image, phi, mu, lambda1, lambda2)
         energies = []
-    phivar = tol + 1
 
+    phivar = tol + 1
```
jakirkham, line 436: Think we want to do the same with `tol`

```suggestion
    tol = cp.asarray(tol, dtype=float_dtype)
    phivar = tol + 1
```

python/cucim/src/cucim/skimage/segmentation/_chan_vese.py
```diff
@@ -89,7 +89,7 @@ def _cv_calculate_variation(image, phi, mu, lambda1, lambda2, dt):
     similar approach is used by Rami Cohen, and it is from there that the
     C1-4 notation is taken.
     """
-    eta = 1e-16
+    eta = cp.asarray(1e-16, dtype=phi.dtype)
```
jakirkham, line 92: Also do we need to apply this in other cases?

```suggestion
    eta = cp.asarray(1e-16, dtype=phi.dtype)
    mu = cp.asarray(mu, dtype=phi.dtype)
    lambda1 = cp.asarray(lambda1, dtype=phi.dtype)
    lambda2 = cp.asarray(lambda2, dtype=phi.dtype)
    dt = cp.asarray(dt, dtype=phi.dtype)
```

python/cucim/src/cucim/skimage/segmentation/_chan_vese.py
```diff
@@ -426,15 +426,15 @@ def chan_vese(
         image = image / cp.max(image)
 
     i = 0
+    mu = cp.asarray(mu, dtype=float_dtype)
+    lambda1 = cp.asarray(lambda1, dtype=float_dtype)
+    lambda2 = cp.asarray(lambda2, dtype=float_dtype)
```
jakirkham, line 431: Should we move this and all other assignment right after the normalization of the `image` input above?

python/cucim/src/cucim/skimage/segmentation/_chan_vese.py
```diff
@@ -89,7 +89,7 @@ def _cv_calculate_variation(image, phi, mu, lambda1, lambda2, dt):
     similar approach is used by Rami Cohen, and it is from there that the
     C1-4 notation is taken.
     """
```
jakirkham, line 91: Do we need to normalize the input `image`'s `dtype` somehow?

```suggestion
    """
    float_dtype = _supported_float_type(image.dtype)
```

jakirkham (commented):
Thanks Greg! 🙏

This seems like a good approach

Had a few questions below

python/cucim/src/cucim/skimage/segmentation/_chan_vese.py
grlee77: except for `eta` those are done outside of the function call already. I should probably refactor so `eta` is also passed into the argument so this doesn't have to be converted repeatedly in the loop

python/cucim/src/cucim/skimage/segmentation/_chan_vese.py
grlee77: This function determines whether to use 32 or 64-bit floating point types based on [this dict](https://github.com/rapidsai/cucim/blob/61d8a3923d2f6c87da136676e64754796f379826/python/cucim/src/cucim/skimage/_shared/utils.py#L870-L887)

python/cucim/src/cucim/skimage/segmentation/_chan_vese.py
grlee77: sure can move it all there

python/cucim/src/cucim/skimage/segmentation/_chan_vese.py
grlee77: For `tol`, it is not passed to any fused kernels, but it may still be good to transfer it to device in this way

jakirkham: Yeah think it is a good idea to do it with all scalars. It is too easy to miss this kind of thing and there is not much cost to converting them all

Can see you already did this above. So marking resolved

Thanks Greg! 🙏

jakirkham: Providing a little more context on why this is helpful for `tol` specifically in the comments review comments below

python/cucim/src/cucim/skimage/segmentation/_chan_vese.py
```diff
@@ -429,18 +437,14 @@ def chan_vese(
...
         old_energy = _cv_energy(image, phi, mu, lambda1, lambda2)
         energies = []
-    phivar = tol + 1
 
-    dt = cp.asarray(dt, dtype=float_dtype)
-    mu = cp.asarray(mu, dtype=float_dtype)
-    lambda1 = cp.asarray(lambda1, dtype=float_dtype)
-    lambda2 = cp.asarray(lambda2, dtype=float_dtype)
+    phivar = tol + 1
     while phivar > tol and i < max_num_iter:
```
jakirkham, line 442: As we compare `phivar` to `tol`, it helps if they are both in the same location (on device or on host). Given `phivar` will be on device (as noted above), it makes sense to have `tol` on device too

python/cucim/src/cucim/skimage/segmentation/_chan_vese.py
```diff
@@ -429,18 +437,14 @@ def chan_vese(
...
         # Save old level set values
         oldphi = phi
 
         # Calculate new level set
-        phi = _cv_calculate_variation(image, phi, mu, lambda1, lambda2, dt)
+        phi = _cv_calculate_variation(image, phi, mu, lambda1, lambda2, dt, eta)
         phivar = phi - oldphi
         phivar *= phivar
         phivar = cp.sqrt(phivar.mean())
```
jakirkham, line 450: Just noting that `phivar` becomes a device scalar here

jakirkham (commented):
Thanks Greg! 🙏

Added a few comments just for context around `tol` and `phivar` as we discussed them

jakirkham (approved):
Thanks Greg! 🙏

LGTM

python/cucim/src/cucim/skimage/segmentation/_chan_vese.py
jakirkham: Gotcha. Thanks Greg! 🙏

PR #890, opened 2025-06-10
refactor(shellcheck): fix all shellcheck warnings/errors

gforsyth:
Finishes off the work proposed in https://github.com/rapidsai/build-planning/issues/135

### DIFF
--- a/.github/workflows/pr.yaml
+++ b/.github/workflows/pr.yaml
@@ -1,14 +1,11 @@
 name: pr
-
 on:
   push:
     branches:
       - "pull-request/[0-9]+"
-
 concurrency:
   group: ${{ github.workflow }}-${{ github.ref }}
   cancel-in-progress: true
-
 jobs:
   pr-builder:
     needs:
@@ -86,7 +83,6 @@ jobs:
     with:
       build_type: pull-request
       script: ci/test_wheel.sh
-
   telemetry-summarize:
     # This job must use a self-hosted runner to record telemetry traces.
     runs-on: linux-amd64-cpu4
@@ -96,3 +92,5 @@ jobs:
     steps:
       - name: Telemetry summarize
         uses: rapidsai/shared-actions/telemetry-dispatch-summarize@main
+    env:
+      GH_TOKEN: ${{ github.token }}
--- a/conda/recipes/cucim/conda_build_config.yaml
+++ b/conda/recipes/cucim/conda_build_config.yaml
@@ -1,14 +1,11 @@
 c_compiler_version:
-  - 13  # [not os.environ.get("RAPIDS_CUDA_VERSION", "").startswith("11")]
-  - 11  # [os.environ.get("RAPIDS_CUDA_VERSION", "").startswith("11")]
+  - 13
 
 cxx_compiler_version:
-  - 13  # [not os.environ.get("RAPIDS_CUDA_VERSION", "").startswith("11")]
-  - 11  # [os.environ.get("RAPIDS_CUDA_VERSION", "").startswith("11")]
+  - 13
 
 cuda_compiler:
-  - cuda-nvcc  # [not os.environ.get("RAPIDS_CUDA_VERSION", "").startswith("11")]
-  - nvcc  # [os.environ.get("RAPIDS_CUDA_VERSION", "").startswith("11")]
+  - cuda-nvcc
 
 c_stdlib:
   - sysroot
--- a/conda/recipes/cucim/meta.yaml
+++ b/conda/recipes/cucim/meta.yaml
@@ -19,9 +19,7 @@ build:
   string: cuda{{ cuda_major }}_py{{ py_version }}_{{ date_string }}_{{ GIT_DESCRIBE_HASH }}_{{ GIT_DESCRIBE_NUMBER }}
   ignore_run_exports_from:
     - {{ compiler('cuda') }}
-    {% if cuda_major != "11" %}
     - cuda-cudart-dev
-    {% endif %}
   script_env:
     - AWS_ACCESS_KEY_ID
     - AWS_SECRET_ACCESS_KEY
@@ -43,11 +41,7 @@ requirements:
   build:
     - {{ compiler("c") }}
     - {{ compiler("cxx") }}
-    {% if cuda_major == "11" %}
-    - {{ compiler('cuda') }} ={{ cuda_version }}
-    {% else %}
     - {{ compiler('cuda') }}
-    {% endif %}
     - cuda-version ={{ cuda_version }}
     - cmake {{ cmake_version }}
     - make
@@ -56,9 +50,7 @@ requirements:
   host:
     - click
     - cuda-version ={{ cuda_version }}
-    {% if cuda_major != "11" %}
     - cuda-cudart-dev
-    {% endif %}
     - cupy >=12.0.0
     - libcucim ={{ version }}
     - python
@@ -68,9 +60,7 @@ requirements:
     - setuptools >=61.0.0
   run:
     - {{ pin_compatible('cuda-version', max_pin='x', min_pin='x') }}
-    {% if cuda_major != "11" %}
     - cuda-cudart
-    {% endif %}
     - numpy >=1.23,<3.0a0
     - click
     - cupy >=12.0.0
--- a/conda/recipes/libcucim/meta.yaml
+++ b/conda/recipes/libcucim/meta.yaml
@@ -20,11 +20,9 @@ build:
     - openslide
   ignore_run_exports_from:
     - {{ compiler('cuda') }}
-    {% if cuda_major != "11" %}
     - cuda-cudart-dev
     - libcufile-dev  # [linux64]
     - libnvjpeg-dev
-    {% endif %}
   script_env:
     - AWS_ACCESS_KEY_ID
     - AWS_SECRET_ACCESS_KEY
@@ -46,11 +44,7 @@ requirements:
   build:
     - {{ compiler("c") }}
     - {{ compiler("cxx") }}
-    {% if cuda_major == "11" %}
-    - {{ compiler('cuda') }} ={{ cuda_version }}
-    {% else %}
     - {{ compiler('cuda') }}
-    {% endif %}
     - cuda-version ={{ cuda_version }}
     - binutils
     - cmake {{ cmake_version }}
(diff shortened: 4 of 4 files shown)


benchmarks/skimage/run-nv-bench-color.sh
```diff
@@ -1,22 +1,22 @@
...
-param_shape=(512,512 3840,2160)
+param_shape=("512,512" "3840,2160")
 param_filt=(deltaE_cie76 deltaE_ciede94 deltaE_ciede2000 deltaE_cmc)
-param_dt=(float32, float64)
+param_dt=(float32 float64)
 for shape in "${param_shape[@]}"; do
     for filt in "${param_filt[@]}"; do
         for dt in "${param_dt[@]}"; do
-            python cucim_color_bench.py -f $filt -i $shape -d $dt -t 10
+            python cucim_color_bench.py -f "$filt" -i "$shape" -d "$dt" -t 10
```
gforsyth: This is one spot where I'm curious to see what happens - given that the first value of the `param_dt` array was `float32,`

benchmarks/skimage/run-nv-bench-morphology.sh
```diff
@@ -15,9 +15,9 @@ done
 param_filt_float=(remove_small_objects remove_small_holes erosion dilation opening closing white_tophat black_tophat reconstruction)
 param_dt=(float32)
 for shape in "${param_shape[@]}"; do
-    for filt in "${param_filt[@]}"; do
+    for filt in "${param_filt_float[@]}"; do
```
gforsyth, line 18: I'm assuming that `param_filt_float` is named that way for a reason -- otherwise the variable is unused.  So this particular benchmark has been running on the larger param set defined above on line 3

gforsyth:
I'm going to wait until tomorrow to debug `run`

benchmarks/skimage/run-nv-bench-color.sh
```diff
@@ -1,22 +1,22 @@
...
         done
     done
 done
 
-param_shape=(512,512 3840,2160)
+param_shape=("512,512" "3840,2160")
 param_filt=(deltaE_cie76 deltaE_ciede94 deltaE_ciede2000 deltaE_cmc)
-param_dt=(float32, float64)
+param_dt=(float32 float64)
```
jakirkham: Should we also quote these?

```suggestion
param_dt=("float32" "float64")
```

benchmarks/skimage/run-all.sh
```diff
@@ -1,5 +1,7 @@
+#!/bin/bash
```
jakirkham, line 1: Should we add copyright headers to these?

conda/recipes/cucim/build.sh
```diff
@@ -1,3 +1,4 @@
+#!/bin/bash
 # Copyright (c) 2021, NVIDIA CORPORATION.
```
jakirkham:
```suggestion
# Copyright (c) 2021-2025, NVIDIA CORPORATION.
```

conda/recipes/libcucim/build.sh
```diff
@@ -1,3 +1,4 @@
+#!/bin/bash
 # Copyright (c) 2021, NVIDIA CORPORATION.
```
jakirkham:
```suggestion
# Copyright (c) 2021-2025, NVIDIA CORPORATION.
```

run
```diff
@@ -17,12 +17,15 @@
 init_globals() {
     if [ "$0" != "/bin/bash" ]; then
         SCRIPT_DIR=$(dirname "$(readlink -f "$0")")
-        export RUN_SCRIPT_FILE="$(readlink -f "$0")"
+        RUN_SCRIPT_FILE="$(readlink -f "$0")"
+        export RUN_SCRIPT_FILE
```
jakirkham: Out of curiosity, what motivates the split?

run
```diff
@@ -17,12 +17,15 @@
 init_globals() {
     if [ "$0" != "/bin/bash" ]; then
         SCRIPT_DIR=$(dirname "$(readlink -f "$0")")
```
jakirkham: Should we reuse `RUN_SCRIPT_FILE` to inform `SCRIPT_DIR`?

```suggestion
        RUN_SCRIPT_FILE="$(readlink -f "$0")"
        SCRIPT_DIR="$(dirname "'"'"${RUN_SCRIPT_FILE}"'"'")"
```

run
```diff
@@ -17,12 +17,15 @@
...
-        export RUN_SCRIPT_FILE="$(readlink -f "$0")"
+        RUN_SCRIPT_FILE="$(readlink -f "$0")"
```
jakirkham, line 19: If we move `RUN_SCRIPT_FILE` above, it can be dropped here

```suggestion
```

run
```diff
@@ -141,9 +148,11 @@ c_str() {
 }
 
 c_echo() {
+    # shellcheck disable=SC2155
     local old_opt="$(shopt -op xtrace)" # save old xtrace option
```
jakirkham, line 152: Guessing breaking out `local old_opt` on a prior line causes the `shopt` call not to work. Is that right?

run
```diff
@@ -372,64 +384,66 @@ build_local() {
...
         if [ "${CONDA_BUILD:-}" != "1" ]; then
             # Copy .so files from libcucim & cuslide/cumed's build folders to cuCIM's Python source folder
             # Since wheel file doesn't support symbolic link (https://github.com/pypa/wheel/issues/203),
             # we don't need to copy symbolic links. Instead copy only libcucim.so.${major_version} (without symbolic link)
-            cp ${TOP}/build-$build_type/lib*/libcucim.so.${major_version} ${TOP}/python/cucim/src/cucim/clara/
-            cp -P ${TOP}/cpp/plugins/cucim.kit.cuslide/build-$build_type/lib*/cucim* ${TOP}/python/cucim/src/cucim/clara/
-            cp -P ${TOP}/cpp/plugins/cucim.kit.cumed/build-$build_type/lib*/cucim* ${TOP}/python/cucim/src/cucim/clara/
+            cp "${TOP}"/build-"$build_type"/lib*/libcucim.so."${major_version}" "${TOP}"/python/cucim/src/cucim/clara/
+            cp -P "${TOP}"/cpp/plugins/cucim.kit.cuslide/build-"$build_type"/lib*/cucim* "${TOP}"/python/cucim/src/cucim/clara/
+            cp -P "${TOP}"/cpp/plugins/cucim.kit.cumed/build-"$build_type"/lib*/cucim* "${TOP}"/python/cucim/src/cucim/clara/
```
jakirkham: Should the `$build_type` variables use `{...}`?

```suggestion
            cp "${TOP}"/build-"${build_type}"/lib*/libcucim.so."${major_version}" "${TOP}"/python/cucim/src/cucim/clara/
            cp -P "${TOP}"/cpp/plugins/cucim.kit.cuslide/build-"${build_type}"/lib*/cucim* "${TOP}"/python/cucim/src/cucim/clara/
            cp -P "${TOP}"/cpp/plugins/cucim.kit.cumed/build-"${build_type}"/lib*/cucim* "${TOP}"/python/cucim/src/cucim/clara/
```

run
```diff
@@ -715,12 +738,6 @@ parse_args() {
     while getopts 'yh' option;
     do
         case "${option}" in
-            # a)
-            #     VALUE=${OPTARG}
-            #     ;;
-            y)
-                ALWAYS_YES=true;
-                ;;
```
jakirkham, line 723: Is there a reason to drop this case

jakirkham (commented):
Thanks Gil! 🙏

These seem like good improvements

Had a few questions below

run
gforsyth: It's a common codesmell in `bash` -- if you `export` and define the variable in a single line it swallows any errors that may get thrown by the captured command.

run
gforsyth: nailed it

run
gforsyth: We can keep it in.  I removed it because it's unused

gforsyth: As in, `ALWAYS_YES` is defined here and then never touched again (nor exported)

run
gforsyth: I'll reuse it, but I think we don't need as many nested quotation marks -- I'll push up a fix

benchmarks/skimage/run-all.sh
gforsyth: We can.  Do we say 2025 since that's when the notice is added?

benchmarks/skimage/run-nv-bench-morphology.sh
jameslamb: I agree with your read of this, it sure *looks like* this was intended to use `param_filt_float`.

All the changes to this file in the blame ([link](https://github.com/rapidsai/cucim/blame/branch-25.08/benchmarks/skimage/run-nv-bench-morphology.sh)) are from you @grlee77, could you comment on this change?

conda/recipes/cucim/build.sh
```diff
@@ -11,13 +12,13 @@ export LD_LIBRARY_PATH="$BUILD_PREFIX/lib:$PREFIX/lib:$LD_LIBRARY_PATH"
 
 # It is assumed that this script is executed from the root of the repo directory by conda-build
 # (https://conda-forge.org/docs/maintainer/knowledge_base.html#using-cmake)
-./run build_local cucim ${CUCIM_BUILD_TYPE}
+./run build_local cucim "${CUCIM_BUILD_TYPE}"
 
 cp -P python/install/lib/* python/cucim/src/cucim/clara/
 
-pushd python/cucim
+pushd python/cucim || exit 1
```
jameslamb: Would `shellcheck` stop complaining about this if we did a `set -e` at the top of this script?

If so, I think that'd be preferable. And honestly either way, I think it'd be helpful for this to get `set -e -u -o pipefail` added to it, so failures in e.g. the `pip install` below actually result in failures in CI. Could you add that please?

jameslamb (approved):
Wow this repo has some SERIOUS `bash` in it. Really glad we're adding `shellcheck`, thanks for doing that!

I left a few very very minor comments. Could you also please link to https://github.com/rapidsai/build-planning/issues/135 from the description?

I don't need to re-review, but  `@` me if you want my input on anything else here.

conda/recipes/cucim/build.sh
gforsyth: Yep, good call

run
jakirkham: Yeah shell quotes are fun

Basically this is [the concatenation of multiple strings]( https://stackoverflow.com/a/1250279 )

That said, welcome a simpler approach

run
jakirkham: Good catch!

Maybe we should file these in a follow up issue

benchmarks/skimage/run-nv-bench-morphology.sh
jakirkham: This would be a good item for a follow up issue

PR #889, opened 2025-06-10
refactor(conda): remove cuda11 conditionals from conda recipes

gforsyth:
xref rapidsai/build-planning#184

PR #905, opened 2025-07-16
Remove CUDA 11 references

jakirkham:
Fixes https://github.com/rapidsai/cucim/issues/904

Removes remaining CUDA 11 references.


jakirkham:
Also seeing some references in the notebooks

https://github.com/rapidsai/cucim/blob/44c559284714c79c8a5af2a90212106f6b2a89f6/notebooks/Accessing_File_with_GDS.ipynb#L30-L38

https://github.com/rapidsai/cucim/blob/44c559284714c79c8a5af2a90212106f6b2a89f6/notebooks/Basic_Usage.ipynb#L25

Though have not addressed those

ci/test_wheel.sh
```diff
@@ -11,8 +11,6 @@ PYTHON_WHEELHOUSE=$(RAPIDS_PY_WHEEL_NAME="cucim_${RAPIDS_PY_CUDA_SUFFIX}" rapids
 # echo to expand wildcard before adding `[extra]` requires for pip
 rapids-pip-retry install "$(echo ${PYTHON_WHEELHOUSE}/cucim*.whl)[test]"
 
-CUDA_MAJOR_VERSION=${RAPIDS_CUDA_VERSION:0:2}
```
jakirkham: Splitting this out into its own PR: https://github.com/rapidsai/cucim/pull/907

jakirkham:
Breaking out the CUDA 11 doc removal work into PR: https://github.com/rapidsai/cucim/pull/914

Will update this after that PR is merged and retarget to 25.10

jakirkham:
The CUDA 11 doc removal changes are in

Updated this PR to pull those in

Also retargeting to 25.10

gds/CMakeLists.txt
```diff
@@ -110,7 +80,7 @@ PUBLIC
 target_include_directories(cufile_stub
     PUBLIC
         $<BUILD_INTERFACE:${CMAKE_CURRENT_LIST_DIR}/include>
-        $<BUILD_INTERFACE:${GDS_INCLUDE_PATH}>
+        CUDAToolkit_INCLUDE_DIRS
     PRIVATE
         # Add path to cufile.h explicitly. ${TOP}/temp/cuda would be available by `./run copy_gds_files_`
```
grlee77: remove this comment as copy_gds_files is being removed in this PR

gds/CMakeLists.txt
```diff
@@ -110,7 +80,7 @@ PUBLIC
 target_include_directories(cufile_stub
     PUBLIC
         $<BUILD_INTERFACE:${CMAKE_CURRENT_LIST_DIR}/include>
-        $<BUILD_INTERFACE:${GDS_INCLUDE_PATH}>
+        CUDAToolkit_INCLUDE_DIRS
```
grlee77: I think we need to expand the value here
```suggestion
        ${CUDAToolkit_INCLUDE_DIRS}
```

grlee77: let me go ahead and commit this and see if it resolves the existing build failure for wheels...

grlee77:
Some notebooks also have outdated `pip` install commands using cuda 11 versions of CuPy. It looks like we should replace `cupy-cuda110` with `cupy-cuda12x` in the following places:

```
notebooks/Basic_Usage.ipynb:25:    "!pip install numpy scipy scikit-image cupy-cuda110  # for cucim dependency (assuming that CUDA 11.0 is used for CuPy)\n",
notebooks/Basic_Usage.ipynb:40:    "# !pip install numpy scipy scikit-image cupy-cuda110  # for cucim dependency (assuming that CUDA 11.0 is used for CuPy)"
notebooks/Accessing_File_with_GDS.ipynb:37:    "!pip install cupy-cuda110\n",
notebooks/Accessing_File_with_GDS.ipynb:52:    "#!pip install cupy-cuda110\n",
notebooks/Working_with_DALI.ipynb:23:    "!pip install --extra-index-url https://developer.download.nvidia.com/compute/redist --upgrade nvidia-dali-cuda110\n",
notebooks/Working_with_DALI.ipynb:27:    "!pip install cucim scipy scikit-image cupy-cuda110 matplotlib\n",
notebooks/Working_with_DALI.ipynb:28:    "!pip install --extra-index-url https://developer.download.nvidia.com/compute/redist --upgrade nvidia-dali-cuda110\n",
notebooks/Working_with_DALI.ipynb:39:    "#!pip install --extra-index-url https://developer.download.nvidia.com/compute/redist --upgrade nvidia-dali-cuda110\n",
notebooks/Working_with_DALI.ipynb:43:    "#!pip install cucim scipy scikit-image cupy-cuda110 matplotlib\n",
notebooks/Working_with_DALI.ipynb:45:    "#!pip install --extra-index-url https://developer.download.nvidia.com/compute/redist --upgrade nvidia-dali-cuda110"
```

grlee77 (approved):
Thanks @jakirkham, it looks good to me. 

I pushed a 1-line commit that seems to have resolved finding the CUDA include path for GDS. Other changes look good to me, but @gigony, please confirm if you have a chance as you are the original author of most of the modified files.

I left a couple of comments for additional minor residual cleanups that could be made.

gds/CMakeLists.txt
bdice:
```suggestion
```

bdice (approved):

jameslamb:
After handling `culibos` being removed from `cuda-cudart`, I didn't face any other issues adding CUDA 13 support: https://github.com/rapidsai/cucim/pull/926

If we can merge this soon, I'll update the CUDA 13 PR and ensure everything is still working together. If more work is needed here, we should merge the CUDA 13 PR first.

grlee77:
I was able to push to John's branch, so just pushed the notebook commit here. Should be able to just merge once CI passes

bdice:
Great. I'll do that now, to help along CUDA 13 work.

cpp/plugins/cucim.kit.cuslide/CMakeLists.txt
```diff
@@ -231,6 +224,22 @@ target_link_libraries(${CUCIM_PLUGIN_NAME}
...
+                CUDA::culibos
+                CUDA::cudart
+        )
+else()
+        target_link_libraries(${CUCIM_PLUGIN_NAME}
+            PRIVATE
+                CUDA::nvjpeg
+                CUDA::cudart
+        )
+endif()
```
jakirkham, line 242: Right now this checks for nvJPEG static and uses it if it finds it. Otherwise dynamically linking to nvJPEG

This works for the needs of cuCIM. Namely wheels continue to statically link to nvJPEG. Meanwhile Conda packages can dynamically link to nvJPEG

However what would be better is to add an option to build nvJPEG either dynamically or statically (like [with cuFile's option]( https://github.com/rapidsai/cucim/blob/b738d2cfcc902dac0b824ab84197e4e8c57160f6/CMakeLists.txt#L95 )). The code here could similarly be updated to check that option (just like [cuFile's linking logic]( https://github.com/rapidsai/cucim/blob/b738d2cfcc902dac0b824ab84197e4e8c57160f6/gds/CMakeLists.txt#L57-L72 )). Then wheel and Conda builds could pass that flag based on their needs

gds/CMakeLists.txt
```diff
@@ -52,52 +55,19 @@ target_compile_options(cufile_stub
         )
 
 ## Link libraries
-target_link_libraries(cufile_stub
+if (CUCIM_STATIC_GDS)
+    # Enabling CUCIM_STATIC_GDS statically links cuFile
+    target_link_libraries(cufile_stub
         PUBLIC
             ${CMAKE_DL_LIBS}
         PRIVATE
```
jakirkham, line 63: This diff (and the diff behind the fold) looks more exciting than it is

All this does is reproduce the previous logic

https://github.com/rapidsai/cucim/blob/54cc342c8aa9a82cd41b080dd764358b327f3ddd/gds/CMakeLists.txt#L96-L101

It just replaces `deps::gds_static` with `CUDA::cuFile_static`. The former was a stand-in for what CMake now provides. So we just switch to the latter

gds/CMakeLists.txt
```diff
@@ -52,52 +55,19 @@ target_compile_options(cufile_stub
...
-        set(GDS_STATIC_LIB_PATH /usr/local/cuda/lib64/libcufile_static.a)
-    endif ()
-
-    message("Set GDS_STATIC_LIB_PATH to '${GDS_STATIC_LIB_PATH}'.")
-
-    set_target_properties(deps::gds_static PROPERTIES
-        IMPORTED_LOCATION "${GDS_STATIC_LIB_PATH}"
-        INTERFACE_INCLUDE_DIRECTORIES "${GDS_INCLUDE_PATH}"
+            CUDA::cuFile_static
     )
```
jakirkham, line 65: Here is where `deps::gds_static` was replaced with `CUDA::cuFile_static` . For some reason GitHub put a large number of deleted lines between the two. Unclear why it did that. In any event, this is connected with the above content

gds/CMakeLists.txt
```diff
@@ -52,52 +55,19 @@ target_compile_options(cufile_stub
         )
 
 ## Link libraries
-target_link_libraries(cufile_stub
+if (CUCIM_STATIC_GDS)
+    # Enabling CUCIM_STATIC_GDS statically links cuFile
+    target_link_libraries(cufile_stub
         PUBLIC
             ${CMAKE_DL_LIBS}
```
jakirkham, line 62: As another note, it appears we define a bunch of stuff unconditionally that needs `dlopen`, etc. and thus need to link to [`${CMAKE_DL_LIBS}`]( https://cmake.org/cmake/help/latest/variable/CMAKE_DL_LIBS.html ). However we [don't actually need to use it in the static case]( https://github.com/rapidsai/cucim/blob/54cc342c8aa9a82cd41b080dd764358b327f3ddd/gds/src/cufile_stub.cpp#L59 ). So it seems this linkage could be dropped (after adding some of [the C++ code in guards]( https://github.com/rapidsai/cucim/blob/54cc342c8aa9a82cd41b080dd764358b327f3ddd/gds/src/cufile_stub.cpp#L18-L21 ))

gds/CMakeLists.txt
```diff
@@ -52,52 +55,19 @@ target_compile_options(cufile_stub
...
-    set(GDS_INCLUDE_PATH "$ENV{PREFIX}/include")
-elseif (DEFINED ENV{CONDA_PREFIX} AND EXISTS $ENV{CONDA_PREFIX}/include/cufile.h)
-    set(GDS_INCLUDE_PATH "$ENV{CONDA_PREFIX}/include")
-elseif (EXISTS ${CMAKE_CURRENT_SOURCE_DIR}/../temp/cuda/include/cufile.h)
-    set(GDS_INCLUDE_PATH ${CMAKE_CURRENT_SOURCE_DIR}/../temp/cuda/include)
-else ()
-    set(GDS_INCLUDE_PATH /usr/local/cuda/include)
-endif ()
-
-message("Set GDS_INCLUDE_PATH to '${GDS_INCLUDE_PATH}'.")
```
jakirkham, line 73: This was just to find the path to the cuFile download we did on CI

Using [CMake's `CUDA::cuFile*` targets]( https://cmake.org/cmake/help/latest/module/FindCUDAToolkit.html#cufile ) to pick up the cuFile in the CUDA Toolkit obviates this need

gds/CMakeLists.txt
```diff
@@ -52,52 +55,19 @@ target_compile_options(cufile_stub
...
-        set(GDS_STATIC_LIB_PATH "$ENV{PREFIX}/lib/libcufile_static.a")
-    elseif (DEFINED ENV{CONDA_PREFIX})
-        set(GDS_STATIC_LIB_PATH "$ENV{CONDA_PREFIX}/lib/libcufile_static.a")
-    elseif (EXISTS ${CMAKE_CURRENT_SOURCE_DIR}/../temp/cuda/lib64/libcufile_static.a)
-        set(GDS_STATIC_LIB_PATH ${CMAKE_CURRENT_SOURCE_DIR}/../temp/cuda/lib64/libcufile_static.a)
-    else ()
-        set(GDS_STATIC_LIB_PATH /usr/local/cuda/lib64/libcufile_static.a)
-    endif ()
-
-    message("Set GDS_STATIC_LIB_PATH to '${GDS_STATIC_LIB_PATH}'.")
```
jakirkham, line 90: Again obviated by [CMake's `CUDA::cuFile*` targets](https://cmake.org/cmake/help/latest/module/FindCUDAToolkit.html#cufile)

gds/CMakeLists.txt
```diff
@@ -52,52 +55,19 @@ target_compile_options(cufile_stub
...
+            CUDA::cuFile_static
     )
+else()
+    # Use `dlopen` to load cuFile at runtime
     target_link_libraries(cufile_stub
         PUBLIC
             ${CMAKE_DL_LIBS}
-        PRIVATE
-            deps::gds_static
     )
```
jakirkham, line 71: This is just the same dynamic cuFile library case as before

https://github.com/rapidsai/cucim/blob/54cc342c8aa9a82cd41b080dd764358b327f3ddd/gds/CMakeLists.txt#L54-L60

For some reason GitHub included `deps::gds_static` as removed, but it is just confusing things. So ignore that

Also as we use `dlopen` to load cuFile, we don't need to link to it in this case. Hence why `CUDA::cuFile` is not here. We do add the CTK header files later (as before)

We did include the `CUDA::cudart` before. However did not actually see usage of `cudart`. So dropped that. If there was an issue, we would have seen that at link time, but we don't see any

That said, @gigony please let me know if I'm missing something here

gds/CMakeLists.txt
```diff
@@ -110,9 +80,8 @@ PUBLIC
 target_include_directories(cufile_stub
     PUBLIC
         $<BUILD_INTERFACE:${CMAKE_CURRENT_LIST_DIR}/include>
-        $<BUILD_INTERFACE:${GDS_INCLUDE_PATH}>
+        ${CUDAToolkit_INCLUDE_DIRS}
```
jakirkham, line 83: Since cuFile is picked up from the CTK now, this can just be the CTK header directories

gds/CMakeLists.txt
```diff
@@ -110,9 +80,8 @@ PUBLIC
...
     PRIVATE
-        # Add path to cufile.h explicitly. ${TOP}/temp/cuda would be available by `./run copy_gds_files_`
```
jakirkham, line 115: As Greg already noted, the comment above is irrelevant. So it is dropped

gds/include/cufile_stub.h
```diff
@@ -16,7 +16,12 @@
 #ifndef CUCIM_CUFILE_STUB_H
 #define CUCIM_CUFILE_STUB_H
 
-#include "cufile.h"
+// Try to include the real cufile.h, fall back to minimal types if not available
+#if __has_include(<cufile.h>)
+    #include <cufile.h>
+#else
+    #include "cufile_stub_types.h"
+#endif
```
jakirkham, line 24: @grlee77 it looks like you added this change. Could you please share a bit more context?

Given the CTK has `cufile.h`, would think we could just use that `#include` directly

Is there some other issue that is coming up?

gds/src/cufile_stub.cpp
```diff
@@ -61,19 +61,9 @@ void CuFileStub::load()
...
                     "libcufile.so.0",
-                    "libcufile.so.1.3.0" /* 11.7.0 */,
-                    "libcufile.so.1.2.1" /* 11.6.2, 11.6.1 */,
-                    "libcufile.so.1.2.0" /* 11.6.0 */,
-                    "libcufile.so.1.1.1" /* 11.5.1 */,
-                    "libcufile.so.1.1.0" /* 11.5.0 */,
-                    "libcufile.so.1.0.2" /* 11.4.4, 11.4.3, 11.4.2 */,
-                    "libcufile.so.1.0.1" /* 11.4.1 */,
-                    "libcufile.so.1.0.0" /* 11.4.0 */
                 },
```
jakirkham, line 67: In CUDA 11, cuFile's SOVERSION changed repeatedly. Hence we had this workaround. Since we used `dlopen`, the workarounds for all CUDA 11 versions were needed even when building with CUDA 11.8 only

Now that CUDA 11 is dropped, none of these workarounds are needed. Hence this now just loads `libcufile.so.0`

notebooks/Accessing_File_with_GDS.ipynb
```diff
@@ -34,7 +34,7 @@
     "\n",
     "or\n",
     "```\n",
-    "!pip install cupy-cuda110\n",
+    "!pip install cupy-cuda12x\n",
     "!pip install torch==1.7.1+cu110 torchvision==0.8.2+cu110 torchaudio==0.7.2 -f https://download.pytorch.org/whl/torch_stable.html\n",
```
jakirkham, line 38: @grlee77 thanks for updating the CuPy packages! 🙏

Should we do the same with PyTorch and friends?

There are a bunch more lines like these throughout the notebooks. Perhaps we should update them all and rerun to make sure they are working ok

notebooks/Basic_Usage.ipynb
```diff
@@ -32,12 +32,12 @@
    "metadata": {},
    "outputs": [],
    "source": [
-    "#!conda install -c conda-forge pillow\n",
+    "# !conda install -c conda-forge pillow\n",
```
jakirkham, line 35: Looks like an errant space snuck in here

run
```diff
@@ -281,11 +281,6 @@ build_local_libcucim_() {
 
     pushd "${source_folder}" > /dev/null
 
-    # Copy cufile SDK from host system to temp/cuda
-    copy_gds_files_ "$source_folder"
-    # Copy libjpeg SDK from host system to temp/cuda
-    copy_nvjpeg_files_ "$source_folder"
```
jakirkham, line 287: This is how we installed the cuFile and nvJPEG when we needed them and they were either not in the CTK or the CTK versions were too old. However this is no longer an issue with CUDA 12.0+

Hence we are able to drop these and the two function definitions below

dependencies.yaml
```diff
@@ -173,7 +173,6 @@ dependencies:
               cuda: "12.*"
             packages:
               - libnvjpeg-dev
-              - libnvjpeg-static
```
jakirkham, line 176: Now that the nvJPEG version we need is in the CTK and there are Conda packages for it, we have switched to dynamic linking

Already commented on the relevant build logic making this change above

cpp/plugins/cucim.kit.cuslide/CMakeLists.txt
```diff
@@ -231,6 +224,22 @@ target_link_libraries(${CUCIM_PLUGIN_NAME}
             deps::json
             deps::libdeflate
         )
+if (TARGET CUDA::nvjpeg_static)
+        target_link_libraries(${CUCIM_PLUGIN_NAME}
+            PRIVATE
+		# Add nvjpeg before cudart so that nvjpeg.h in static library takes precedence.
+                CUDA::nvjpeg_static
+                # Add CUDA::culibos to link necessary methods for 'deps::nvjpeg_static'
+                CUDA::culibos
```
jakirkham, line 233: Looks like I goofed on the indenting here. Will follow up on this

gds/CMakeLists.txt
```diff
@@ -24,6 +24,9 @@ if (NOT APPLE)
     set(CMAKE_INSTALL_RPATH $ORIGIN)
 endif ()
 
+# Find CUDA Toolkit for cudart and cufile.
+find_package(CUDAToolkit REQUIRED)
```
jakirkham, line 28: To ensure cuFile is picked up from the CTK, we add this `find_package` line

Note that in the case where nvJPEG is used, we already had an equivalent line. Hence no change there

conda/recipes/libcucim/conda_build_config.yaml
```diff
@@ -1,14 +1,11 @@
...
 
 cxx_compiler_version:
-  - 13  # [not os.environ.get("RAPIDS_CUDA_VERSION", "").startswith("11")]
-  - 11  # [os.environ.get("RAPIDS_CUDA_VERSION", "").startswith("11")]
+  - 13
 
 cuda_compiler:
-  - cuda-nvcc  # [not os.environ.get("RAPIDS_CUDA_VERSION", "").startswith("11")]
-  - nvcc  # [os.environ.get("RAPIDS_CUDA_VERSION", "").startswith("11")]
+  - cuda-nvcc
```
jakirkham, line 8: No more GCC 11 now that we are on CUDA 12+

This sticks with GCC 13 on CUDA 12 for simplicity

Though we have been bumping to GCC 14 in RAPIDS projects. So that could be done as a follow up

conda/recipes/libcucim/conda_build_config.yaml
```diff
@@ -18,16 +15,3 @@ c_stdlib_version:
...
-# and the "*_run_*" version specifiers correspond to `11.x` packages.
-
-cuda11_libcufile_host_version:
-  - "1.4.0.31"
-
-cuda11_libcufile_run_version:
-  - ">=1.0.0.82,<=1.4.0.31"
-
-cuda11_libnvjpeg_host_version:
-  - "11.6.0.55"
```
jakirkham, line 33: These are CUDA 11 pinnings for dependencies outside the Conda `cudatoolkit` package

None of that is relevant for CUDA 12+, which has the full CTK as packages with associated pinning

Also the usage of these values was dropped already with PR: https://github.com/rapidsai/cucim/pull/889

Hence these can just be dropped

conda/recipes/libcucim/meta.yaml
```diff
@@ -55,15 +55,14 @@ requirements:
   host:
     - cuda-version ={{ cuda_version }}
     - cuda-cudart-dev
-    - libcufile-dev     # [linux64]
+    - libcufile-dev
```
jakirkham, line 58: cuFile is available on all platforms. Hence it is added as a dependency here

conda/recipes/libcucim/meta.yaml
```diff
@@ -55,15 +55,14 @@ requirements:
...
     - libnvjpeg-dev
-    - libnvjpeg-static
```
jakirkham, line 60: nvJPEG is now dynamically linked to in the Conda case. Hence no need for the static library

conda/recipes/libcucim/meta.yaml
```diff
@@ -55,15 +55,14 @@ requirements:
...
     - nvtx-c >=3.1.0
     - openslide
   run:
     - {{ pin_compatible('cuda-version', max_pin='x', min_pin='x') }}
     - cuda-cudart
-    - libcufile  # [linux64]
+    - libcufile
```
jakirkham, line 65: This adds the cuFile dependency on ARM

However the CTK only added cuFile for ARM in CUDA 12.2: https://github.com/conda-forge/libcufile-feedstock/pull/9

So this won't work for ARM with CUDA 12.0 or 12.1. Meaning we need to use a similar workaround as KvikIO did: https://github.com/rapidsai/kvikio/pull/754

Another option would be to just start depending on KvikIO

cpp/plugins/cucim.kit.cuslide/CMakeLists.txt
```diff
@@ -125,8 +125,6 @@ superbuild_depend(cli11)
 superbuild_depend(pugixml)
 superbuild_depend(json)
 superbuild_depend(libdeflate)
-superbuild_depend(nvjpeg)
-superbuild_depend(libculibos)
```
jakirkham, line 129: These handled finding the nvJPEG and cuLIBOS dependencies via custom CMake logic

That is no longer needed now that they come from the CTK

Hence this and the associated files are dropped

jakirkham (commented):
Just wanted to say thank you all for pushing this across the line ❤️

Had the sense this got 80-90% of the way during 25.08. However wasn't able to finish in that cycle so just split out the doc fixes and moved this to 25.10

There were some remaining things that I hoped to do here. Mainly fixing CI, cleaning up more CUDA 11 references, and some build fixes for issues I caught along the way

Think the changes that all of you made are good enough for this to be in. Will leave some comments here to discuss the remaining work that can be split out. Also to leave some context (especially for Gigon when he looks at this)

conda/recipes/libcucim/meta.yaml
jakirkham: Addressing in PR: https://github.com/rapidsai/cucim/pull/930

cpp/plugins/cucim.kit.cuslide/CMakeLists.txt
jakirkham: Addressing in PR: https://github.com/rapidsai/cucim/pull/931

cpp/plugins/cucim.kit.cuslide/CMakeLists.txt
jakirkham: Raised this in a new issue: https://github.com/rapidsai/cucim/issues/932

gds/CMakeLists.txt
jakirkham: Raised in issue: https://github.com/rapidsai/cucim/issues/933

PR #906, opened 2025-07-16
remove cuspatial references

jameslamb:
## Description

Contributes to https://github.com/rapidsai/build-planning/issues/197

* removes references to `cuspatial` in some TODO comments... those are outdated because they refer to things possibly being added to `cuspatial` in the future, and that library's no longer being actively maintained
* fixes `pre-commit` checks

### How I tested this

Looked for references like this:

```shell
git grep -i -E 'cuproj|cuspatial'
```

### DIFF
--- a/python/cucim/src/cucim/skimage/feature/corner.py
+++ b/python/cucim/src/cucim/skimage/feature/corner.py
@@ -4,6 +4,7 @@ from itertools import combinations_with_replacement
 
 import cupy as cp
 import numpy as np
+
 # TODO: use CuPy's KDTree once it becomes available
 # xref: https://github.com/rapidsai/cucim/issues/732
 from scipy import spatial


ci/test_wheel.sh
```diff
@@ -11,8 +11,6 @@ PYTHON_WHEELHOUSE=$(RAPIDS_PY_WHEEL_NAME="cucim_${RAPIDS_PY_CUDA_SUFFIX}" rapids
 # echo to expand wildcard before adding `[extra]` requires for pip
 rapids-pip-retry install "$(echo ${PYTHON_WHEELHOUSE}/cucim*.whl)[test]"
 
-CUDA_MAJOR_VERSION=${RAPIDS_CUDA_VERSION:0:2}
-
```
jameslamb: This unrelated-looking changed fixes `pre-commit` checks, which reported this:

```text
In ci/test_wheel.sh line 14:
CUDA_MAJOR_VERSION=${RAPIDS_CUDA_VERSION:0:2}
^----------------^ SC2034 (warning): CUDA_MAJOR_VERSION appears unused. Verify use (or export if used externally).
```

That variable was left behind by the changes in #903, and CI there reported the same error ([build link](https://github.com/rapidsai/cucim/actions/runs/16124571929/job/45498333365#step:7:194)), but that PR was admin-merged with that accidentally unfixed.

jakirkham: Thanks James! 🙏

Ran into the same thing when doing more CUDA 11 removals in PR: https://github.com/rapidsai/cucim/pull/905

Given this is a common issue, have split that out in PR: https://github.com/rapidsai/cucim/pull/907

That will help us track the change (in case we learn there is some other script consuming this variable)

jameslamb: Ok, sure. I didn't really think it was worth an entire separate CI run... the other change in this PR is so small and non-controversial, I'd expected it to get approved and merged quickly.

python/cucim/src/cucim/skimage/feature/corner.py
```diff
@@ -1370,7 +1370,7 @@ def corner_peaks(
     )
 
     if len(coords):
-        # TODO: modify to do KDTree on the GPU (cuSpatial?)
+        # TODO: modify to do KDTree on the GPU
         coords = cp.asnumpy(coords)
```
jakirkham: Yeah as noted in the comment, this code falls back to the CPU for something that is best done on the GPU

At the time, we thought cuSpatial might provide a solution, in which case we could switch to cuSpatial for the GPU

However this comment (and the associated code) was written a while ago. As you note cuSpatial would not be the way to go today

That said, CuPy, with PR ( https://github.com/cupy/cupy/pull/7671 ), has added [`KDTree`]( https://docs.cupy.dev/en/latest/reference/generated/cupyx.scipy.spatial.KDTree.html ). This won't be available until CuPy 14.0.0

So instead would update this comment like so

```suggestion
        # TODO: modify to do KDTree on the GPU using CuPy
        # xref: https://github.com/rapidsai/cucim/issues/732
        coords = cp.asnumpy(coords)
```

python/cucim/src/cucim/skimage/feature/corner.py
```diff
@@ -4,7 +4,7 @@
 
 import cupy as cp
 import numpy as np
-from scipy import spatial  # TODO: use cuSpatial if KDTree becomes available
+from scipy import spatial
```
jakirkham, line 10: Let's leave this comment and simply reword it like so

```suggestion
# TODO: use CuPy's KDTree once it becomes available
# xref: https://github.com/rapidsai/cucim/issues/732
from scipy import spatial
```

jakirkham (commented):
Thanks James! 🙏

Have included some suggestions below based on more recent history, which these comments should now reflect

jakirkham:
Side note: Would have expected the Python changes would request a review from `@rapidsai/cucim-python-codeowners`. Though that doesn't appear to have happened 

https://github.com/rapidsai/cucim/blob/a4da26bdc655148e31f50c43182f68808f9f8906/.github/CODEOWNERS#L5-L6

python/cucim/src/cucim/skimage/feature/corner.py
jameslamb: ok thanks for that context! Works for me, and glad there's a path to maybe doing more of this on the GPU 😁

jameslamb:
> _Would have expected the Python changes would request a review from `@rapidsai/cucim-python-codeowners`_

Oh, it did!

When I opened this, GitHub reported that like this:

Screenshot 2025-07-17 at 11 30 21 AM

You just don't see that team under "Reviewers" any more because your review was "on behalf of" that group.

Screenshot 2025-07-17 at 11 31 14 AM

So CODEOWNERS is working exactly as it's expected to here.

jakirkham:
It didn't show up in the "Reviews Required" section at the bottom of the PR

jakirkham:
Dropping unneeded review requests and rerunning failed CI

PR #907, opened 2025-07-16
Drop unused variable `CUDA_MAJOR_VERSION`

jakirkham:
Previously `CUDA_MAJOR_VERSION` was used to special case CUDA 11 wheel builds. However that check went away in PR: https://github.com/rapidsai/cucim/pull/903

As a result, `shellcheck` now flags `CUDA_MAJOR_VERSION` as an unused variable

Fix this by dropping the unneeded `CUDA_MAJOR_VERSION` assignment


jakirkham:
Thanks Bradley! 🙏

PR #914, opened 2025-07-31
Drop CUDA 11 references from docs

jakirkham:
Partial fixes https://github.com/rapidsai/cucim/issues/904

Removes the remaining CUDA 11 doc references. Code is left unchanged.

### DIFF
--- a/python/cucim/pyproject.toml
+++ b/python/cucim/pyproject.toml
@@ -40,7 +40,6 @@ classifiers = [
     "Topic :: Scientific/Engineering",
     "Operating System :: POSIX :: Linux",
     "Environment :: Console",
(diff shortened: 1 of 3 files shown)


python/cucim/pyproject.toml
```diff
@@ -40,7 +40,6 @@ classifiers = [
     "Topic :: Scientific/Engineering",
     "Operating System :: POSIX :: Linux",
     "Environment :: Console",
-    "Environment :: GPU :: NVIDIA CUDA :: 11",
     "Environment :: GPU :: NVIDIA CUDA :: 12",
```
jakirkham, line 43: Also dropped the CUDA 11 trove classifier so CUDA 11 doesn't show up on the PyPI page

This is what the packaging review is for. The rest is just docs

PR #926, opened 2025-08-19
Build and test with CUDA 13.0.0, use GCC 14 for conda builds

jameslamb:
Contributes to https://github.com/rapidsai/build-planning/issues/208

* uses CUDA 13.0.0 to build and test
* moves some dependency pins:
  - `cupy`: `>=13.6.0`

Contributes to https://github.com/rapidsai/build-planning/issues/68

* updates to CUDA 13 dependencies in fallback entries in `dependencies.yaml` matrices (i.e., the ones that get written to `pyproject.toml` in source control)

## Notes for Reviewers

This switches GitHub Actions workflows to the `cuda13.0` branch from here: https://github.com/rapidsai/shared-workflows/pull/413

A future round of PRs will revert that back to `branch-25.10`, once all of RAPIDS supports CUDA 13.

### DIFF
--- a/dependencies.yaml
+++ b/dependencies.yaml
@@ -235,7 +235,7 @@ dependencies:
           - numpy>=1.23.4,<3.0a0
           - scikit-image>=0.19.0,<0.26.0a0
           - scipy>=1.11.2
-      - output_types: [conda]
+      - output_types: conda
         packages:
           - &cupy_unsuffixed cupy>=13.6.0
     specific:


jameslamb:
CUDA 13 C++ builds are failing like this:

```text
CMake Error at cmake/deps/libculibos.cmake:38 (message):
  Could not find CUDA::culibos.
Call Stack (most recent call first):
  cmake/modules/SuperBuildUtils.cmake:22 (include)
  CMakeLists.txt:129 (superbuild_depend)

BuildScriptException: Command '['/bin/bash', '-o', 'errexit', '/opt/conda/conda-bld/work/conda_build.sh']' returned non-zero exit status 1.
```

([build link](https://github.com/rapidsai/cucim/actions/runs/17080316548/job/48432407540?pr=926#step:10:2310))

dependencies.yaml
```diff
@@ -110,26 +113,14 @@ dependencies:
           - matrix:
               arch: x86_64
             packages:
+              - gcc_linux-64=13.*
```
bdice: We're using GCC 14 everywhere else in RAPIDS. Should this be bumped?

jameslamb: Sure, let's try that.

Looks like that update was not attempted as part of https://github.com/rapidsai/build-planning/issues/188, so hopefully the use of GCC 13 here is just an oversight and not an intentional choice.

jameslamb:
One `conda-python-build` job failed like this:

```text
Getting pinned dependencies: ...working... done

# >>>>>>>>>>>>>>>>>>>>>> ERROR REPORT <<<<<<<<<<<<<<<<<<<<<<

    Traceback (most recent call last):
      File "/opt/conda/lib/python3.11/site-packages/conda/exception_handler.py", line 28, in __call__
        return func(*args, **kwargs)
               ^^^^^^^^^^^^^^^^^^^^^
      File "/opt/conda/lib/python3.11/site-packages/conda/cli/main.py", line 61, in main_subshell
        exit_code = do_call(args, parser)
                    ^^^^^^^^^^^^^^^^^^^^^
      File "/opt/conda/lib/python3.11/site-packages/conda/cli/conda_argparse.py", line 182, in do_call
        result = plugin_subcommand.action(getattr(args, "_args", args))
                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
      File "/opt/conda/lib/python3.11/site-packages/conda_build/plugin.py", line 17, in build
        return execute(args)
               ^^^^^^^^^^^^^
      File "/opt/conda/lib/python3.11/site-packages/conda_build/cli/main_build.py", line 622, in execute
        api.build(
      File "/opt/conda/lib/python3.11/site-packages/conda_build/api.py", line 211, in build
        return build_tree(
               ^^^^^^^^^^^
      File "/opt/conda/lib/python3.11/site-packages/conda_build/build.py", line 3669, in build_tree
        packages_from_this = build(
                             ^^^^^^
      File "/opt/conda/lib/python3.11/site-packages/conda_build/build.py", line 2350, in build
        output_metas = expand_outputs([(m, need_source_download, need_reparse_in_env)])
                       ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
      File "/opt/conda/lib/python3.11/site-packages/conda_build/render.py", line 1000, in expand_outputs
        for output_dict, m in deepcopy(_m).get_output_metadata_set(
                              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
      File "/opt/conda/lib/python3.11/site-packages/conda_build/metadata.py", line 2781, in get_output_metadata_set
        conda_packages = finalize_outputs_pass(
                         ^^^^^^^^^^^^^^^^^^^^^^
      File "/opt/conda/lib/python3.11/site-packages/conda_build/metadata.py", line 1077, in finalize_outputs_pass
        fm = finalize_metadata(
             ^^^^^^^^^^^^^^^^^^
      File "/opt/conda/lib/python3.11/site-packages/conda_build/render.py", line 722, in finalize_metadata
        build_unsat, host_unsat = add_upstream_pins(
                                  ^^^^^^^^^^^^^^^^^^
      File "/opt/conda/lib/python3.11/site-packages/conda_build/render.py", line 477, in add_upstream_pins
        build_deps, build_unsat, extra_run_specs_from_build = _read_upstream_pin_files(
                                                              ^^^^^^^^^^^^^^^^^^^^^^^^^
      File "/opt/conda/lib/python3.11/site-packages/conda_build/render.py", line 463, in _read_upstream_pin_files
        extra_run_specs = get_upstream_pins(m, precs, env)
                          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
      File "/opt/conda/lib/python3.11/site-packages/conda_build/render.py", line 433, in get_upstream_pins
        loc, dist = execute_download_actions(
                    ^^^^^^^^^^^^^^^^^^^^^^^^^
      File "/opt/conda/lib/python3.11/site-packages/conda_build/render.py", line 388, in execute_download_actions
        link_prec = [
                    ^
    IndexError: list index out of range

`$ /opt/conda/bin/conda build --no-test --channel /tmp/tmp.eDNgGfJHkK conda/recipes/cucim`
```

([conda-python-build link](https://github.com/rapidsai/cucim/actions/runs/17136062724/job/48612994806?pr=926#step:10:472))

Hoping that's some temporary interruption. Waiting for the last wheel-testing job to finish, then I'll re-run the failed jobs.

dependencies.yaml
jameslamb: Looks like GCC 14 worked 😀 

https://github.com/rapidsai/cucim/actions/runs/17136062724?pr=926

conda/recipes/libcucim/meta.yaml
```diff
@@ -55,6 +55,10 @@ requirements:
   host:
     - cuda-version ={{ cuda_version }}
     - cuda-cudart-dev
+    {% if cuda_major != "12" %}
+    # cuda-culibos was split out from cudart in CUDA 13: https://docs.nvidia.com/cuda/cuda-toolkit-release-notes/index.html#cuda-runtime
+    - cuda-culibos-static
+    {% endif %}
```
jakirkham: This was only needed for statically linking nvJPEG

Conda is dynamically linking nvJPEG as of PR: https://github.com/rapidsai/cucim/pull/905

So we can drop this

```suggestion
```

dependencies.yaml
```diff
@@ -110,26 +113,14 @@ dependencies:
...
-              cuda: "12.*"
-            packages:
-              - gcc_linux-64=13.*
-              - cuda-nvcc
-          - matrix:
-              arch: aarch64
-              cuda: "12.*"
-            packages:
-              - gcc_linux-aarch64=13.*
-              - cuda-nvcc
```
jakirkham, line 132: Thanks for cleaning this up! 🙏

dependencies.yaml
```diff
@@ -250,18 +256,6 @@ dependencies:
...
-          - cupy>=12.0.0
-          # Not sure where these go, if anywhere:
-          # - openslide
-          # - xorg-libxcb
-    specific:
-      - output_types: [requirements, pyproject]
-        matrices:
-          - matrix: # All CUDA 12 versions
-            packages:
-              - cupy-cuda12x>=12.0.0
```
jakirkham: Previously we just had this logic with CUDA 11 & 12

Could we do something similar for CUDA 13.0?

https://github.com/rapidsai/cucim/blob/becb456415439c263b9c11a0bd11737b26ae60a9/dependencies.yaml#L289-L315

dependencies.yaml
```diff
@@ -250,18 +256,6 @@ dependencies:
           - numpy>=1.23.4,<3.0a0
           - scikit-image>=0.19.0,<0.26.0a0
           - scipy>=1.11.2
-      - output_types: conda
-        packages:
-          - cupy>=12.0.0
-          # Not sure where these go, if anywhere:
-          # - openslide
-          # - xorg-libxcb
```
jakirkham: This comment traces back to this PR discussion with @bdice : https://github.com/rapidsai/cucim/pull/471#discussion_r1063810045

That said, think we have gotten on without doing anything here for some time. So it is probably ok to drop

@gigony do you have thoughts on how these dependencies (`openslide` and `xorg-libxcb`) should be listed/used if at all?

In either case this can probably be taken into a new issue. Just flagging since it is getting removed and we are likely to forget without some way of tracking this question

dependencies.yaml
```diff
@@ -159,35 +150,50 @@ dependencies:
...
+              cuda: "12.*"
+            packages:
+              - cupy-cuda12x>=13.6.0
+          - matrix:
+              cuda: "13.*"
+            packages:
+              - &cupy_cu13 cupy-cuda13x>=13.6.0
+          - matrix:
+            packages:
+              - *cupy_cu13
```
jakirkham: Think we can keep this local to the one case it is used as noted below

dependencies.yaml
```diff
@@ -100,6 +102,7 @@ dependencies:
       - output_types: conda
         packages:
           - c-compiler
+          - cuda-nvcc
           - cxx-compiler
```
jakirkham, line 104: We can consolidate this to `cuda-compiler`, which pulls in [the other compilers]( https://github.com/conda-forge/cuda-compiler-feedstock/blob/b4313ac3b6a046f95c764d4d6aecd0ed4956f334/recipe/meta.yaml#L22-L23 ) as well

```suggestion
        packages:
          - cuda-compiler
```

jakirkham (commented):
Thanks James! 🙏

Appreciate the additional cleanups following PR: https://github.com/rapidsai/cucim/pull/905

Had a few comments below

dependencies.yaml
jakirkham: Let's handle discussion in issue: https://github.com/rapidsai/cucim/issues/929

conda/recipes/libcucim/meta.yaml
jameslamb: oh great!

dependencies.yaml
jameslamb: It looks like that is pulling in more than just compilers, like very specific versions of `cuda-cuobjdump` and `cuda-ctadvisor`. Is pulling in all those other tightly-pinned things worth it in exchange for not needing to explicitly write `c-compiler` and `cxx-compiler` here?

jakirkham: Yeah the versions are bumped as part of each CTK release. They look tight, but effectively they are no different than specifying `cuda-version=12.9` for example

Would think these things are useful for development. That said, no strong feelings here

Thought it might be nice to list one thing instead of several, but if we don't feel that is worth it we can skip it

dependencies.yaml
jameslamb: For these complex cases with suffixes or extras, having this dependency spelled the same way across all of RAPIDS is helpful, for automation with `rapids-reviser` and even just for manual copy-pasting.

That said, this is only used in one place in thie repo and I want to keep this moving forward, so I've restored the original form where it's included in this `run:` group, here: https://github.com/rapidsai/cucim/pull/926/commits/f2af692ac0f8a48e0cf27a7ee53cb5c5a6e334d1

dependencies.yaml
jameslamb: did that here: https://github.com/rapidsai/cucim/pull/926#discussion_r2292552372

dependencies.yaml
```diff
@@ -250,18 +235,20 @@ dependencies:
           - numpy>=1.23.4,<3.0a0
           - scikit-image>=0.19.0,<0.26.0a0
           - scipy>=1.11.2
-      - output_types: conda
+      - output_types: [conda]
```
jakirkham: This could be a nice change. Though it seems we don't do it consistently

Maybe we skip this for now and revisit when we want to handle more formatting cleanup

```suggestion
      - output_types: conda
```

dependencies.yaml
```diff
@@ -159,28 +148,24 @@ dependencies:
...
       - output_types: conda
         matrices:
           - matrix:
               arch: aarch64
             packages:
           - matrix:
               arch: x86_64
-              cuda: "12.*"
             packages:
               - libcufile-dev
```
jakirkham, line 170: Good catch!

Likely missed this (along with other things) during CUDA 11 removal

Nice seeing all this clutter going away 🙂

dependencies.yaml
jakirkham: Went ahead and put this small style change in. Hope that is ok

dependencies.yaml
jameslamb: Thanks for explaining that. I think we should skip it until there's a specific need.

dependencies.yaml
jameslamb: I actually prefer the `[]` form for everything, even single-item ones, because it makes it slightly easier to compare the lists at a glance. But don't feel strongly either way and yes we've been very inconsistent about it. Fine to leave in the style change you already pushed.

dependencies.yaml
jameslamb: Yep I'm also happy to see simplifications like this across the CUDA 13 PRs 😁

dependencies.yaml
jameslamb: Thanks for opening that! I've subscribed there. Agree, think an issue is a better place to track something this open-ended.

dependencies.yaml
jameslamb: Since this is the only remaining item and there aren't any correctness issues with `cuda-nvcc` (just developer convenience), I'm going to merge this PR as-is.

CONTRIBUTING.md
```diff
@@ -97,7 +97,7 @@ If codespell is finding false positives in newly added code, the `ignore-words-l
 
 Compiler requirements:
 
-* `gcc`     version 13.0+
+* `gcc`     version 14.0+
 * `nvcc`    version 12.0+
```
bdice, line 101: For the gcc and nvcc versions to make sense in context together, we'd need to require nvcc 12.8+ to support the stated gcc minimum of 14.0. nvcc 12.0 only supports gcc up to 12.0.

dependencies.yaml
bdice: We don't want `cuda-compiler` here. As James pointed out, we _intentionally_ use `c-compiler`, `cxx-compiler`, `cuda-nvcc` to get a more minimal dependency set than what comes with `cuda-compiler`.

PR #930, opened 2025-08-22
Relax `libcufile` dependency for old CTK 12 on ARM

jakirkham:
Following up on this thread: https://github.com/rapidsai/cucim/pull/905#discussion_r2292609197

The CUDA Toolkit (CTK) added cuFile for ARM in CTK 12.2. As a result always requiring `libcufile` with `libcucim` can make it uninstallable on ARM with CTK 12.0 & 12.1.

To fix this, build two variants of the `libcucim` package. One with `libcufile` and one without. Constrain `cuda-version` accordingly so that `libcufile` is installed whenever possible. Only when `libcufile` cannot be installed, fallback to the `libcucim` package where the `libcufile` dependency is left off.

The `x86_64` CTK has had cuFile since the CTK 11.4. As RAPIDS now requires CUDA 12.0+, there is no issue always requiring `libcufile` on `x86_64`.

This issue will not be present in CUDA 13+. So always require `libcufile` in that case.

### DIFF
--- a/ci/build_cpp.sh
+++ b/ci/build_cpp.sh
@@ -26,7 +26,7 @@ sccache --zero-stats
 # In most cases, the CTK has cuFile.
 # However the CTK only added cuFile for ARM in 12.2.
 # So for ARM users on CTK 12.0 & 12.1, relax the cuFile requirement.
-# On x86_64 or CTK 13, always require cuFile.
+# On x86_64 or CTK 13 or ARM with CTK 12.2+, always require cuFile.
 cat > extra_variants.yaml <<EOF
 has_cufile:
   - True


gds/include/cufile_stub.h
```diff
@@ -16,12 +16,7 @@
 #ifndef CUCIM_CUFILE_STUB_H
 #define CUCIM_CUFILE_STUB_H
 
-// Try to include the real cufile.h, fall back to minimal types if not available
-#if __has_include(<cufile.h>)
-    #include <cufile.h>
-#else
-    #include "cufile_stub_types.h"
-#endif
+#include <cufile.h>
```
jakirkham, line 19: Trying to `#include "cufile_stub_types.h"` causes [the following error on CI]( https://github.com/rapidsai/cucim/actions/runs/17146818503/job/48644697316?pr=930#step:10:1425 ):

```cpp
In file included from /opt/conda/conda-bld/work/gds/src/cufile_stub.cpp:17:
/opt/conda/conda-bld/work/gds/include/cufile_stub.h:23:14: fatal error: cufile_stub_types.h: No such file or directory
   23 |     #include "cufile_stub_types.h"
      |              ^~~~~~~~~~~~~~~~~~~~~
compilation terminated.
```

Given we either have a new enough CUDA Toolkit and it has cuFile or it doesn't, we can simply check for cuFile and build or not

Unsure what we gain by having this in-between mode. Though would be happy to learn if I'm missing something

ci/build_cpp.sh
```diff
@@ -21,6 +21,25 @@ conda config --set path_conflict warn
...
+cat > extra_variants.yaml <<EOF
+has_cufile:
+  - True
+EOF
+if [[ "$(arch)" == "aarch64" ]] && [[ "${RAPIDS_CUDA_VERSION%%.*}" == "12" ]]
+then
+  cat >> extra_variants.yaml <<EOF
+  - False
+EOF
+fi
```
grlee77, line 39: Looks like the current logic will have both in the aarch64+CUDA12 case?

```
has_cufile:
  - True
  -  False
```

```suggestion
if [[ "$(arch)" == "aarch64" ]] && [[ "${RAPIDS_CUDA_VERSION%%.*}" == "12" ]]; then
  cat > extra_variants.yaml <<EOF
has_cufile:
  - False
EOF
else
  cat > extra_variants.yaml <<EOF
has_cufile:
  - True
EOF
fi
```

grlee77: Let me commit this and see if it helps

jakirkham:
The error in [the CI log]( https://github.com/rapidsai/cucim/actions/runs/17587012959/job/49960108189#step:10:1 ) points to cuFile being used in the build and not being found

```cpp
In file included from /opt/conda/conda-bld/work/gds/src/cufile_stub.cpp:17:
 /opt/conda/conda-bld/work/gds/include/cufile_stub.h:19:10: fatal error: cufile.h: No such file or directory
Z    19 | #include <cufile.h>
4Z       |          ^~~~~~~~~~
76Z compilation terminated.
111Z
```

Will either need to...

1. Always build with `libcufile-dev` and simply selectively drop the runtime dependency on `libcufile`
2. Rework the build logic to handle cuFile being absent

Personally would lean towards 1 as that should be less effort. Plus the effort in 2 becomes irrelevant once CUDA 12.0 & 12.1 are dropped from support

conda/recipes/libcucim/meta.yaml
```diff
@@ -55,14 +55,26 @@ requirements:
   host:
     - cuda-version ={{ cuda_version }}
     - cuda-cudart-dev
+    {% if aarch64 and cuda_major == "12" %}
+    # Skip building with `libcufile-dev` if it cannot be used at runtime
+    - libcufile-dev  # [has_cufile]
+    {% else %}
     - libcufile-dev
+    {% endif %}
```
jakirkham:
```suggestion
    - libcufile-dev
```

jakirkham:
Pushed the option 1 approach. Let's see how that goes

ci/build_cpp.sh
jakirkham: Yes this is intended. We want to build once for each of these values, which is what the original logic does

Also for CUDA 13+ or `x86_64` builds we want this to only be `True`

The CI failure is due to something else unrelated to this syntax. Please see this comment: https://github.com/rapidsai/cucim/pull/930#issuecomment-3303386248

Have gone ahead and restored the original logic. Though happy to chat more if you have additional questions 🙂

jakirkham:
At this point this PR has the needed pieces to address the issue

High-level cuCIM is designed to always build with cuFile and optionally use it at runtime (only if it can find and use the cuFile library)

As RAPIDS takes the approach of building with the latest CUDA minor version for any major version (like 12.9 for the 12 series), cuFile is available at build time for all architectures. So building cuCIM with cuFile on CUDA 12.9 works unconditionally

At runtime, cuCIM has two modes of using cuFile:

1. Statically linked cuFile
2. `dlopen` to load the library

In RAPIDS we only ever build cuCIM with 2 (`dlopen`)

https://github.com/rapidsai/cucim/blob/16c07c6313ad148b2122e52f5d1efdd2f83616ab/gds/src/cufile_stub.cpp#L64-L68

With this approach, cuCIM will only use cuFile if it can `dlopen` it

https://github.com/rapidsai/cucim/blob/16c07c6313ad148b2122e52f5d1efdd2f83616ab/cpp/src/filesystem/cufile_driver.cpp#L313-L316

However if cuCIM cannot `dlopen` cuFile, it falls back to normal POSIX IO

https://github.com/rapidsai/cucim/blob/16c07c6313ad148b2122e52f5d1efdd2f83616ab/cpp/src/filesystem/cufile_driver.cpp#L332-L339

So all we ever need do for old CUDA 12 ARM builds is drop the cuFile dependency at runtime, which is what this PR does

Opening this up for review

ci/build_cpp.sh
```diff
@@ -21,6 +21,29 @@ conda config --set path_conflict warn
 
 sccache --zero-stats
 
-RAPIDS_PACKAGE_VERSION=$(rapids-generate-version) rapids-conda-retry build conda/recipes/libcucim
+
+# TODO: Remove when CUDA 12.1 is dropped.
+# In most cases, the CTK has cuFile.
+# However the CTK only added cuFile for ARM in 12.2.
+# So for ARM users on CTK 12.0 & 12.1, relax the cuFile requirement.
+# On x86_64 or CTK 13, always require cuFile.
```
jakirkham:
```suggestion
# On x86_64 or CTK 13 or ARM with CTK 12.2+, always require cuFile.
```

conda/recipes/libcucim/meta.yaml
```diff
@@ -15,13 +15,13 @@ source:
 
 build:
   number: {{ GIT_DESCRIBE_NUMBER }}
-  string: cuda{{ cuda_major }}_{{ date_string }}_{{ GIT_DESCRIBE_HASH }}_{{ GIT_DESCRIBE_NUMBER }}
+  string: cuda{{ cuda_major }}_{{ date_string }}_gh{{ GIT_DESCRIBE_HASH }}_gn{{ GIT_DESCRIBE_NUMBER }}}_ph{{ PKG_HASH }}
```
jakirkham, line 18: Since we are building for two different variants of `has_cufile`, the package hash will change. However the rest of these values stay the same for each variant build. As a result the package name would be the same unless we add the package hash, which we do here

conda/recipes/libcucim/meta.yaml
```diff
@@ -15,13 +15,13 @@ source:
...
   ignore_run_exports:
     - openslide
   ignore_run_exports_from:
     - {{ compiler('cuda') }}
     - cuda-cudart-dev
-    - libcufile-dev  # [linux64]
+    - libcufile-dev
```
jakirkham, line 24: We want to ignore the `run_exports_from` `libcufile-dev` in all cases where it is used as we already constrain the version via `cuda-version`

This avoids overly tight version constraints that otherwise cause solver errors when testing CUDA 12.9 packages on CUDA 12.0

conda/recipes/libcucim/meta.yaml
```diff
@@ -60,9 +60,16 @@ requirements:
     - nvtx-c >=3.1.0
     - openslide
   run:
+    {% if aarch64 and cuda_major == "12" %}
+    # Relax `libcufile` dependency for old CUDA 12 ARM installs
+    - {{ pin_compatible('cuda-version', lower_bound='12.0', upper_bound='12.2') }}  # [not has_cufile]
+    - {{ pin_compatible('cuda-version', lower_bound='12.2', upper_bound='13.0') }}  # [has_cufile]
+    - libcufile                                                                     # [has_cufile]
+    {% else %}
```
jakirkham, line 68: This ensures ARM packages that support CUDA 12.2+ still depend io cuFile whereas packages that require CUDA 12.0 or 12.1 packages on ARM don't require cuFile (as it is not given in this case)

jakirkham (commented):
Added a few notes for context below

jameslamb (approved):
Thank you @jakirkham ! The changes and comments all make sense to me, I'd be happy to see this go in.

Once it does, we can do https://github.com/rapidsai/integration/issues/797 (adding the `cucim` dependency back to the `rapids` metapackage unconditionally)

bdice (approved):
Generally looks fine to me.

jakirkham:
Thanks all! 🙏

jakirkham:
Saw a flaky networking issue in [one CI job]( https://github.com/rapidsai/cucim/actions/runs/17837815258/job/50720918891?pr=930#step:11:256 ):

```
Get:22 http://ports.ubuntu.com/ubuntu-ports jammy-security/restricted arm64 Packages [5000 kB]
Fetched 38.1 MB in 32s (1181 kB/s)
Reading package lists...
E: Failed to fetch http://ports.ubuntu.com/ubuntu-ports/dists/jammy-security/main/binary-arm64/Packages.gz  File has unexpected size (3038653 != 3054887). Mirror sync in progress? [IP: 91.189.91.104 80]
   Hashes of expected file:
    - Filesize:3054887 [weak]
    - SHA256:04e8cacff94dccf44b0d14079f4f896262844739a764aa6333f42a0678c1195a
    - SHA1:447449b9caef6590e3ad69ce93872cebed2c302a [weak]
    - MD5Sum:451978ccf71438e21b7cbaf35ca1390d [weak]
   Release file created at: Thu, 18 Sep 2025 13:17:15 +0000
E: Some index files failed to download. They have been ignored, or old ones used instead.
```

Mentioned offline. Went ahead and restarted this job

PR #941, opened 2025-09-18
Add `pip` to `cucim`'s `requirements/host`

jakirkham:
The `cucim` Python package builds with `pip`

https://github.com/rapidsai/cucim/blob/42cf58f25b105dd2094cbd26c15ac2f498889211/conda/recipes/cucim/build.sh#L25

However the `pip` dependency was not listed explicitly in `requirements/host`. This fixes that issue


jakirkham:
Thanks Kyle! 🙏

PR #961, opened 2025-10-28
Require CUDA 12.2+

jakirkham:
Make CUDA 12.2 the new minimum. Drops logic related to CUDA 12.0 & 12.1. Thus simplifying the CUDA build and dependency logic.

<hr>

Part of issue:
* https://github.com/rapidsai/build-planning/issues/223

### DIFF
--- a/conda/recipes/libcucim/meta.yaml
+++ b/conda/recipes/libcucim/meta.yaml
@@ -60,15 +60,12 @@ requirements:
     - nvtx-c >=3.1.0
     - openslide
   run:
-    {% if aarch64 and cuda_major == "12" %}
-    # Relax `libcufile` dependency for old CUDA 12 ARM installs
-    - {{ pin_compatible('cuda-version', lower_bound='12.0', upper_bound='12.2') }}  # [not has_cufile]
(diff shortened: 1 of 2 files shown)


jakirkham:
The API doesn't break, but this is dropping support for some CUDA versions. So in that sense it is breaking

conda/recipes/libcucim/meta.yaml
```diff
@@ -61,14 +61,10 @@ requirements:
     - openslide
   run:
     {% if aarch64 and cuda_major == "12" %}
```
bdice: Do we need to retain the ARM specifier here? We want to require `cuda-version>=12.2,<13.0a0` for x64, too.

jakirkham: That's a good point. Wonder if there is a way we can write this to have no branch

jakirkham: The `aarch64` bit is now dropped

Perhaps there is a branch-less tweak that can be done separately

KyleFromNVIDIA (approved):

jakirkham:
Thanks all! 🙏

PR #967, opened 2025-11-07
Fix crash on unsigned integer input to `cucim.skimage.morphology.reconstruction`

### LINKED ISSUE #964: [BUG] crash in cucim.skimage.morphology.reconstruction for unsigned integer input
**Describe the bug**
Trying to run the skimage example [Filling Holes and Finding Peaks](https://scikit-image.org/docs/stable/auto_examples/features_detection/plot_holes_and_peaks.html#filling-holes-and-finding-peaks) adapted to the GPU results in a crash with

```
double free or corruption (out)
Aborted (core dumped)
```

A minimal reproducer is
```python
import cupy as cp
from skimage import data

from cucim.skimage.exposure import rescale_intensity
from cucim.skimage.morphology import reconstruction


image_host = data.moon()
image = cp.asarray(image_host)
image = rescale_intensity(image, in_range=(50, 200))

seed[1:-1, 1:-1] = image.max()
mask = image

filled = reconstruction(seed, mask, method='erosion')
```

**Expected behavior**
No crash should occur and results should be equivalent to the ones from scikit-image's CPU implementation

 - Environment location: Bare-metal (linux x86_64)
 - Method of cuCIM install: from source

grlee77:
resolves #964

This `reconstruction` function is a hybrid implementation that still relies on an inner Cython `reconstruction_loop` function from upstream scikit-image. I found that upgrading the input as done here avoids the observed crash and gives output that matches the scikit-image one.

### DIFF
--- a/python/cucim/src/cucim/skimage/morphology/tests/test_reconstruction.py
+++ b/python/cucim/src/cucim/skimage/morphology/tests/test_reconstruction.py
@@ -112,25 +112,29 @@ def test_erosion_uint8_input(unsigned_dtype):
     # Rescale image intensity so that we can see dim features.
     image = rescale_intensity(image, in_range=(50, 200))
 
-    seed = cp.copy(image)
-    seed[1:-1, 1:-1] = image.max()
+    image2 = cp.copy(image)
+    image2[1:-1, 1:-1] = image.max()
     mask = image
 
-    seed = seed.astype(unsigned_dtype)
-    filled = reconstruction(seed, mask, method="erosion")
+    image2 = image2.astype(unsigned_dtype)
+    image2_erosion = reconstruction(image2, mask, method="erosion")
 
-    expected_out_type = cp.promote_types(seed.dtype, cp.int8)
-    assert filled.dtype == expected_out_type
+    expected_out_type = cp.promote_types(image2.dtype, cp.int8)
+    assert image2_erosion.dtype == expected_out_type
     # promoted to signed dtype (or float in the case of cp.uint64)
-    assert filled.dtype.kind == "i" if unsigned_dtype != cp.uint64 else "f"
-    assert filled.dtype.itemsize >= seed.dtype.itemsize
+    assert (
+        image2_erosion.dtype.kind == "i" if unsigned_dtype != cp.uint64 else "f"
+    )
+    assert image2_erosion.dtype.itemsize >= image2.dtype.itemsize
 
     # compare to scikit-image CPU result
-    filled_cpu = reconstruction_cpu(
-        cp.asnumpy(seed), cp.asnumpy(mask), method="erosion"
+    image2_erosion_cpu = reconstruction_cpu(
+        cp.asnumpy(image2), cp.asnumpy(mask), method="erosion"
     )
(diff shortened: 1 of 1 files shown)


python/cucim/src/cucim/skimage/morphology/grayreconstruct.py
```diff
@@ -197,6 +201,12 @@ def reconstruction(seed, mask, method="dilation", footprint=None, offset=None):
     # CuPy Backend: modified to allow images_dtype based on input dtype
     #               instead of float64
     images_dtype = np.promote_types(seed.dtype, mask.dtype)
+    # For erosion, we need to negate the array, so ensure we use a signed type
+    # that can represent negative values without wraparound
+    if method == "erosion" and cp.issubdtype(images_dtype, cp.unsignedinteger):
+        # Promote unsigned types to signed types with sufficient range
+        images_dtype = cp.promote_types(images_dtype, cp.int8)
```
jakirkham: Should we use `np.promote_types` below as well?

```suggestion
    images_dtype = np.promote_types(seed.dtype, mask.dtype)
    # For erosion, we need to negate the array, so ensure we use a signed type
    # that can represent negative values without wraparound
    if method == "erosion" and cp.issubdtype(images_dtype, cp.unsignedinteger):
        # Promote unsigned types to signed types with sufficient range
        images_dtype = np.promote_types(images_dtype, cp.int8)
```

python/cucim/src/cucim/skimage/morphology/tests/test_reconstruction.py
```diff
@@ -102,6 +104,35 @@ def test_invalid_seed():
...
 
+@pytest.mark.parametrize(
+    "unsigned_dtype", [cp.uint8, cp.uint16, cp.uint32, cp.uint64]
+)
+def test_erosion_uint8_input(unsigned_dtype):
+    image = cp.asarray(data.moon())
+    # Rescale image intensity so that we can see dim features.
+    image = rescale_intensity(image, in_range=(50, 200))
+
+    seed = cp.copy(image)
```
jakirkham: `seed` could be confused with random number generator seed. Should we use a different name like `image2` or similar?

python/cucim/src/cucim/skimage/morphology/tests/test_reconstruction.py
```diff
@@ -102,6 +104,35 @@ def test_invalid_seed():
...
+    seed[1:-1, 1:-1] = image.max()
+    mask = image
+
+    seed = seed.astype(unsigned_dtype)
+    filled = reconstruction(seed, mask, method="erosion")
```
jakirkham: Maybe this could be `image2_eroded`?

python/cucim/src/cucim/skimage/morphology/tests/test_reconstruction.py
grlee77: sure, updated as suggested

python/cucim/src/cucim/skimage/morphology/tests/test_reconstruction.py
grlee77: updated

PR #968, opened 2025-11-08
add missing import of ellipse in morphology module

grlee77:
`cucim.skimage.morphology.footprints.ellipse` should also be available via `cucim.skimage.morphology.ellipse` like the other footprints


jakirkham (commented):
Thanks Greg! 🙏

Looks good

Should we test this is `import`able somehow? Either with its own test or tweaking an existing test to `import` from this location

grlee77:
> Should we test this is `import`able somehow? Either with its own test or tweaking an existing test to `import` from this location

Sure, updated the existing tests to also import from this location (some test cases still use via the footprint submodule as well)

PR #983, opened 2025-11-26
[WIP] PR #3 nvimagecodec v0.7.0

cdinea:
### Description

This PR updates the `cucim.kit.cuslide2` plugin to support nvImageCodec v0.7.0 from internal release, which includes new features for direct TIFF tag retrieval and API changes.

#### 1. CMake Configuration (`cmake/deps/nvimgcodec.cmake`)
- Updated default version from `0.6.0` to `0.7.0`
- Added `NVIMGCODEC_DIR` and `NVIMGCODEC_ROOT` variables for local installations
- Added automatic CUDA version detection (CUDA 12 vs CUDA 13)
- Improved library discovery for conda, pip, and system installations

#### 2. API Compatibility Fix (`nvimgcodec_decoder.cpp`)
- **Breaking change in v0.7.0**: Removed `buffer_size` assignment from `nvimgcodecImageInfo_t`
  - The `buffer_size` member was removed in v0.7.0; buffer size is now inferred from `plane_info`

#### 3. TIFF Tag Support (`nvimgcodec_tiff_parser.cpp`)
- **New in v0.7.0**: Added direct TIFF tag retrieval using `NVIMGCODEC_METADATA_KIND_TIFF_TAG`
- Implemented tag querying by ID for:
  - `SUBFILETYPE` (254) - for associated image classification
  - `COMPRESSION` (259) - compression method detection
  - `IMAGEDESCRIPTION` (270) - vendor metadata
  - `SUBIFD` (330) - sub-IFD references
  - `JPEGTABLES` (347) - JPEG quantization/huffman tables
- Updated enum values to match v0.7.0 API:
  ```cpp
  NVIMGCODEC_METADATA_KIND_TIFF_TAG = 1
  NVIMGCODEC_METADATA_KIND_MED_APERIO = 5
  NVIMGCODEC_METADATA_KIND_MED_PHILIPS = 6
  NVIMGCODEC_METADATA_KIND_MED_LEICA = 7
  NVIMGCODEC_METADATA_KIND_MED_VENTANA = 8
  NVIMGCODEC_METADATA_KIND_MED_TRESTLE = 9
  ```
- Added support for multiple TIFF value types (ASCII, SHORT, LONG, BYTE, LONG8, RATIONAL)

#### 4. Dependencies (`dependencies.yaml`)
- Updated `libnvimgcodec` and `nvimgcodec` versions to `0.7.0`

### Related
- nvImageCodec v0.7.0 release notes: Adds generic TIFF tag reading, removes `buffer_size` from `nvimgcodecImageInfo_t`
- Requires nvTIFF v0.6.0 for full functionality


cpp/cmake/deps/nvimgcodec.cmake
```diff
@@ -0,0 +1,232 @@
...
+# cmake-format: off
+# SPDX-FileCopyrightText: Copyright (c) 2020-2025, NVIDIA CORPORATION.
+# SPDX-License-Identifier: Apache-2.0
+# cmake-format: on
+#
+
+if (NOT TARGET deps::nvimgcodec)
+    # Option to automatically install nvImageCodec via conda
+    option(AUTO_INSTALL_NVIMGCODEC "Automatically install nvImageCodec via conda" ON)
+    set(NVIMGCODEC_VERSION "0.6.0" CACHE STRING "nvImageCodec version to install")
```
mkepa-nv: Shouldn't this say "0.7.0"? Maybe not right now, but when 0.7 is released

conda/environments/all_cuda-129_arch-aarch64.yaml
```diff
@@ -17,12 +17,15 @@ dependencies:
...
 - lazy-loader>=0.4
+- libnvimgcodec-dev=0.6.0
+- libnvimgcodec==0.6.0
 - libnvjpeg-dev
 - matplotlib-base>=3.7
 - nbsphinx
 - ninja
 - numpy>=1.23.4,<3.0a0
 - numpydoc>=1.7
+- nvimgcodec>=0.6.0
```
mkepa-nv: Shouldn't this say 0.7.0?

conda/environments/all_cuda-129_arch-aarch64.yaml
```diff
@@ -17,12 +17,15 @@ dependencies:
 - imagecodecs>=2021.6.8
 - ipython
 - lazy-loader>=0.4
+- libnvimgcodec-dev=0.6.0
+- libnvimgcodec==0.6.0
```
mkepa-nv: Are these two libs conda specific? I never heard about them (but I also don't use conda)

cpp/plugins/cucim.kit.cuslide2/src/cuslide/nvimgcodec/nvimgcodec_decoder.cpp
```diff
@@ -0,0 +1,425 @@
...
+        
+        // Select decoder based on target device
+        // CPU-only backend can handle in-bounds ROI decoding for TIFF files
+        std::string device_str = std::string(out_device);
+        bool target_is_cpu = (device_str.find("cpu") != std::string::npos);
+        
+        nvimgcodecDecoder_t decoder;
+        if (target_is_cpu && manager.has_cpu_decoder())
+        {
+            decoder = manager.get_cpu_decoder();
```
mkepa-nv: cpu decoder doesn't support out of bounds ROI decoding. Maybe you should add checks for this case? Otherwise decode will fail later on

cpp/plugins/cucim.kit.cuslide2/src/cuslide/nvimgcodec/nvimgcodec_tiff_parser.cpp
```diff
@@ -0,0 +1,1087 @@
...
+        // Extract metadata fields
+        int kind = metadata->kind;
+        int format = metadata->format;
+        size_t buffer_size = metadata->buffer_size;
+        const uint8_t* buffer = static_cast<const uint8_t*>(metadata->buffer);
+        
+        #ifdef DEBUG
+        // Map kind to human-readable name for debugging
+        // nvImageCodec 0.7.0 nvimgcodecMetadataKind_t enum values:
+        //   0 = UNKNOWN, 1 = TIFF_TAG, 2 = ICC_PROFILE, 3 = EXIF, 4 = GEO
```
mkepa-nv: I think this comment can be removed now, if you use enum values

cpp/plugins/cucim.kit.cuslide2/src/cuslide/nvimgcodec/nvimgcodec_tiff_parser.cpp
```diff
@@ -0,0 +1,1087 @@
...
+        if (buffer && buffer_size > 0)
+        {
+            IfdInfo::MetadataBlob blob;
+            blob.format = format;
+            blob.data.assign(buffer, buffer + buffer_size);
+            ifd_info.metadata_blobs[kind] = std::move(blob);
+            
+            // Special handling: extract ImageDescription if it's a text format
+            // nvImageCodec 0.7.0: Use proper enum values for medical formats
+            if (kind == NVIMGCODEC_METADATA_KIND_MED_APERIO && ifd_info.image_description.empty())
```
mkepa-nv, line 665: I would first if `ifd_info.image_description` is empty and then do `if else` for each possible kind. 

Another question: maybe it would be better to directly query for tiff tag ImageDescription instead of checking metadata kind?

cpp/plugins/cucim.kit.cuslide2/src/cuslide/nvimgcodec/nvimgcodec_tiff_parser.cpp
```diff
@@ -0,0 +1,1087 @@
...
+                // Show preview of XML
+                if (buffer_size > 0) {
+                    std::string preview(buffer, buffer + std::min(buffer_size, size_t(100)));
+                    fmt::print("     XML preview: {}...\n", preview);
+                }
+                #endif
+            }
+            else if (kind == NVIMGCODEC_METADATA_KIND_MED_LEICA && ifd_info.image_description.empty())
+            {
+                // WORKAROUND: nvImageCodec 0.6.0 sometimes misclassifies Aperio as Leica
```
mkepa-nv: Could you share example image when nvImageCodec does it? Maybe this was some bug with checking wrong metadata kind? We set it based on text inside aperio/leica metadata - if such name is present we set appropriate kind

cpp/plugins/cucim.kit.cuslide2/src/cuslide/nvimgcodec/nvimgcodec_tiff_parser.cpp
```diff
@@ -0,0 +1,1087 @@
...
+        
+        // Second call: retrieve actual value
+        status = nvimgcodecDecoderGetMetadata(
+            manager.get_decoder(),
+            ifd_info.sub_code_stream,
+            &metadata_ptr,
+            &metadata_count
+        );
+        
+        if (status != NVIMGCODEC_STATUS_SUCCESS || metadata.buffer_size == 0)
```
mkepa-nv: This is unexpected - maybe you should signal this error to the user?

cpp/plugins/cucim.kit.cuslide2/src/cuslide/nvimgcodec/nvimgcodec_tiff_parser.cpp
```diff
@@ -0,0 +1,1087 @@
...
+                
+            case NVIMGCODEC_METADATA_VALUE_TYPE_BYTE:
+                if (metadata.value_count == 1)
+                {
+                    tag_value = std::to_string(static_cast<int>(buffer[0]));
+                }
+                else
+                {
+                    // Binary data - store as hex or size info
+                    tag_value = fmt::format("[{} bytes]", metadata.buffer_size);
```
mkepa-nv: Maybe you should limit size as in SHORT and LONG case?

cpp/plugins/cucim.kit.cuslide2/src/cuslide/nvimgcodec/nvimgcodec_tiff_parser.cpp
```diff
@@ -0,0 +1,1087 @@
...
+                }
+                break;
+                
+            case NVIMGCODEC_METADATA_VALUE_TYPE_LONG8:
+                if (metadata.value_count == 1 && metadata.buffer_size >= sizeof(uint64_t))
```
mkepa-nv: What if there is more that one LONG8 value?

cpp/plugins/cucim.kit.cuslide2/src/cuslide/nvimgcodec/nvimgcodec_tiff_parser.cpp
```diff
@@ -0,0 +1,1087 @@
...
+                {
+                    uint64_t val = *reinterpret_cast<const uint64_t*>(buffer.data());
+                    tag_value = std::to_string(val);
+                }
+                break;
+                
+            case NVIMGCODEC_METADATA_VALUE_TYPE_RATIONAL:
+                if (metadata.buffer_size >= 8)
```
mkepa-nv: Can there be more than 1 rational value?

cpp/plugins/cucim.kit.cuslide2/src/cuslide/nvimgcodec/nvimgcodec_tiff_parser.cpp
```diff
@@ -0,0 +1,1087 @@
...
+                if (metadata.buffer_size <= 8 && metadata.value_count == 1)
+                {
+                    // Small value - try to interpret as number
+                    uint64_t val = 0;
+                    std::memcpy(&val, buffer.data(), std::min(metadata.buffer_size, sizeof(val)));
+                    tag_value = std::to_string(val);
+                }
+                else
+                {
+                    tag_value = fmt::format("[{} bytes, type={}]", metadata.buffer_size,
```
mkepa-nv: Should you limit max size as in other cases?

cpp/plugins/cucim.kit.cuslide2/src/cuslide/nvimgcodec/nvimgcodec_tiff_parser.cpp
cdinea: thank you for the feedback @mkepa-nv - i addressed this feedback(together with a few oother fixes) in[this](https://github.com/rapidsai/cucim/pull/983/commits/a172ecf1e64fbbe0a8e37896a7abf99ff4785a75) committ

cpp/plugins/cucim.kit.cuslide2/src/cuslide/nvimgcodec/nvimgcodec_tiff_parser.cpp
cdinea: thank you for the feedback @mkepa-nv - yes, we can simplify and remove redundant ImageDescription extraction from vendor metadata blobs - i addressed this feedback in [this](https://github.com/rapidsai/cucim/pull/983/commits/df6474073012218047d336a57e0deaf69a3e4aeb) commit

cpp/cmake/deps/nvimgcodec.cmake
cdinea: thank you @mkepa-nv - indeed, the version should be nvImageCodec v0.7.0 - although we are not using it in thiss version of the  PR, it is improtannt to update.I addressed this feedback in [this](https://github.com/rapidsai/cucim/pull/983/commits/4daa355645b81daba8d2047122e224db05ec2fe8) commit

conda/environments/all_cuda-129_arch-aarch64.yaml
cdinea: thank you @mkepa-nv -  nvImageCodec 0.7.0 is not yet on conda-forge, so I've removed these dependencies for now in [this](https://github.com/rapidsai/cucim/pull/983/commits/23e6d9a655ff4ec8fa19e0cc97ef725c9fd2eec1) commit.

conda/environments/all_cuda-129_arch-aarch64.yaml
cdinea: Yes, those were conda-forge specific packages (libnvimgcodec-dev for headers, libnvimgcodec for runtime library). I've removed them since nvImageCodec 0.7.0 isn't published to conda-forge yet. Based on my understanding,for pip users, the equivalent is nvidia-nvimgcodec-cu12 or nvidia-nvimgcodec-cu13

cpp/plugins/cucim.kit.cuslide2/src/cuslide/nvimgcodec/nvimgcodec_decoder.cpp
cdinea: Good catch @mkepa-nv ! Added a check for out-of-bounds ROI in [this](https://github.com/rapidsai/cucim/pull/983/commits/0de4b2176065aeb5cbe8bc64577179b4dc4eb17b) -when the requested region extends beyond image boundaries, the code now falls back to the hybrid decoder which supports this case.Let me know your thoughts on thsi fix

cpp/plugins/cucim.kit.cuslide2/src/cuslide/nvimgcodec/nvimgcodec_tiff_parser.cpp
cdinea: thank you @mkepa-nv -this WAR was for nvImageCodec 0.6.0 - I observed it with some Aperio SVS test files where the metadata kind was reported as MED_LEICA instead of MED_APERIO. However, I've since updated the code to use nvImageCodec 0.7.0's direct TIFF tag query API (via nvimgcodecTiffTagGetValue), which bypasses this issue entirely. The workaround code has been removed in the latest commits - we now extract ImageDescription directly from TIFF tag 270 rather than relying on the vendor metadata kind detection.
If you still want the specific test file to reproduce, I can try to locate it, but it may no longer be relevant with the v0.7.0 API.We can probably address this in PR#2 and understand the WAR

cpp/plugins/cucim.kit.cuslide2/src/cuslide/nvimgcodec/nvimgcodec_tiff_parser.cpp
cdinea: thank you for the the feedback @mkepa-nv - i've separated the error handling - API failures now log a warning (in DEBUG builds) with the tag ID and status code, while buffer_size == 0 (tag not found) is silently skipped since that's expected behavior for optional tags in [this](https://github.com/rapidsai/cucim/pull/983/commits/b5e29b5ddb66110e52fc5290beab07bb298647a8) commit

cpp/plugins/cucim.kit.cuslide2/src/cuslide/nvimgcodec/nvimgcodec_tiff_parser.cpp
cdinea: thank you for the suggestion @mkepa-nv - I added a 4KB size limit for binary metadata storage in [this](https://github.com/rapidsai/cucim/pull/983/commits/324d959fdf1440e6a7f318f90ce0532361ab915c) to prevent storing huge blobs. Truncation is logged in DEBUG builds. Do you think 4KB is enough?

cpp/plugins/cucim.kit.cuslide2/src/cuslide/nvimgcodec/nvimgcodec_tiff_parser.cpp
cdinea: arrays of LONG8 values are handled in the else if (metadata.value_count > 1) branch . When there are multiple values, they're stored as a std::vector<uint64_t>. I also added buffer size validation (std::min) to ensure we don't read beyond the buffer, matching how SHORT and LONG arrays are handled in [this](https://github.com/rapidsai/cucim/pull/983/commits/edf3ea987a947d91bfa50e7e9780102644971ac7) committ

cpp/plugins/cucim.kit.cuslide2/src/cuslide/nvimgcodec/nvimgcodec_tiff_parser.cpp
cdinea: yes, TIFF tags can have multiple RATIONAL values. I've added array support - when value_count > 1, the values are stored as comma-separated strings (e.g., '1/2, 3/4'). Same fix applied to SRATIONAL as well in [thiss](https://github.com/rapidsai/cucim/pull/983/commits/b0fb586c6438643a174f04a58b8076ae2044bbaf) commit

cpp/plugins/cucim.kit.cuslide2/src/cuslide/nvimgcodec/nvimgcodec_tiff_parser.cpp
cdinea: thank you for the feedback @mkepa-nv - I believe I already addressed this as the default case now limits binary data to 4KB (MAX_BINARY_SIZE = 4096) and logs truncation in DEBUG builds

conda/environments/all_cuda-129_arch-aarch64.yaml
mkepa-nv: Yes, you are right about pip. There is also `nvidia-nvimgcodec-tegra-cu12` for tegra (but no cuda 13 as there was unification of sbsa and tegra)

cpp/plugins/cucim.kit.cuslide2/src/cuslide/nvimgcodec/nvimgcodec_decoder.cpp
```diff
@@ -181,6 +185,18 @@ bool decode_ifd_region_nvimgcodec(const IfdInfo& ifd_info,
         std::string device_str = std::string(out_device);
         bool target_is_cpu = (device_str.find("cpu") != std::string::npos);
         
+        // Check if ROI is out of bounds (extends beyond image boundaries)
+        // CPU decoder doesn't support out-of-bounds ROI decoding, must use hybrid decoder
+        bool roi_out_of_bounds = (x + width > ifd_info.width) || (y + height > ifd_info.height);
```
mkepa-nv: `x` or `y` can be negative

cpp/plugins/cucim.kit.cuslide2/src/cuslide/nvimgcodec/nvimgcodec_tiff_parser.cpp
mkepa-nv: Sure, we can skip investigating this issue in this MR if it works correctly with 0.7

cpp/plugins/cucim.kit.cuslide2/src/cuslide/nvimgcodec/nvimgcodec_decoder.cpp
```diff
@@ -34,15 +32,21 @@ namespace cuslide2::nvimgcodec
 // RAII Helpers for nvImageCodec Resources
 // ============================================================================
 
-// RAII wrapper for nvimgcodecCodeStream_t
-struct CodeStreamDeleter
+// NOTE: Sub-code streams obtained via nvimgcodecCodeStreamGetSubCodeStream() are VIEWS
+// into the parent stream and share internal data. They should NOT be explicitly destroyed.
+// The parent stream destruction handles cleanup of all sub-streams.
+// Using nvimgcodecCodeStreamDestroy() on sub-streams causes "free(): invalid pointer".
```
mkepa-nv: This shouldn't happen, maybe there is a bug in some other place? You should be able to call `nvimgcodecCodeStreamDestroy` on parent or sub code stream in any order and all should be fine. Also each codeStream have some non shared state, so you need to call destroy for each of them. Destroying parent codeStream doesn't clean up sub code stream

I played around with creating and destroying CodeStreams locally and it all worked as expected. Could you investigate further what is causing the issue? And if this is confirmed to be on nvimgcodec side, could you create a minimal reproducer of the issue?

cpp/plugins/cucim.kit.cuslide2/src/cuslide/nvimgcodec/nvimgcodec_decoder.cpp
mkepa-nv: @cdinea check for roi is fine. But changes about destroying codeStream are not. I added a comment here: https://github.com/rapidsai/cucim/pull/983#discussion_r2567628449

cpp/plugins/cucim.kit.cuslide2/src/cuslide/nvimgcodec/nvimgcodec_manager.h
```diff
@@ -178,23 +178,37 @@ class NvImageCodecManager
 
     ~NvImageCodecManager()
     {
-        if (cpu_decoder_)
+        // Use try-catch to prevent segfaults during static destruction
```
mkepa-nv: @cdinea I don't think this work as you expect. Segfaults are not exceptions, so try catch will not work for them. Are you sure this work?

cpp/plugins/cucim.kit.cuslide2/src/cuslide/nvimgcodec/nvimgcodec_tiff_parser.cpp
```diff
@@ -217,35 +308,40 @@ TiffFileParser::TiffFileParser(const std::string& file_path)
...
-    }
+    // NOTE: We intentionally do NOT destroy main_code_stream_ or sub_code_streams here.
+    //
+    // Reason: Static destruction order is undefined at program exit. The singleton
+    // (NvImageCodecTiffParserManager) that owns the nvimgcodec instance might be
+    // destroyed BEFORE this TiffFileParser. If we try to call nvimgcodecCodeStreamDestroy()
+    // on a code stream whose parent instance is already destroyed, we get
+    // "free(): invalid pointer".
+    //
+    // Solution: Let nvimgcodec handle cleanup. When nvimgcodecInstanceDestroy() is called
```
mkepa-nv: @cdinea  This is not true. `nvimgcodecInstanceDestroy()` doesn't cleanup resources allocated by other objects (like CodeStream or Decoder). Did we write somewhere in docs that it does? Where did you find this info?

cpp/plugins/cucim.kit.cuslide2/src/cuslide/nvimgcodec/nvimgcodec_tiff_parser.cpp
```diff
@@ -0,0 +1,1361 @@
...
+                if (!str_val.empty())
+                {
+                    tag_value = std::move(str_val);
+                    value_stored = true;
+                }
+                break;
+            }
+                
+            case NVIMGCODEC_METADATA_VALUE_TYPE_SHORT:
+                if (metadata.value_count == 1 && metadata.buffer_size >= sizeof(uint16_t))
```
mkepa-nv: This patter is repeated a lot. I would extract it into function templated on type. Also you don't need to check for buffer_size, count should be enough

cpp/plugins/cucim.kit.cuslide2/src/cuslide/nvimgcodec/nvimgcodec_tiff_parser.cpp
```diff
@@ -0,0 +1,1361 @@
...
+                {
+                    uint16_t val = *reinterpret_cast<const uint16_t*>(buffer.data());
+                    tag_value = val;
+                    value_stored = true;
```
mkepa-nv: What is the definition of `TiffTagValue`? Maybe you can add `std::monostate` to it, which would represent empty state

cpp/plugins/cucim.kit.cuslide2/src/cuslide/nvimgcodec/nvimgcodec_tiff_parser.cpp
```diff
@@ -0,0 +1,1361 @@
...
+                    double val = *reinterpret_cast<const double*>(buffer.data());
+                    tag_value = val;
+                    value_stored = true;
+                }
+                else if (metadata.value_count > 1)
+                {
+                    // Array of doubles - store as string representation
+                    const double* vals = reinterpret_cast<const double*>(buffer.data());
+                    std::string result;
+                    for (uint32_t i = 0; i < metadata.value_count && i < 10; ++i)
```
mkepa-nv: I am a bit confused. Why do we store float array as string, while we can store int array of single float value directly?

cpp/plugins/cucim.kit.cuslide2/src/cuslide/nvimgcodec/nvimgcodec_tiff_parser.cpp
```diff
@@ -0,0 +1,1361 @@
...
+                        tag_value = fmt::format("{}/{}", num, den);
+                    else
+                        tag_value = std::to_string(num);
+                    value_stored = true;
+                }
+                else if (metadata.value_count > 1)
+                {
+                    // Array of Rationals - store as comma-separated string
+                    size_t rational_size = 8;  // 2 × uint32_t
+                    size_t count = std::min(static_cast<size_t>(metadata.value_count),
```
mkepa-nv: What about `metadata.value_count`? Does it contain wrong value?

cpp/plugins/cucim.kit.cuslide2/src/cuslide/nvimgcodec/nvimgcodec_tiff_parser.cpp
mkepa-nv: Oh, I must have missed the else case, sorry about that

cpp/plugins/cucim.kit.cuslide2/src/cuslide/nvimgcodec/nvimgcodec_tiff_parser.cpp
```diff
@@ -0,0 +1,1361 @@
...
+                {
+                    uint64_t val = *reinterpret_cast<const uint64_t*>(buffer.data());
+                    tag_value = val;
+                    value_stored = true;
+                }
+                else if (metadata.value_count > 1)
+                {
+                    // Array of 64-bit values - validate count against buffer size
+                    const uint64_t* vals = reinterpret_cast<const uint64_t*>(buffer.data());
+                    size_t count = std::min(static_cast<size_t>(metadata.value_count),
```
mkepa-nv: Why are you not using `metada.value_count`?

cpp/plugins/cucim.kit.cuslide2/src/cuslide/nvimgcodec/nvimgcodec_tiff_parser.cpp
```diff
@@ -0,0 +1,1361 @@
...
+                    if (metadata.buffer_size <= 8 && metadata.value_count == 1)
+                    {
+                        // Small value - try to interpret as number and store as string
+                        uint64_t val = 0;
+                        std::memcpy(&val, buffer.data(), std::min(metadata.buffer_size, sizeof(val)));
+                        tag_value = std::to_string(val);
+                    }
+                    else
+                    {
+                        // Store raw bytes - limit size to prevent storing huge blobs
```
mkepa-nv: What if use would need the whole buffer? And right now the limits it somehow arbitrary when it comes to types - e.g. SHORT tags are not limited

cpp/plugins/cucim.kit.cuslide2/src/cuslide/nvimgcodec/nvimgcodec_tiff_parser.cpp
```diff
@@ -0,0 +1,1361 @@
...
+    if (ext == ".svs" || ext == ".ndpi" || ext == ".vms" || ext == ".vmu")
+    {
+        ifd_info.tiff_tags["COMPRESSION"] = static_cast<uint16_t>(7);  // TIFF_COMPRESSION_JPEG
+        #ifdef DEBUG
+        fmt::print("  ✅ Inferred JPEG compression (WSI format: {})\n", ext);
+        #endif // DEBUG
+    }
+    
+    // Store ImageDescription if available from vendor metadata
+    if (!ifd_info.image_description.empty())
```
mkepa-nv, line 793: Didn't you remove code that sets it based on vendor metadata in a recent commit?

cpp/plugins/cucim.kit.cuslide2/src/cuslide/nvimgcodec/nvimgcodec_tiff_parser.cpp
```diff
@@ -0,0 +1,1361 @@
...
+            bool compression_inferred = false;
+            
+            // Check if we have TIFF Compression tag (stored as typed value)
+            auto compression_it = ifd_info.tiff_tags.find("COMPRESSION");
+            if (compression_it != ifd_info.tiff_tags.end())
+            {
+                try
+                {
+                    // Get compression value from typed variant
+                    uint16_t compression_value = tiff_tag_value_to_uint16(compression_it->second);
```
mkepa-nv: Compression tag should be SHORT right? So maybe it is better to just check if that is the case and throw exception in the unlikely case that it is not? And we can get rid of this whole `tiff_tag_value_to_uint16` function

cpp/plugins/cucim.kit.cuslide2/src/cuslide/nvimgcodec/nvimgcodec_tiff_parser.cpp
mkepa-nv: Yes, but for SHORT array you keep all elements - so why you limit here?

cpp/plugins/cucim.kit.cuslide2/src/cuslide/nvimgcodec/nvimgcodec_tiff_parser.cpp
mkepa-nv: I don't know, I didn't work that close with tiff. What is the reason to limit metadata in the first place?

cpp/plugins/cucim.kit.cuslide2/src/cuslide/nvimgcodec/nvimgcodec_decoder.cpp
cdinea: thank you for the clarification  @mkepa-nv -  I implemented proper sub-code stream destruction as you suggested in [this](https://github.com/rapidsai/cucim/pull/983/commits/a55b1cf360c845ad4fc68f767fe3daa735a8ef34) commit 
 1. ROI sub-code streams in nvimgcodec_decoder.cpp are now explicitly destroyed after each decode
 2. IFD sub-code streams are destroyed in TiffFileParser::~TiffFileParser()
3. Singleton destructor properly calls nvimgcodecDecoderDestroy() and nvimgcodecInstanceDestroy()

The problem i am seeing is when i test the aperio svs files, I run into "free(): invalid pointer" crash at exit.

For reproducibilitty,  I've created a standalone reproducer (nvimgcodec_substream_crash_reproducer.cpp) that demonstrates the issue:

```

 // Pattern that crashes:
 nvimgcodecCodeStreamCreateFromFile(instance, &main_stream, file, ...);
 nvimgcodecCodeStreamGetSubCodeStream(main_stream, &sub_stream, &view);
 nvimgcodecDecoderDecode(decoder, &sub_stream, &image, 1, &params, &future);
 // ... decode succeeds ...
 nvimgcodecCodeStreamDestroy(sub_stream); // <-- CRASH: "free(): invalid pointer"
```


Environment:
-nvImageCodec v0.7.0 (internal build 11)
- Linux x86_64, CUDA 12.x
- Test file: Aperio SVS (CMU-1-Small-Region.svs)

 The reproducer is included in the PR. Would you be able to test it on your side to see if you can reproduce the crash? Happy to provide any additional information that would help debug this.

cpp/plugins/cucim.kit.cuslide2/src/cuslide/nvimgcodec/nvimgcodec_manager.h
cdinea: yes @mkepa-nv , you are right -yhis code has been removed in [this](https://github.com/rapidsai/cucim/pull/983/commits/ef123810a867a39575e229e86cfa03159adff660) commit - the file nvimgcodec_manager.h has been deleted, and the singleton manager is now in nvimgcodec_tiff_parser.cpp with proper cleanup (no try-catch blocks)

cpp/plugins/cucim.kit.cuslide2/src/cuslide/nvimgcodec/nvimgcodec_tiff_parser.cpp
cdinea: thank you @mkepa-nv -that comment was based on an incorrect assumption. I've since learned (through testing and your feedback) that nvimgcodecInstanceDestroy() does not automatically clean up code streams or decoders- this incorrect comment has been removed in [this](https://github.com/rapidsai/cucim/pull/983/commits/a55b1cf360c845ad4fc68f767fe3daa735a8ef34) commit,currently we  destroy all resources explicitly:
 - sub-code streams destroyed in TiffFileParser::~TiffFileParser()
- decoders destroyed in NvImageCodecTiffParserManager::~NvImageCodecTiffParserManager()
- instance destroyed last

cpp/plugins/cucim.kit.cuslide2/src/cuslide/nvimgcodec/nvimgcodec_tiff_parser.cpp
cdinea: this is a great idea @mkepa-nv - i added a template function in [this](https://github.com/rapidsai/cucim/pull/983/commits/5a7d785b85a59f7c2497dbc01cc2193704ceb6f1) commit

cpp/plugins/cucim.kit.cuslide2/src/cuslide/nvimgcodec/nvimgcodec_tiff_parser.cpp
cdinea: thank you for the suggestion @mkepa-nv .Re:TiffTagValue definition,it's a std::variant that stores typed TIFF tag values. Previously it was:

```
 using TiffTagValue = std::variant<
 std::string, // ASCII strings
 int8_t, uint8_t, // SBYTE/BYTE
 int16_t, uint16_t, // SSHORT/SHORT
 int32_t, uint32_t, // SLONG/LONG
 int64_t, uint64_t, // SLONG8/LONG8 (BigTIFF)
 float, double, // FLOAT/DOUBLE
 std::vector<uint8_t>, // Binary/arrays
 std::vector<uint16_t>, std::vector<uint32_t>, std::vector<uint64_t>
```

I've added std::monostate as you suggested in [this](https://github.com/rapidsai/cucim/pull/983/commits/e990bb482e18a2cebc5c1c91d0189bb5f05dd2e0) commit 
```
 using TiffTagValue = std::variant<
 std::monostate, // Empty/unset state - NEW!
 std::string,
 // ... rest of types
```

cpp/plugins/cucim.kit.cuslide2/src/cuslide/nvimgcodec/nvimgcodec_tiff_parser.cpp
cdinea: thank you for the feedback @mkepa-nv - you're right - this was inconsistent. Float and double arrays were being stored as comma-separated strings while integer arrays were properly stored as typed vectorss. I've fixed it to store float/double arrays as std::vector<float> and std::vector<double> instead, matching the behavior of integer arrays. This provides consistent typed storage across all TIFF data types. Fixed in [this](https://github.com/rapidsai/cucim/pull/983/commits/c0e854f1cbd5e3a50124b1a3f89d4b63697c69b4) commit

cpp/plugins/cucim.kit.cuslide2/src/cuslide/nvimgcodec/nvimgcodec_tiff_parser.cpp
cdinea: thank you for the feedback @mkepa-nv , that defensive check was unnecessary and inconsistent. I've removed the std::min() calculations for RATIONAL and SRATIONAL arrays to match how all other array types are handled (trusting metadata.value_count directly). Fixed in [this](https://github.com/rapidsai/cucim/pull/983/commits/4921b6e3b14feea25d28cfd90cf77b63d6ca2bf2) coommit

cpp/plugins/cucim.kit.cuslide2/src/cuslide/nvimgcodec/nvimgcodec_tiff_parser.cpp
cdinea: this has already been addressed in the current version. The LONG8/IFD8 arrays now use the extract_value_array<uint64_t>() template which trusts metadata.value_count directly

cpp/plugins/cucim.kit.cuslide2/src/cuslide/nvimgcodec/nvimgcodec_tiff_parser.cpp
cdinea: good point @mkepa-nv - I've made it configurable with unlimited as the default (matching typed array behavior). Users can call set_max_binary_tag_size() if they need to limit memory usage. Fixed in [this](https://github.com/rapidsai/cucim/pull/983/commits/6991a95e826c26e0e0b54909c480ed68b1ffae57) commit

cpp/plugins/cucim.kit.cuslide2/src/cuslide/nvimgcodec/nvimgcodec_tiff_parser.cpp
cdinea: thank you for the feedback @mkepa-nv -image_description is now only set from tiff_tags, so it would always be empty in that fallback path. Removed in [this](https://github.com/rapidsai/cucim/pull/983/commits/6dd0d47de5761b24be7b5fd63f35399a6fca8196) commit

cpp/plugins/cucim.kit.cuslide2/src/cuslide/nvimgcodec/nvimgcodec_tiff_parser.cpp
cdinea: i think youu are right @mkepa-nv .the COMPRESSION tag is always SHORT per TIFF spec. I've removed the conversion helper and switched to std::get<uint16_t>() directly in [this](https://github.com/rapidsai/cucim/pull/983/commits/487852afc33312b5f1bc1622d706eea2d73fff5a) commit

cpp/plugins/cucim.kit.cuslide2/src/cuslide/nvimgcodec/nvimgcodec_tiff_parser.cpp
```diff
@@ -0,0 +1,1255 @@
...
+        
+        if (metadata.buffer_size == 0)
+        {
+            // Unexpected: buffer was allocated but size is now 0
+            continue;
+        }
+        
+        // Convert value based on type and store as typed variant
+        TiffTagValue tag_value;
+        bool value_stored = false;
```
mkepa-nv: `value_stored` is not needed. The variant will be initialized into first type from its list - so `std::monotstate`. We can just check at the end if type is still monostate with either:
- `tag_value.index() == 0`
- `std::holds_alternative<std::monotstate>(tag_value))`

cpp/plugins/cucim.kit.cuslide2/src/cuslide/nvimgcodec/nvimgcodec_tiff_parser.cpp
mkepa-nv: @cdinea If you added monostate you can get rid of `value_stored`, here is my idea: https://github.com/rapidsai/cucim/pull/983#discussion_r2571161522

cpp/plugins/cucim.kit.cuslide2/src/cuslide/nvimgcodec/nvimgcodec_tiff_parser.cpp
```diff
@@ -523,8 +486,9 @@ void TiffFileParser::parse_tiff_structure()
             {
                 try
                 {
-                    // Get compression value from typed variant
-                    uint16_t compression_value = tiff_tag_value_to_uint16(compression_it->second);
+                    // COMPRESSION tag is always SHORT (uint16_t) per TIFF spec
+                    // If it's not, that indicates a bug in nvImageCodec or our parsing
+                    uint16_t compression_value = std::get<uint16_t>(compression_it->second);
```
mkepa-nv: Maybe it would be better to directly check if type is correct via https://en.cppreference.com/w/cpp/utility/variant/holds_alternative.html instead of relying on exceptions? But I don't have strong opinions on that, we can leave catch too

cpp/plugins/cucim.kit.cuslide2/src/cuslide/nvimgcodec/nvimgcodec_tiff_parser.cpp
cdinea: thank you for the feedback @mkepa-nv - I addressed this feedback in [this](https://github.com/rapidsai/cucim/pull/983/commits/4a0225c7fdfc67ecb027dd21479edf3f45ef9d0b) commit

cpp/plugins/cucim.kit.cuslide2/src/cuslide/nvimgcodec/nvimgcodec_tiff_parser.cpp
cdinea: thank you for the suggestion @mkepa-nv - this is a good idea and i addressed it in [this](https://github.com/rapidsai/cucim/pull/983/commits/2716b26675d8bcd7cb325692490ce263f2f77220) committ

cdinea:
@KyleFromNVIDIA @jakirkham  - The CI is experiencing persistent conda-forge network timeouts. 
Could this be a CI runner network/configuration issue?

conda/recipes/libcucim/meta.yaml
```diff
@@ -69,8 +70,10 @@ requirements:
     - libcufile
     - cuda-cudart
     - libnvjpeg
+    - libnvimgcodec0 {{ nvimgcodec_version }}  # nvImageCodec runtime library
   run_constrained:
     - {{ pin_compatible('openslide') }}
+    - libnvimgcodec-dev {{ nvimgcodec_version }}  # Optional: for development/debugging
```
carterbox, line 77:
```suggestion
    - libnvjpeg
  run_constrained:
    - {{ pin_compatible('openslide') }}
```
These constraints aren't necessary. `libnvimgcodec-dev` has appropriate run_exports, and your constraints are less strict than the run_exports.

VERSION
```diff
@@ -1 +1 @@
-26.02.00
+25.12.00
```
jakirkham: It looks like the version was reverted here

Think this is due to some rebasing issues. Could we either...

1. Squash all commits and rebase
2. Rebase and remove commits related to 25.12

branchremote/cucim
jakirkham, line 1: It seems a git submodule snuck in. Could we remove this?

conda/environments/all_cuda-129_arch-aarch64.yaml
```diff
@@ -17,8 +17,12 @@ dependencies:
 - imagecodecs>=2021.6.8
 - ipython
 - lazy-loader>=0.4
+<<<<<<< HEAD
 - libnvimgcodec-dev>=0.6.0,<0.7.0
 - libnvimgcodec>=0.6.0,<0.7.0
+=======
+- libnvimgcodec-dev=0.6.0
+>>>>>>> d3e322e (feat: Add comprehensive nvImageCodec installation verification script)
```
jakirkham: Looks like some merge conflicts were committed. Fortunately these files are autogenerated

So to fix this, would just run [`rapids-dependency-file-generator`]( https://github.com/rapidsai/dependency-file-generator ) and commit the changes

cpp/plugins/cucim.kit.cuslide/cmake/deps/libjpeg-turbo.cmake
```diff
@@ -29,7 +29,7 @@ if (NOT TARGET deps::libjpeg-turbo)
     # Tell CMake where to find the compiler by setting either the environment
     # variable "ASM_NASM" or the CMake cache entry CMAKE_ASM_NASM_COMPILER to the
     # full path to the compiler, or to the compiler name if it is in the PATH.
-    # yasm is available through `sudo apt-get install yasm` on Debian Linux.
+    # nasm is available through `sudo apt-get install nasm` on Debian Linux.
```
jakirkham, line 32: Just checking, did we mean to change this?

cuslide2_build_env.yaml
```diff
@@ -0,0 +1,56 @@
+name: cuslide2-build
```
jakirkham: Am curious if this is still needed given `libnvimgcodec` is in the `conda/environments` files which are updated automatically

cuslide2_build_env.yaml
```diff
@@ -0,0 +1,56 @@
...
+  
+  # CUDA
+  - cuda-version=12.9
+  - cuda-nvcc
+  - cuda-cudart-dev
+  - libnvjpeg-dev
+  
+  # nvImageCodec - critical for cuslide2
+  - libnvimgcodec-dev=0.6.0
+  - libnvimgcodec0=0.6.0
```
jakirkham: This is referring to `libnvimgcodec` version `0.6.0`, but we are upgrading to `0.7.0`

If we don't need this file, dropping it will fix it

If we do need this file, this would need to be updated, but let's discuss whether we can consolidate it first

dependencies.yaml
```diff
@@ -100,6 +100,7 @@ dependencies:
         packages:
           - cmake>=3.30.4
           - ninja
+          - libnvimgcodec-dev=0.6.0
```
jakirkham:
```suggestion
          - libnvimgcodec-dev=0.7.0
```

jakirkham (commented):
Thanks Cristiana! 🙏

It looks like there were some issues from the rebase. Let's try to clean that up first. That likely fixes most of the issues seen here

We can then take a closer look at the remainder

branchremote/cucim
cdinea: yes,thank you for flagging this @jakirkham

conda/environments/all_cuda-129_arch-aarch64.yaml
cdinea: thank you @jakirkham - indeed, the branch has some rebase commits, I pushed a commit to address this

cdinea:
this PR is closed in favour of PR #997

PR #990, opened 2025-12-07
resolve static analysis complaints about potentially unsafe C-style string-handling

grlee77:
This PR is intended to resolve a few issues marked as high severity by static analysis checks. In practice, none of these seems like a true issue in practice:

1. in places where `strlen` was used, currently the values passed in are always from string literals in the existing `cumed` and `cuslide` plugins, so it is not possible that the pointer is null or the C-style string is not null-terminated. This PR adds logic to handle the non null-terminated `string_view` case via a copy.

Examples of where the plugins set these via string literals:
https://github.com/rapidsai/cucim/blob/f328b1f1bc7950972b89bf0f65064875ef759334/cpp/plugins/cucim.kit.cuslide/src/cuslide/cuslide.cpp#L175-L203
https://github.com/rapidsai/cucim/blob/f328b1f1bc7950972b89bf0f65064875ef759334/cpp/plugins/cucim.kit.cumed/src/cumed/cumed.cpp#L142
https://github.com/rapidsai/cucim/blob/f328b1f1bc7950972b89bf0f65064875ef759334/cpp/plugins/cucim.kit.cuslide/src/cuslide/cuslide.cpp#L215

2. For the benchmarks, a check for `malloc` failure was added, although this is highly unlikely to occur in practice for the small amounts of memory involved in the benchmarks.

### DIFF
--- a/benchmarks/primitives.cpp
+++ b/benchmarks/primitives.cpp
@@ -1,5 +1,5 @@
 /*
- * SPDX-FileCopyrightText: Copyright (c) 2020-2021, NVIDIA CORPORATION.
+ * SPDX-FileCopyrightText: Copyright (c) 2020-2025, NVIDIA CORPORATION.
  * SPDX-License-Identifier: Apache-2.0
  */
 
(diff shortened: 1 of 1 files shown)


jakirkham:
Thanks Greg! 🙏

In discussion with @caryr35 offline, we want to include this in 25.12. Have changed the branch this PR is targeting accordingly

Greg, can you please rebase to clear out unrelated commits?

PR #992, opened 2025-12-09
Use static linkage for CUDA runtime

bdice:
## Summary
- Remove `cuda-cudart` from run requirements in conda recipes

With static linking of the CUDA runtime (the default for C++ and Python packages), the runtime is embedded in the binaries and the `cuda-cudart` package is not needed at runtime.

Part of https://github.com/rapidsai/build-planning/issues/235


jakirkham:
Thanks Bradley! 🙏

The commit history is pulling in a bunch of old commits from past releases. Could we please rebase this?

robertmaynard (changes requested):
Locations we directly link to shared cudart:

```
cpp/CMakeLists.txt:            CUDA::cudart
cpp/plugins/cucim.kit.cumed/CMakeLists.txt:            CUDA::cudart
cpp/plugins/cucim.kit.cuslide/CMakeLists.txt:                CUDA::cudart
cpp/plugins/cucim.kit.cuslide/CMakeLists.txt:                CUDA::cudart
cpp/plugins/cucim.kit.cuslide/tests/CMakeLists.txt:            CUDA::cudart
cpp/plugins/cucim.kit.cuslide2/CMakeLists.txt:                CUDA::cudart
cpp/plugins/cucim.kit.cuslide2/CMakeLists.txt:                CUDA::cudart
cpp/plugins/cucim.kit.cuslide2/tests/CMakeLists.txt:            CUDA::cudart
cpp/tests/CMakeLists.txt:            CUDA::cudart
```

robertmaynard (approved):

conda/recipes/cucim/meta.yaml
jakirkham, line 54: This needs to be dropped too

```suggestion
```

conda/recipes/libcucim/meta.yaml
jakirkham, line 59: Same here

```suggestion
```

jakirkham (dismissed):
Thanks Kyle! 🙏

This is looking better, but still needs a few things

conda/recipes/cucim/meta.yaml
KyleFromNVIDIA: No, we're still compiling against cudart, so we still need it as a `host` dependency. We've already removed it as a `run` dependency since we've statically linked it.

jakirkham:
Also please drop any other instances of `cuda-cudart-dev` like this one

https://github.com/rapidsai/cucim/blob/2129c3bc4573d4b8707669f22b09e7de070fb888/dependencies.yaml#L218

KyleFromNVIDIA:
We're still including `cuda-cudart-dev` in the `cuda` dependency set because it's still needed to build. We just don't include it in the runtime dependencies anymore.

conda/recipes/cucim/meta.yaml
jakirkham: For static linkage this is already part of the NVCC compiler package

KyleFromNVIDIA: Depending on the NVCC Conda package for cudart is wrong. NVCC lives in `$BUILD_PREFIX` (the build environment), since that's where the executable for the building architecture lives, while we need to get cudart from `$PREFIX` (the host environment), since that's where the library for the built architecture lives.

This isn't just a theoretical cross-compilation problem. We have previously had real issues with this. See https://github.com/rapidsai/cudf/pull/22290/changes#r3184782324 - we had to add `cuda-cudart-dev` back as a `host` dependency because the location of compute-sanitizer was being computed relative to cudart.

The bottom line is that if we're going to build against cudart, we need to have a dependency on `cuda-cudart-dev` at build time.

bdice (commented):
Thanks @KyleFromNVIDIA, this looks fine to me. I agree with your statement about still needing `cuda-cudart-dev` in `host`.

KyleFromNVIDIA (approved):

KyleFromNVIDIA:
Given that both @bdice and @robertmaynard have given their approval, I'm going to go ahead and merge this. @jakirkham I think I've adequately addressed your concerns. If you still feel there is more to be addressed, we can do that in a follow-up.

PR #997, opened 2025-12-17
[REVIEW] nvImageCodec v0.7.0 infra only

cdinea:
- Pinned nvimgcodec/nvimagecodec deps to 0.7.0 in deps/conda/pyproject

- Made cuslide2 plugin a MODULE + add linkable core lib for tests

- Avoided global include_directories; use target_include_directories

PR #998, opened 2025-12-19
[REVIEW] nvImageCodec v0.7.0 decoding

cdinea:
### Description

This PR upgrades the `cuslide2` plugin from nvImageCodec v0.6.0 to **v0.7.0**, bringing support for the new metadata API that enables direct TIFF tag querying. This is a follow-up to PR #997 which introduced the initial infrastructure.

### Motivation and Context

nvImageCodec v0.7.0 introduces a significant improvement in its TIFF metadata handling capabilities through the new `NVIMGCODEC_METADATA_KIND_TIFF_TAG` API. This allows cuCIM to:

- Query individual TIFF tags directly (e.g., `COMPRESSION`, `SUBFILETYPE`, `IMAGEDESCRIPTION`, `JPEGTABLES`)
- Retrieve typed values instead of relying on file extension heuristics
- Better support for multi-vendor whole-slide image (WSI) formats like Aperio SVS and Philips TIFF

#### Dependencies
- **Updated nvimgcodec constraints to v0.7.0+** across all build configurations:
  - `libnvimgcodec-dev>=0.7.0,<0.8.0` for conda builds
  - `nvidia-nvimgcodec-cu12>=0.7.0,<0.8.0` / `nvidia-nvimgcodec-cu13>=0.7.0,<0.8.0` for pip wheels
  - Updated all conda environment YAML files (CUDA 12.9/13.1, x86_64/aarch64)
  - Updated `conda/recipes/libcucim/conda_build_config.yaml`

#### cuslide2 Plugin Enhancements

**TIFF Parser (`nvimgcodec_tiff_parser.cpp/.h`):**
- Added compile-time detection macro `CUSLIDE2_NVIMGCODEC_HAS_TIFF_TAG_METADATA` for API compatibility
- Implemented `TiffTagValue` variant type for storing typed TIFF tag values (strings, integers, floats, byte arrays)
- Added helper functions `tiff_tag_value_to_string()`, `extract_single_value<T>()`, `extract_value_array<T>()` for type-safe value extraction
- Enhanced `IfdInfo` struct with new `tiff_tags` unordered_map for per-tag storage
- Updated `extract_tiff_tags()` to use nvImageCodec's native tag API when available
- Removed debug `fprintf` statements from production code

**CMake Build System (`CMakeLists.txt`):**
- Fixed include directory propagation for nvimgcodec headers (`SYSTEM PUBLIC` instead of directory-scoped)
- Improved comments and documentation for dynamic loading configuration

**Header Cleanup:**
- Added extended `IfdInfo` struct with metadata blob and TIFF tag storage
- Added `set_max_binary_tag_size()` / `get_max_binary_tag_size()` configuration methods
- Maintained backward compatibility stubs when `CUCIM_HAS_NVIMGCODEC` is not defined

### Testing

- Existing cuslide2 unit tests pass with the upgraded library
- Tested with Aperio SVS and Philips TIFF formats

### Breaking Changes

None. The API remains backward compatible. The implementation gracefully falls back to file extension heuristics when running with older nvImageCodec versions.

### Checklist

- [x] Use looser nvimgcodec constraints (`>=0.7.0,<0.8.0`) per reviewer feedback
- [x] Add compile-time detection for new metadata API
- [x] Implement typed TIFF tag extraction
- [x] Fix include propagation in CMake
- [x] Remove debug print statements
- [x] Update all conda environment files
- [x] Update libcucim conda recipe
- [x] Maintain backward compatibility with v0.6.0 fallback
- [x] Update copyright years to 2026

### Related Issues/PRs

- Follows #997: nvImageCodec infrastructure PR

### DIFF
--- a/cpp/plugins/cucim.kit.cuslide2/src/cuslide/nvimgcodec/nvimgcodec_tiff_parser.cpp
+++ b/cpp/plugins/cucim.kit.cuslide2/src/cuslide/nvimgcodec/nvimgcodec_tiff_parser.cpp
@@ -970,7 +970,7 @@ void TiffFileParser::extract_tiff_tags(IfdInfo& ifd_info)
         std::transform(ext.begin(), ext.end(), ext.begin(), ::tolower);
     }
 
-    if (ext == ".svs" || ext == ".ndpi" || ext == ".vms" || ext == ".vmu")
+    if (ext == ".svs" || ext == ".scn" || ext == ".ndpi" || ext == ".bif")
     {
         ifd_info.tiff_tags["COMPRESSION"] = static_cast<uint16_t>(7);  // TIFF_COMPRESSION_JPEG
     }


dependencies.yaml
```diff
@@ -161,7 +161,7 @@ dependencies:
         packages:
           - cuda-cudart-dev
           - libnvjpeg-dev
-          - libnvimgcodec-dev>=0.6.0,<0.7.0
+          - libnvimgcodec-dev==0.7.0
```
jameslamb:
```suggestion
          - libnvimgcodec-dev>=0.7.0,<0.8.0
```

Is it necessary to pin so tightly (`==0.7.0` here and `==0.7.0.11` further down in the file)? Generally I think we'd prefer to have looser constraints... that reduces the risk of conflicts when `cucim` is installed alongside other software.

Would you consider something like this? (here and in every other place touched in this file)

cdinea:
> I see that this has `[WIP]` but since it was opened as a non-draft PR I assumed you wanted a review. Sorry if that's not true.
> 
> I left one small comment on the pinning changes. Please note you'll also have to manually update the conda recipes for `libcucim`:
> 
> https://github.com/rapidsai/cucim/blob/ec077cae16a5eb5fca48fb874360feb5f5671971/conda/recipes/libcucim/conda_build_config.yaml#L19-L20
> 
> Could you also add a brief description to the PR explaining the goal? That'd help reviewers... I'm unsure about the purpose of some of the `cucim.kit.cuslide2/CMakeLists.txt` changes.

thank you for the help @jameslamb - indeed, per cuCIM PR template description a PR marked as [WIP] is not ready for review yet.I will, however , add a PR description shortly.This PR is the second PR following [this](https://github.com/rapidsai/cucim/pull/997) PR

cdinea:
> I see that this has `[WIP]` but since it was opened as a non-draft PR I assumed you wanted a review. Sorry if that's not true.
> 
> I left one small comment on the pinning changes. Please note you'll also have to manually update the conda recipes for `libcucim`:
> 
> https://github.com/rapidsai/cucim/blob/ec077cae16a5eb5fca48fb874360feb5f5671971/conda/recipes/libcucim/conda_build_config.yaml#L19-L20
 thank you for the feedback @jameslamb - I addressed this feedback in [this](https://github.com/rapidsai/cucim/pull/998/commits/cb1f9c3c6d95eef4b8256e09a0c72f2a0c43e8ad) commit

dependencies.yaml
cdinea: thank you for the great feedback @jameslamb - yes, I made this change in the infra-only PR #997 that will be merged into this one . It is a great idea to have the nvImageCodec version as a range rather than the exact version

cpp/plugins/cucim.kit.cuslide2/src/cuslide/nvimgcodec/nvimgcodec_tiff_parser.h
```diff
@@ -357,13 +395,27 @@ class NvImageCodecTiffParserManager
...
 
 // Stub implementations when nvImageCodec is not available
-enum class ImageType {
-    RESOLUTION_LEVEL,
-    THUMBNAIL,
-    LABEL,
-    MACRO,
-    UNKNOWN
-};
+// Keep a compatible TiffTagValue type even when nvImageCodec is disabled.
```
mkepa-nv: The type definition is the same, so I would just move it above `#ifdef CUCIM_HAS_NVIMGCODEC`

cpp/plugins/cucim.kit.cuslide2/CMakeLists.txt
```diff
@@ -286,6 +286,8 @@ endif()
 
 target_include_directories(cuslide2_core
         PRIVATE
+            ${CMAKE_CURRENT_SOURCE_DIR}/../cucim.kit.cuslide/src
+            ${CMAKE_CURRENT_SOURCE_DIR}/src  # Include cuslide2 src for nvimgcodec headers
             ${CMAKE_CURRENT_SOURCE_DIR}/src
 )
```
jakirkham, line 290: Is this change intentional?

Looks like one entry was duplicated

Also not sure about the `cuslide` reference in `cuslide2`

cpp/plugins/cucim.kit.cuslide2/src/cuslide/nvimgcodec/nvimgcodec_decoder.cpp
```diff
@@ -105,9 +110,12 @@ class DecodeBuffer
...
         {
-            // Use pinned memory for faster GPU-to-host transfers when GPU backend is used
-            cudaError_t status = cudaMallocHost(&buffer_, size);
-            return status == cudaSuccess;
+            // Use standard malloc for CPU memory
+            // NOTE: Must use malloc() (not cudaMallocHost()) because cuCIM's cleanup
+            // code uses free(). Using cudaMallocHost() would require cudaFreeHost(),
+            // causing "free(): invalid pointer" crash when cuCIM calls free().
+            buffer_ = malloc(size);
+            return buffer_ != nullptr;
```
jakirkham, line 118: So pinned memory is no longer needed?

How does this perform without pinned memory?

cpp/plugins/cucim.kit.cuslide2/src/cuslide/nvimgcodec/nvimgcodec_decoder.cpp
```diff
@@ -147,9 +155,14 @@ bool decode_ifd_region_nvimgcodec(const IfdInfo& ifd_info,
                                   nvimgcodecCodeStream_t main_code_stream,
                                   uint32_t x, uint32_t y,
                                   uint32_t width, uint32_t height,
-                                  uint8_t*& output_buffer,
+                                  uint8_t** output_buffer,
```
jakirkham, line 158: Why are we using a double pointer here?

cpp/plugins/cucim.kit.cuslide2/src/cuslide/nvimgcodec/nvimgcodec_tiff_parser.cpp
```diff
@@ -229,19 +301,10 @@ TiffFileParser::TiffFileParser(const std::string& file_path)
...
-            // Don't call destroy in error path - might cause crash
-            // nvimgcodecCodeStreamDestroy(main_code_stream_);
-            fprintf(stderr, "[DEBUG] Skipping nvimgcodecCodeStreamDestroy to avoid crash\n");
-            main_code_stream_ = nullptr;
-        }
-
-        fprintf(stderr, "[DEBUG] Re-throwing exception...\n");
+        // Don't explicitly destroy main_code_stream_ here; let the destructor handle it
+        // (static destruction order / nvimgcodec teardown can be tricky).
+        (void)e;
```
mkepa-nv: If you don't care about exception and rethrow it anyway, then you can change to empty catch (which will catch any exception) : `catch()`

cpp/plugins/cucim.kit.cuslide2/src/cuslide/nvimgcodec/nvimgcodec_tiff_parser.cpp
```diff
@@ -229,19 +301,10 @@ TiffFileParser::TiffFileParser(const std::string& file_path)
...
+        main_code_stream_ = nullptr;
```
mkepa-nv: You will leak memory if `main_code_stream_` is not a nullptr, you need to call `nvimgcodecCodeStreamDestroy()` in that case. What kind of crash were you seeing before?

cpp/plugins/cucim.kit.cuslide2/src/cuslide/nvimgcodec/nvimgcodec_tiff_parser.cpp
```diff
@@ -229,19 +301,10 @@ TiffFileParser::TiffFileParser(const std::string& file_path)
     }
     catch (const std::exception& e)
```
mkepa-nv: nvImageCodec has C api - it will never throw. The only place that can throw is `throw std::runtime_error...` added by you. So I don't think the try catch block is necessary - you can just handle memory free inside that `if` block

cpp/plugins/cucim.kit.cuslide2/src/cuslide/nvimgcodec/nvimgcodec_tiff_parser.cpp
```diff
@@ -759,55 +804,175 @@ void TiffFileParser::extract_tiff_tags(IfdInfo& ifd_info)
...
-    if (extracted_count > 0 || !ifd_info.tiff_tags.empty())
+    if (extracted_count > 0)
     {
-        fmt::print("  ✅ Compression detection successful (file extension heuristics)\n");
+        return;
     }
-    else
+#else
+#ifdef DEBUG
+    fmt::print("  ℹ️  Skipping TIFF tag extraction: nvimgcodec TIFF_TAG metadata not supported by this nvimgcodec version\n");
```
mkepa-nv: NIT: I would add to this debug message which version of nvImageCodec is needed to support this feature

cpp/plugins/cucim.kit.cuslide2/src/cuslide/nvimgcodec/nvimgcodec_tiff_parser.cpp
```diff
@@ -759,55 +804,175 @@ void TiffFileParser::extract_tiff_tags(IfdInfo& ifd_info)
...
+                if (!str_val.empty()) tag_value = std::move(str_val);
+                break;
+            }
+            case NVIMGCODEC_METADATA_VALUE_TYPE_SHORT:
+                extract_single_value<uint16_t>(buffer, metadata.value_count, tag_value) ||
+                extract_value_array<uint16_t>(buffer, metadata.value_count, tag_value);
+                break;
+            case NVIMGCODEC_METADATA_VALUE_TYPE_LONG:
+                extract_single_value<uint32_t>(buffer, metadata.value_count, tag_value) ||
+                extract_value_array<uint32_t>(buffer, metadata.value_count, tag_value);
```
mkepa-nv: NIT: you always use both single and array extract. So maybe it would be better to merge them into a single `extract_tag_value` function?

cpp/plugins/cucim.kit.cuslide2/src/cuslide/nvimgcodec/nvimgcodec_tiff_parser.cpp
```diff
@@ -759,55 +804,175 @@ void TiffFileParser::extract_tiff_tags(IfdInfo& ifd_info)
...
+                break;
+            case NVIMGCODEC_METADATA_VALUE_TYPE_BYTE:
+                if (metadata.value_count == 1) tag_value = buffer[0];
+                else tag_value = std::vector<uint8_t>(buffer.begin(), buffer.begin() + metadata.buffer_size);
+                break;
+            case NVIMGCODEC_METADATA_VALUE_TYPE_SBYTE:
+                if (metadata.value_count == 1) tag_value = static_cast<int8_t>(buffer[0]);
+                else tag_value = std::vector<uint8_t>(buffer.begin(), buffer.begin() + metadata.buffer_size);
```
mkepa-nv: Shouldn't this be a `std::vector<int8_t>`?

cpp/plugins/cucim.kit.cuslide2/src/cuslide/nvimgcodec/nvimgcodec_tiff_parser.cpp
```diff
@@ -759,55 +804,175 @@ void TiffFileParser::extract_tiff_tags(IfdInfo& ifd_info)
...
+                break;
+            case NVIMGCODEC_METADATA_VALUE_TYPE_UNDEFINED:
+                tag_value = std::vector<uint8_t>(buffer.begin(), buffer.begin() + metadata.buffer_size);
+                break;
+            case NVIMGCODEC_METADATA_VALUE_TYPE_SSHORT:
+                extract_single_value<int16_t>(buffer, metadata.value_count, tag_value) ||
+                extract_value_array<uint16_t>(buffer, metadata.value_count, tag_value);
```
mkepa-nv: Here and below you are using vector with unsigned type for signed tags. Is this on purpose?

cpp/plugins/cucim.kit.cuslide2/src/cuslide/nvimgcodec/nvimgcodec_tiff_parser.cpp
mkepa-nv: Oh, I didn't see that you use different sign type for single and array extraction. But it doesn't looks right. If you will change the code to use the same signess then please unify this extract function

cpp/plugins/cucim.kit.cuslide2/src/cuslide/nvimgcodec/nvimgcodec_tiff_parser.cpp
```diff
@@ -759,55 +804,175 @@ void TiffFileParser::extract_tiff_tags(IfdInfo& ifd_info)
...
-        if (ifd_info.tiff_tags.find("IMAGEDESCRIPTION") == ifd_info.tiff_tags.end()) {
-            ifd_info.tiff_tags["IMAGEDESCRIPTION"] = ifd_info.image_description;
-        }
+        ifd_info.image_description = tiff_tag_value_to_string(desc_it->second);
     }
 
-    // Summary
-    #ifdef DEBUG
-    if (extracted_count > 0 || !ifd_info.tiff_tags.empty())
+    if (extracted_count > 0)
```
mkepa-nv: What if "compression" tag was not in the file? Maybe we should still use heuristic in that case? (And a better way to early exit would be to directly check if compression tag was read

cpp/plugins/cucim.kit.cuslide2/src/cuslide/nvimgcodec/nvimgcodec_decoder.h
```diff
@@ -52,6 +52,66 @@ bool decode_ifd_region_nvimgcodec(const IfdInfo& ifd_info,
...
+ * @param ifd_index IFD index (resolution level)
+ * @param tile_x Tile X coordinate in pixels
+ * @param tile_y Tile Y coordinate in pixels
+ * @param tile_width Tile width in pixels
+ * @param tile_height Tile height in pixels
+ * @param dest Output buffer pointer (will be allocated)
+ * @param out_device Output device ("cpu" or "cuda")
+ * @return true if successful, false to fallback to other decoders
+ */
+bool decode_tile_nvtiff_roi(const char* file_path,
```
mkepa-nv: This function is declared but never defined

cpp/plugins/cucim.kit.cuslide2/src/cuslide/nvimgcodec/nvimgcodec_decoder.h
```diff
@@ -52,6 +52,66 @@ bool decode_ifd_region_nvimgcodec(const IfdInfo& ifd_info,
...
+ *
+ * This function uses the parsed IfdInfo (from TiffFileParser) to decode
+ * a full resolution level. It separates parsing from decoding.
+ *
+ * @param ifd_info Parsed IFD information with sub_code_stream
+ * @param output_buffer Pointer to receive allocated buffer (caller must free)
+ * @param out_device Output device ("cpu" or "cuda")
+ * @return true if successful, false otherwise
+ */
+bool decode_ifd_nvimgcodec(const IfdInfo& ifd_info,
```
mkepa-nv: This function is declared but never defined

cpp/plugins/cucim.kit.cuslide2/src/cuslide/nvimgcodec/nvimgcodec_decoder.h
```diff
@@ -52,6 +52,66 @@ bool decode_ifd_region_nvimgcodec(const IfdInfo& ifd_info,
...
+ * @return true if successful, false to fallback to other decoders
+ */
+bool decode_tile_nvtiff_roi(const char* file_path,
+                            uint32_t ifd_index,
+                            uint32_t tile_x, uint32_t tile_y,
+                            uint32_t tile_width, uint32_t tile_height,
+                            uint8_t** dest,
+                            const cucim::io::Device& out_device);
+
+#ifdef CUCIM_HAS_NVIMGCODEC
```
mkepa-nv: Why are function declared inside `#ifdef`? I though that cuCIM approach is to function be always available, but they would throw if build was without nvimgcodec?

cpp/plugins/cucim.kit.cuslide2/src/cuslide/nvimgcodec/nvimgcodec_decoder.h
```diff
@@ -52,6 +52,66 @@ bool decode_ifd_region_nvimgcodec(const IfdInfo& ifd_info,
...
+ * @param main_code_stream Main TIFF code stream (for creating ROI views)
+ * @param x Starting x coordinate (column)
+ * @param y Starting y coordinate (row)
+ * @param width Width of region in pixels
+ * @param height Height of region in pixels
+ * @param output_buffer Pointer to receive allocated buffer (caller must free)
+ * @param out_device Output device ("cpu" or "cuda")
+ * @return true if successful, false otherwise
+ */
+bool decode_ifd_region_nvimgcodec(const IfdInfo& ifd_info,
```
mkepa-nv: As I wrote above, I don't think that function should be inside #ifdef
Right now you have two `decode_ifd_region_nvimgcodec` functions which different only that one takes ` uint8_t** output_buffer` and the other ` uint8_t*& output_buffer` which I suspect is a bug

mkepa-nv: Oh, nevermind, you just don't want  a breaking change. But still I would move this outside #ifdef. And think about moving old function definition (that just delegates to the new one) inside the header, so that future readers are not surprised as I was

cpp/plugins/cucim.kit.cuslide2/src/cuslide/nvimgcodec/nvimgcodec_decoder.cpp
```diff
@@ -283,158 +298,120 @@ bool decode_ifd_region_nvimgcodec(const IfdInfo& ifd_info,
...
     catch (const std::exception& e)
     {
+        #ifdef DEBUG
         fmt::print("❌ Exception in ROI decoding: {}\n", e.what());
+        #endif
         return false;
     }
 }
 
+bool decode_ifd_region_nvimgcodec(const IfdInfo& ifd_info,
```
mkepa-nv: I would move this definition outside #ifdef block - it can be the same whether we build with or without nvimgcodec. Or even consider moving it directly to header file

cpp/plugins/cucim.kit.cuslide2/src/cuslide/nvimgcodec/nvimgcodec_decoder.cpp
```diff
@@ -283,158 +298,120 @@ bool decode_ifd_region_nvimgcodec(const IfdInfo& ifd_info,
         output_image_info.plane_info[0].num_channels = num_channels;
         output_image_info.plane_info[0].row_stride = row_stride;
         output_image_info.plane_info[0].sample_type = NVIMGCODEC_SAMPLE_DATA_TYPE_UINT8;
-        output_image_info.buffer_size = buffer_size;
+        // Note: buffer_size removed in nvImageCodec v0.7.0 - size is inferred from plane_info
         output_image_info.cuda_stream = 0;
```
mkepa-nv, line 302: I would consider using custom stream, so that you don't need to sync the whole gpu

cpp/plugins/cucim.kit.cuslide2/src/cuslide/nvimgcodec/nvimgcodec_decoder.cpp
```diff
@@ -213,68 +242,54 @@ bool decode_ifd_region_nvimgcodec(const IfdInfo& ifd_info,
...
             return false;
         }
         UniqueCodeStream roi_stream(roi_stream_raw);
 
-        // Step 2: Determine buffer kind
-        // IMPORTANT: nvImageCodec v0.6.0 GPU JPEG decoder can ONLY output to device memory
-        // For CPU requests, we must decode to GPU then copy D2H (device-to-host)
-
-        // Verify GPU is available
+        // Step 2: Determine buffer kind based on target device and availability
```
mkepa-nv, line 258: If the output is provided by the user, I think it should be rather set based on the source of that buffer (which can be checked with `cudaPointerGetAttributes`)

cpp/plugins/cucim.kit.cuslide2/src/cuslide/nvimgcodec/nvimgcodec_decoder.cpp
```diff
@@ -283,158 +298,120 @@ bool decode_ifd_region_nvimgcodec(const IfdInfo& ifd_info,
...
             fmt::print("❌ Failed to get processing status (status: {})\n", static_cast<int>(status));
-            return false;  // RAII handles cleanup
+            #endif
+            return false;
         }
 
         if (use_device_memory)
         {
-            cudaStreamSynchronize(output_image_info.cuda_stream);
+            cudaDeviceSynchronize();
```
mkepa-nv, line 401: As I wrote above, I would consider using cuda streams to avoid syncing the whole gpu (in case there is some other work on it)

python/cucim/tests/performance/clara/test_read_region_memory_usage.py
```diff
@@ -250,7 +255,11 @@ def get_total_mem_usage():
...
-                        "{memory_increase:4d} MB"
+                        f"{memory_increase:4d} MB"
                     )
-        # Memory usage difference should be less than 20MB
+        # Memory usage difference should be less than 20MB.
+        # (Include history in the failure message to aid debugging.)
+                # Memory usage difference should be less than 20MB
+        if mem_usage_history[-1] - mem_usage_history[1] < 20:
+            assert False, f"mem_usage_history: {mem_usage_history}"
         assert mem_usage_history[-1] - mem_usage_history[1] < 20
```
mkepa-nv: This assert can be now remove, you check the condition in the if above

cpp/plugins/cucim.kit.cuslide2/CMakeLists.txt
cdinea: thank you for the feedback @jakirkham - I removed the duplicates in [this](https://github.com/rapidsai/cucim/pull/998/commits/9708967daf812b814c2fb35fdb9d3a960949f24e) commit

cpp/plugins/cucim.kit.cuslide2/src/cuslide/nvimgcodec/nvimgcodec_decoder.cpp
cdinea: thank you for the feedback @jakirkham - this change isn't about pinned memory being unnecessary for performance - it's about memory ownership compatibility :the issue is that cuCIM's cleanup code uses free() to deallocate this buffer (we release ownership via decode_buffer.release() on line 387). Using cudaMallocHost() would require cudaFreeHost() for proper deallocation, but since cuCIM calls free() instead, we'd get a "free(): invalid pointer" crash.Regarding performance:
- pinned memory primarily benefits GPU<->Host DMA transfers.In the CPU target path (where this else branch executes), decoding happens via nvImageCodec's CPU backend and the data stays on the host - no GPU transfers occur. For GPU targets, we still use cudaMalloc() (line 108), whicch provides proper GPU memory- so the performance impact should be minimal since this path is specifically for CPU-only decoding where pinned memory's DMA advantages don't apply.

cpp/plugins/cucim.kit.cuslide2/src/cuslide/nvimgcodec/nvimgcodec_decoder.cpp
cdinea: thank you for the feedback @jakirkham - the double pointer enables an optional input/output pattern:
-> if the caller provides a pre-allocated buffer, the function decodes into it
-> if the caller passes nullptr, the function allocates a buffer and returns it via the pointer

this allows flexibility - callers can either manage their own buffer or let the function handle allocation

cpp/plugins/cucim.kit.cuslide2/src/cuslide/nvimgcodec/nvimgcodec_tiff_parser.h
cdinea: thank you for the feedback @mkepa-nv - i addressed this feedback in [this](https://github.com/rapidsai/cucim/pull/998/commits/96872baa201529393244064d6cdffc68160a9543) commit

cpp/plugins/cucim.kit.cuslide2/src/cuslide/nvimgcodec/nvimgcodec_tiff_parser.cpp
cdinea: thank you for the feedback @mkepa-nv - I addressed this feedback in [this](https://github.com/rapidsai/cucim/pull/998/commits/c8636773b5db84e9df5c43faeeb03600045fd983) commit

cpp/plugins/cucim.kit.cuslide2/src/cuslide/nvimgcodec/nvimgcodec_tiff_parser.cpp
cdinea: thank you for the feedback @mkepa-nv - i do not remember the exact crash right now, I will repro locally and come back to this feedback

cpp/plugins/cucim.kit.cuslide2/src/cuslide/nvimgcodec/nvimgcodec_tiff_parser.cpp
cdinea: thank you for the feedback @mkepa-nv - I addressed this feedback in [this](https://github.com/rapidsai/cucim/pull/998/commits/c504a9d1fc4fda0dd4a85efc54f30966a41218b0) comit

cpp/plugins/cucim.kit.cuslide2/src/cuslide/nvimgcodec/nvimgcodec_tiff_parser.cpp
```diff
@@ -759,55 +805,175 @@ void TiffFileParser::extract_tiff_tags(IfdInfo& ifd_info)
...
+    std::string ext;
+    size_t dot_pos = file_path_.rfind('.');
+    if (dot_pos != std::string::npos)
     {
-        fmt::print("  ⚠️  Unable to determine compression type from file extension\n");
+        ext = file_path_.substr(dot_pos);
+        std::transform(ext.begin(), ext.end(), ext.begin(), ::tolower);
+    }
+
+    if (ext == ".svs" || ext == ".ndpi" || ext == ".vms" || ext == ".vmu")
```
gigony:
```suggestion
    if (ext == ".svs" || ext == ".scn" || ext == ".ndpi" || ext == ".bif")
```

AFAIU, nvImageCodec supports the following formats:
- Aperio (.svs, .tif)
- Ventana (.bif/.tif)
- Hamamatsu NDPI (.ndpi) # note: no '.vms'/'.vmu' format supported by nvImageCodec for now. so please remove '.vms' and '.vmu'.
- Leica (.scn)
- Trestle (.tif)

python/cucim/tests/performance/clara/test_read_region_memory_usage.py
```diff
@@ -247,7 +255,11 @@ def get_total_mem_usage():
...
                     print(
                         f"mem increase (iteration: {i:3d}): "
-                        "{memory_increase:4d} MB"
+                        f"{memory_increase:4d} MB"
                     )
-        # Memory usage difference should be less than 20MB
+        # Memory usage difference should be less than 20MB.
+        # (Include history in the failure message to aid debugging.)
+                # Memory usage difference should be less than 20MB
+        if mem_usage_history[-1] - mem_usage_history[1] < 20:
```
jakirkham: The [style checker caught this spacing difference]( https://github.com/rapidsai/cucim/actions/runs/21183827413/job/60932845982?pr=998#step:7:239 ). Think this is all we need to fix it

```suggestion
        # Memory usage difference should be less than 20MB.
        # (Include history in the failure message to aid debugging.)
        # Memory usage difference should be less than 20MB
        if mem_usage_history[-1] - mem_usage_history[1] < 20:
```

python/cucim/tests/performance/clara/test_read_region_memory_usage.py
```diff
@@ -5,6 +5,7 @@
 
 import cupy as cp
 import pytest
+import os
```
jakirkham: Also [the style check recommended reordering these `import`s]( https://github.com/rapidsai/cucim/actions/runs/21183827413/job/60932845982?pr=998#step:7:229 )

```suggestion
import os

import cupy as cp
import pytest
```

python/cucim/tests/performance/clara/test_read_region_memory_usage.py
```diff
@@ -5,6 +5,7 @@
```
jakirkham, line 5: The copyright header in this file needs to be updated to include 2026 as [noted by this style check error]( https://github.com/rapidsai/cucim/actions/runs/21183827413/job/60932845982?pr=998#step:7:224 )

dependencies.yaml
jakirkham: Thanks James and Cristiana! 🙏

As this is now addressed, marking this thread as resolved

cpp/plugins/cucim.kit.cuslide2/src/cuslide/nvimgcodec/nvimgcodec_tiff_parser.cpp
cdinea: thank you for the feedback @mkepa-nv - I addrressed this feedback in [this](https://github.com/rapidsai/cucim/pull/998/commits/c5bd864fd59753c6b8774262c4d3df2c23704ca7) commit

cpp/plugins/cucim.kit.cuslide2/src/cuslide/nvimgcodec/nvimgcodec_tiff_parser.cpp
cdinea: thank you for the feedback @mkepa-nv - I made these changes to address this feedback 
->unified extract_tag_value function - merged extract_single_value and extract_value_array into a single function that handles both cases based on value_count.
->fixed signedness consistency - now signed types use signed types consistently:
SSHORT → int16_t (was mixing int16_t and uint16_t)
SLONG → int32_t (was mixing int32_t and uint32_t)
SLONG8 → int64_t (was mixing int64_t and uint64_t

the changes are in [this](https://github.com/rapidsai/cucim/pull/998/commits/f9def03341d9d27eae20a10c3f8d953037467866) commit

cpp/plugins/cucim.kit.cuslide2/src/cuslide/nvimgcodec/nvimgcodec_tiff_parser.cpp
cdinea: thank you for the feedback @mkepa-nv - i fixed the SBYTE case to use std::vector<int8_t> for consistency in [this](https://github.com/rapidsai/cucim/pull/998/commits/b020c6afd2d5667a0b1830dbdf7868c99ea5ba52) commit

cpp/plugins/cucim.kit.cuslide2/src/cuslide/nvimgcodec/nvimgcodec_tiff_parser.cpp
cdinea: thank you for the feedback @mkepa-nv - this was fixed in [this](https://github.com/rapidsai/cucim/pull/998/commits/f9def03341d9d27eae20a10c3f8d953037467866) commit

cpp/plugins/cucim.kit.cuslide2/src/cuslide/nvimgcodec/nvimgcodec_tiff_parser.cpp
cdinea: thank you for the feedback @mkepa-nv - i addressed this feedback in [this](https://github.com/rapidsai/cucim/pull/998/commits/4f74aedb0ed14cf364cba11c0ad0badc2720eb31) commit

cpp/plugins/cucim.kit.cuslide2/src/cuslide/nvimgcodec/nvimgcodec_decoder.h
cdinea: thank you for the feedback @mkepa-nv - I addressed this feedback in [this](https://github.com/rapidsai/cucim/pull/998/commits/8aa4bda04250298986c43230dedb200e84a8e85b) commit

cpp/plugins/cucim.kit.cuslide2/src/cuslide/nvimgcodec/nvimgcodec_decoder.h
cdinea: thank you for the feedback @mkepa-nv - i addressed this feedback in [this](https://github.com/rapidsai/cucim/pull/998/commits/84bc0c8db70ae1d311d15f423cf6ace32444b9ed) commit

cpp/plugins/cucim.kit.cuslide2/src/cuslide/nvimgcodec/nvimgcodec_decoder.h
cdinea: thank you for the feedback @mkepa-nv - i addressed this feedback in [this](https://github.com/rapidsai/cucim/pull/998/commits/7207c08052eb7b9c23fda61f2bf9b9701df6382b) commit

cpp/plugins/cucim.kit.cuslide2/src/cuslide/nvimgcodec/nvimgcodec_decoder.h
cdinea: thank you for the feedback - i addressed this feedback in [this](https://github.com/rapidsai/cucim/pull/998/commits/0db0a322b6c153cd8e3be8e73d7b4ed5ff7a0265) -   uint8_t*& version is just a wrapper , i added its implementation in the header as an inline function

cpp/plugins/cucim.kit.cuslide2/src/cuslide/nvimgcodec/nvimgcodec_decoder.cpp
cdinea: thank you for the feedback @mkepa-nv - i addressed this in a previous commit [here](0db0a32)

cpp/plugins/cucim.kit.cuslide2/src/cuslide/nvimgcodec/nvimgcodec_decoder.cpp
cdinea: thank you for the feedback @mkepa-nv - using a custom stream would defintiely improve performance-however, this requires careful design decisions (stream ownership, lifetime management, whether to use per-decoder or caller-provided streams). I'd prefer to address this in a follow-up PR focused on stream optimization, rather than expanding the scope of this PR.
 Filed as a TODO for future improvement.

cpp/plugins/cucim.kit.cuslide2/src/cuslide/nvimgcodec/nvimgcodec_decoder.cpp
cdinea: thank you  for the feedback @mkepa-nv - i added cudaPointerGetAttributes() when caller provides a buffer in [this](https://github.com/rapidsai/cucim/pull/998/commits/e6daf1e3bb6dde0b8f31e5c4cef7a074d969a583) commit

cpp/plugins/cucim.kit.cuslide2/src/cuslide/nvimgcodec/nvimgcodec_decoder.cpp
cdinea: thank you for the feedback @mkepa-nv - I'll address this in a follow-up PR focused on performance improvements

python/cucim/tests/performance/clara/test_read_region_memory_usage.py
cdinea: thank you for the feedback @mkepa-nv - indeed, we added this for an unexpected CI/CD failure - i addressed this feedback in [this](https://github.com/rapidsai/cucim/pull/998/commits/2f2f614ff3c83d0267c3002f4921995cae966d5b) commit

python/cucim/tests/performance/clara/test_read_region_memory_usage.py
cdinea: thank you for the guidance @jakirkham - i addressed this feedback in [this](https://github.com/rapidsai/cucim/pull/998/commits/40dc23af7a3dd7663eb4e4509d503a9c3df76237) commit

cpp/plugins/cucim.kit.cuslide2/src/cuslide/nvimgcodec/nvimgcodec_tiff_parser.cpp
```diff
@@ -272,40 +272,39 @@ TiffFileParser::TiffFileParser(const std::string& file_path)
...
+        &main_code_stream_,
+        file_path.c_str()
+    );
 
-        #ifdef DEBUG
-        fmt::print("✅ Opened TIFF file: {}\n", file_path);
-        #endif // DEBUG
+    if (status != NVIMGCODEC_STATUS_SUCCESS)
+    {
+        throw std::runtime_error(fmt::format("Failed to create code stream from file: {} (status: {})",
```
mkepa-nv, line 287: `nvimgcodecCodeStreamCreateFromFile` may return error but still create codeStream. This is the case, when there is some problem with reading filed provided by the user. So you should destroy `main_code_stream_`  here if it is not a nullptr. I can see this is not a great design of nvImageCodec and we will try to improve it in the future

cpp/plugins/cucim.kit.cuslide2/src/cuslide/nvimgcodec/nvimgcodec_decoder.cpp
```diff
@@ -306,6 +306,37 @@ bool decode_ifd_region_nvimgcodec(const IfdInfo& ifd_info,
         DecodeBuffer decode_buffer;
         if (caller_provided_buffer)
         {
+            // Query actual memory type of caller-provided buffer
+            cudaPointerAttributes attrs{};
+            cudaError_t ptr_err = cudaPointerGetAttributes(&attrs, *output_buffer);
+            if (ptr_err == cudaSuccess)
+            {
+                // Determine buffer kind based on actual memory type
+                #if CUDART_VERSION >= 10000
```
mkepa-nv: Is this a cuda 10.0 check? Do you support such old versions?

jantonguirao: nvimagecodec doesn't support CUDA 10.

cpp/plugins/cucim.kit.cuslide2/src/cuslide/nvimgcodec/nvimgcodec_decoder.cpp
jantonguirao: Even if using stream 0, I would definitely keep cudaStreamSynchronize. cudaDeviceSynchronize() is way to aggressive, as it will wail for all work scheduled to that device, which is unnecessary.

cpp/plugins/cucim.kit.cuslide2/src/cuslide/nvimgcodec/nvimgcodec_decoder.cpp
```diff
@@ -213,68 +242,54 @@ bool decode_ifd_region_nvimgcodec(const IfdInfo& ifd_info,
...
-        // Use interleaved RGB format
         output_image_info.sample_format = NVIMGCODEC_SAMPLEFORMAT_I_RGB;
         output_image_info.color_spec = NVIMGCODEC_COLORSPEC_SRGB;
         output_image_info.chroma_subsampling = NVIMGCODEC_SAMPLING_NONE;
         output_image_info.num_planes = 1;
         output_image_info.buffer_kind = buffer_kind;
 
-        // Calculate buffer requirements for the region
-        uint32_t num_channels = 3;  // RGB
+        uint32_t num_channels = 3;
```
jantonguirao, line 292: are we only interested in RGB outputs? Perhaps we should do the original format: 
output_image_info.image_info.sample_format = NVIMGCODEC_SAMPLEFORMAT_I_UNCHANGED;
and num_channels the original one.

cpp/plugins/cucim.kit.cuslide2/src/cuslide/nvimgcodec/nvimgcodec_tiff_parser.cpp
cdinea: thank you for catching this @mkepa-nv - it is fixed in [this](https://github.com/rapidsai/cucim/pull/998/commits/363ffe1f533943373e9c5b618f459a37a367d853) commit - now we clean up main_code_stream_ if it's non-null before throwing

cpp/plugins/cucim.kit.cuslide2/src/cuslide/nvimgcodec/nvimgcodec_decoder.cpp
cdinea: thank you for the feedback @mkepa-nv - good point, cuCIM requires CUDA 12+, so the CUDA 10 check was unnecessary.Fixed in [this](https://github.com/rapidsai/cucim/pull/998/commits/1c3a94ed6a7144ae8dfcf571dcc554b1057c7267) commit

cpp/plugins/cucim.kit.cuslide2/src/cuslide/nvimgcodec/nvimgcodec_decoder.cpp
cdinea: thank you for the feedback @jantonguirao - currently the cuslide2 decoder is focused on RGB output to match the behavior of the existing cuslide plugin and typical WSI use cases - @gigony @grlee77 to confirm.Supporting NVIMGCODEC_SAMPLEFORMAT_I_UNCHANGED would be a nice enhancement for grayscale/CMYK images, but it would require:
->querying original format from IFD metadata
->adjusting buffer size calculations dynamically
->updating downstream consumers that expect RGB

grlee77: @gigony is the original author of the cuslide plugin. I think it is okay to match current behavior for now, but please document this for future update/improvement (possibly via GitHub issue or wherever you are tracking future enhancements for the nvimgcodec integration).

cpp/plugins/cucim.kit.cuslide2/src/cuslide/nvimgcodec/nvimgcodec_tiff_parser.cpp
```diff
@@ -42,6 +39,88 @@ namespace cuslide2::nvimgcodec
...
+            return "";
+        }
+        else if constexpr (std::is_same_v<T, std::string>)
+        {
+            return v;
+        }
+        else if constexpr (std::is_same_v<T, std::vector<int8_t>> ||
+                           std::is_same_v<T, std::vector<uint8_t>>)
+        {
+            return fmt::format("[{} bytes]", v.size());
```
grlee77, line 74: why the difference between printing [{size} bytes] here vs. all other vector data types which are printing a subset of values in the else if case just below this one?

cpp/plugins/cucim.kit.cuslide2/src/cuslide/nvimgcodec/nvimgcodec_tiff_parser.cpp
```diff
@@ -42,6 +39,88 @@ namespace cuslide2::nvimgcodec
...
+            // Print a small prefix of arrays.
+            std::string result;
+            const size_t limit = 10;
+            for (size_t i = 0; i < v.size() && i < limit; ++i)
+            {
+                if (i > 0) result += ",";
+                result += std::to_string(v[i]);
+            }
+            if (v.size() > limit) result += ",...";
+            return result;
```
grlee77: Repeated string concatenation is fairly inefficient. The `fmt` library has ways to do this more efficiently such as the following

```cpp
    const std::size_t limit = 10;
    const auto n = std::min(v.size(), limit);

    auto result = fmt::format("{}", fmt::join(v.begin(), v.begin() + n, ","));  // e.g. "1,2,3"
    if (v.size() > limit)
        result += ",...";

    return result;
```

cpp/plugins/cucim.kit.cuslide2/src/cuslide/nvimgcodec/nvimgcodec_tiff_parser.cpp
```diff
@@ -42,6 +39,88 @@ namespace cuslide2::nvimgcodec
...
+        }
+        else if constexpr (std::is_same_v<T, float> || std::is_same_v<T, double>)
```
grlee77: only float and double are needed here or are other integer/unsigned scalar types also possible?

cpp/plugins/cucim.kit.cuslide2/src/cuslide/nvimgcodec/nvimgcodec_tiff_parser.cpp
cdinea: thank you for the feedback @grlee77 - I believe  byte arrays (uint8_t/int8_t vectors) often contain binary data like JPEG tables, raw TIFF tag data, or other opaque blobs where printing individual byte values as numbers (e.g., "72,101,108,108,111") isn't meaningful or useful for debugging - @mkepa-nv to confirm 
for other numeric types (int16, int32, float, etc.), the values themselves are typically meaningful metadata values worth displaying-  added a comment in the code to clarify this in [this](https://github.com/rapidsai/cucim/pull/998/commits/cc0ae0f8590df8d24be02cce25dfd27db654dd3c) commit

cpp/plugins/cucim.kit.cuslide2/src/cuslide/nvimgcodec/nvimgcodec_tiff_parser.cpp
cdinea: thank you for the feedback @grlee77 - this is a great suggestion . I addressed this feedback in [this](https://github.com/rapidsai/cucim/pull/998/commits/856f91f3f12259f486b289c412483f14c83cac5e) commit

cpp/plugins/cucim.kit.cuslide2/src/cuslide/nvimgcodec/nvimgcodec_tiff_parser.cpp
cdinea: thank you for catching this @grlee77 - the distinction for float and double is not really needed here. pushed the fix in [this](https://github.com/rapidsai/cucim/pull/998/commits/fb505de6aead765b447624ff298e07e859619919) commit

cpp/plugins/cucim.kit.cuslide2/src/cuslide/nvimgcodec/nvimgcodec_tiff_parser.cpp
cdinea: thank you for the feedback @gigony - I changed the supported file formats in [this](https://github.com/rapidsai/cucim/pull/998/commits/cf83f72bd4267d802d1f168865daa7472d15282d) commit

cpp/plugins/cucim.kit.cuslide2/src/cuslide/nvimgcodec/nvimgcodec_decoder.cpp
jakirkham: Ah ok. Agree if it is CPU only, there is no need for pinned memory

cpp/plugins/cucim.kit.cuslide2/src/cuslide/nvimgcodec/nvimgcodec_decoder.cpp
jakirkham: Ok though it is usually safer to pass by reference

Am wondering if overloading the function to make this parameter optional would handle the same use case (optional allocation) while maintaining the added safety that passing by reference provides

What do you think?

cdinea: thanks for the follow-up @jakirkham - agreed. We can refactor to model the optional output-buffer contract explicitly, while avoiding a T** out-parameter

cpp/plugins/cucim.kit.cuslide2/src/cuslide/nvimgcodec/nvimgcodec_decoder.h
```diff
@@ -49,9 +49,25 @@ bool decode_ifd_region_nvimgcodec(const IfdInfo& ifd_info,
...
 
+/**
+ * Convenience overload that takes a reference to pointer.
+ * Simply delegates to the pointer-to-pointer version above.
+ */
+inline bool decode_ifd_region_nvimgcodec(const IfdInfo& ifd_info,
+                                         nvimgcodecCodeStream_t main_code_stream,
+                                         uint32_t x, uint32_t y,
+                                         uint32_t width, uint32_t height,
+                                         uint8_t*& output_buffer,
```
gigony, line 63: looks like we also had to update this? (`uint8_t*&` to `uint8_t**`)

cpp/plugins/cucim.kit.cuslide2/src/cuslide/nvimgcodec/nvimgcodec_decoder.cpp
cdinea: filed as an issue here https://github.com/rapidsai/cucim/issues/1013

cpp/plugins/cucim.kit.cuslide2/src/cuslide/nvimgcodec/nvimgcodec_decoder.cpp
cdinea: tracked in jira issue here https://github.com/rapidsai/cucim/issues/1013

cpp/plugins/cucim.kit.cuslide2/src/cuslide/nvimgcodec/nvimgcodec_decoder.h
jakirkham: Thanks Gigon! 🙏

We discussed this above ( https://github.com/rapidsai/cucim/pull/998#discussion_r2696071757 ) and decided to roll this change back, which is what Cristiana is doing in PR: https://github.com/rapidsai/cucim/pull/1007

PR #1001, opened 2026-01-06
Fix broken binary_fill_holes in vendored code

grlee77:
Fixes a missing import that causes failure to run the vendored version of `binary_fill_holes`. This specific function is not currently relied on by any cuCIM code, but it is still good to have all functions work from the vendored module.


jakirkham (approved):

jakirkham:
Thanks Greg and Gigon! 🙏

PR #1002, opened 2026-01-06
Misc fixes and improvements to existing benchmark scripts

grlee77:
While working on a couple of new things I came across a few issues in the existing benchmark codes. This PR

- Fixes a bug that prevented benchmarks being run on GPU only via the `--no_cpu` command-line argument.
- Fixes a bug with replicated device names in the generated benchmark tables
- Adds a new CUCIM_BENCHMARK_MAX_DURATION environment variable for setting benchmark case duration without modifying the bash scripts
- stores any kwargs that were pass to the function in the benchmark table

### DIFF
--- a/benchmarks/skimage/run-nv-bench-transform.sh
+++ b/benchmarks/skimage/run-nv-bench-transform.sh
@@ -3,7 +3,7 @@
 # SPDX-License-Identifier: Apache-2.0
 
 # Use env var if set/non-empty, otherwise default to 10
-MAX_DURATION="${CUCIM_BENCHMARK_MAX_DURATION:-10}"  # [web:1][web:10]
+MAX_DURATION="${CUCIM_BENCHMARK_MAX_DURATION:-10}"
 
 param_shape=("512,512" "3840,2160" "3840,2160,3" "192,192,192")
 param_filt=(resize resize_local_mean rescale rotate downscale_local_mean warp_polar integral_image pyramid_gaussian pyramid_laplacian)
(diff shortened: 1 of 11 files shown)


benchmarks/skimage/run-nv-bench-transform.sh
```diff
@@ -1,14 +1,17 @@
 #!/bin/bash
-# SPDX-FileCopyrightText: Copyright (c) 2022-2025, NVIDIA CORPORATION. All rights reserved.
+# SPDX-FileCopyrightText: Copyright (c) 2022-2026, NVIDIA CORPORATION. All rights reserved.
 # SPDX-License-Identifier: Apache-2.0
 
+# Use env var if set/non-empty, otherwise default to 10
+MAX_DURATION="${CUCIM_BENCHMARK_MAX_DURATION:-10}"  # [web:1][web:10]
```
gigony: Just curious, where do those references (`[web:1]` and `[web:10]`) exist?

gigony (approved):
Thanks @grlee77 for improving the benchmark script! Looks good to me!

benchmarks/skimage/run-nv-bench-transform.sh
jakirkham: Based on our discussion with Greg, it sounds like this is copy-pasta and unneeded. Given that, let's go ahead and drop it

```suggestion
MAX_DURATION="${CUCIM_BENCHMARK_MAX_DURATION:-10}"
```

Edit: This shows up in several files. So we will want to update them all accordingly

PR #1015, opened 2026-01-28
fix(core): Replace strlen with strnlen for safer string handling

gigony:
- Replace strlen() with strnlen() in cuimage.cpp to prevent potential
  buffer overread if strings are unexpectedly not null-terminated
- Add maximum length constraints for spacing_units (256 bytes) and
  coord_sys (16 bytes) based on expected string sizes
- Addresses SonarQube security analysis for safe C string handling


jakirkham:
Thank you both! 🙏

cc @caryr35 (for awareness)

PR #1027, opened 2026-02-13
refactor: build wheels and conda packages using Python limited API

gforsyth:
Moving `cucim` over to using the limited API as part of https://github.com/rapidsai/build-planning/issues/42

Ops-Bot-Merge-Barrier: true

### DIFF
--- a/conda/recipes/cucim/meta.yaml
+++ b/conda/recipes/cucim/meta.yaml
@@ -62,7 +62,7 @@ requirements:
     - python-abi3 {{ py_abi_min }}.*
     - pip
     - rapids-build-backend >=0.4.0,<0.5.0
-    - scikit-image >=0.19.0,<0.25.0
+    - scikit-image >=0.23.2,<0.27.0
     - scipy >=1.6
     - setuptools >=80.9.0
   run:
@@ -74,7 +74,7 @@ requirements:
     - lazy_loader >=0.1
     - libcucim ={{ version }}
     - python
-    - scikit-image >=0.19.0,<0.25.0
+    - scikit-image >=0.23.2,<0.27.0
     - scipy >=1.6
   run_constrained:
     - openslide-python >=1.3.0


conda/recipes/cucim/meta.yaml
```diff
@@ -1,11 +1,12 @@
...
+# SPDX-FileCopyrightText: Copyright (c) 2021-2026, NVIDIA CORPORATION.
 # SPDX-License-Identifier: Apache-2.0
 
 {% set version = environ['RAPIDS_PACKAGE_VERSION'].lstrip('v') %}
-{% set py_version = environ['CONDA_PY'] %}
 {% set cuda_version = '.'.join(environ['RAPIDS_CUDA_VERSION'].split('.')[:2]) %}
 {% set cuda_major = cuda_version.split('.')[0] %}
 {% set date_string = environ['RAPIDS_DATE_STRING'] %}
+{% set py_abi_min = environ["RAPIDS_PY_VERSION"] %}
+{% set py_buildstring = py_abi_min.replace(".", "") %}
```
jakirkham, line 9:
```suggestion
{% set py_abi_min = environ["RAPIDS_PY_VERSION"] %}
{% set py_buildstring = py_abi_min.replace(".", "") %}
{% set py_runtime_latest = "3.13" %}
```

conda/recipes/cucim/meta.yaml
jakirkham, line 86:
```suggestion
tests:
  requirements:
    - cuda-version ={{ cuda_version }}
    - python ={{ py_runtime_latest }}
```

ci/test_python.sh
```diff
@@ -13,7 +13,7 @@ conda config --set channel_priority strict
 
 rapids-logger "Downloading artifacts from previous jobs"
 CPP_CHANNEL=$(rapids-download-conda-from-github cpp)
-PYTHON_CHANNEL=$(rapids-download-conda-from-github python)
+PYTHON_CHANNEL=$(rapids-download-from-github "$(rapids-package-name conda_python cucim --stable --cuda "$RAPIDS_CUDA_VERSION")")
```
jakirkham, line 16: Think we may not be installing everything we need here. Seeing [the following missing module errors in CI]( https://github.com/rapidsai/cucim/actions/runs/22004868795/job/63587685721?pr=1027#step:12:26626 ):

```
FAILED tests/unit/clara/test_batch_decoding.py::TestBatchDecoding::test_batch_read_multiple_locations[testimg_tiff_stripe_4096x4096_256_jpeg] - ModuleNotFoundError: No module named 'cucim.clara._cucim'
```

Are we...?

* Not uploading a package we need to the cache?
* Missing downloading a package from the cache?
* Not installing a third-party dependency we need?

ci/test_wheel.sh
```diff
@@ -6,11 +6,10 @@ set -eou pipefail
 
 source rapids-init-pip
 
-RAPIDS_PY_CUDA_SUFFIX="$(rapids-wheel-ctk-name-gen "${RAPIDS_CUDA_VERSION}")"
-PYTHON_WHEELHOUSE=$(RAPIDS_PY_WHEEL_NAME="cucim_${RAPIDS_PY_CUDA_SUFFIX}" rapids-download-wheels-from-github python)
+PYTHON_WHEELHOUSE=$(rapids-download-from-github "$(rapids-package-name wheel_python cucim --stable --cuda "$RAPIDS_CUDA_VERSION")")
 
 # echo to expand wildcard before adding `[extra]` requires for pip
-rapids-pip-retry install "$(echo ${PYTHON_WHEELHOUSE}/cucim*.whl)[test]"
+rapids-pip-retry install "$(echo "${PYTHON_WHEELHOUSE}"/cucim*.whl)[test]"
```
jakirkham, line 12: With the wheel case we see [a different error on CI]( https://github.com/rapidsai/cucim/actions/runs/22004868795/job/63587046036?pr=1027#step:13:47 ):

```
Looking in indexes: https://pip-cache.local.gha-runners.nvidia.com/simple, https://pypi.anaconda.org/rapidsai-wheels-nightly/simple
ERROR: cucim_cu13-26.4.0a19-cp311-cp311-manylinux_2_26_aarch64.manylinux_2_28_aarch64.whl is not a supported wheel on this platform.
```

Are we...?

* Installing the right wheel with this step
* Generating the name correctly

ci/build_wheel.sh
jakirkham, line 69: It is possible this step needs to be changed based on how the wheels are now being built

ci/test_python.sh
gforsyth: Thanks, John!

I think all the errors come down to the package not getting built as an `abi3` wheel for whatever reason -- the missing module errors are from trying to load (say) a `cp312` version of an object when there's only a `cp311` one available.

Going to poke at the builds locally and see where I'm missing a switch.

ci/test_wheel.sh
gforsyth: We're pulling down the right artifact, but the contents are incorrect (because we're generating them incorrectly).

gforsyth:
Ahh, right, `cucim` doesn't use scikit-build-core, so settings flags isn't going to do anything.  This one might be more involved

gforsyth:
`cucim` uses `pybind11` which doesn't support the limited API: https://github.com/pybind/pybind11/issues/5905

Issue #1030, opened 2026-02-18
[BUG] Compiler failure with _filters_core._call_kernel with cupy 14.0.0


NickGeneva:
**Describe the bug**

A recent update to cupy (version 14.0.0 as of [feb 16th](https://pypi.org/project/cupy-cuda12x/#history)) , moved the default c++ version to 17 which seems to cause some issues for me when using cucim.

**Steps/Code to reproduce bug**
Follow this guide http://matthewrocklin.com/blog/work/2018/02/28/minimal-bug-reports to craft a minimal bug report. This helps us reproduce the issue you're having and resolve the issue more quickly.

Minimal script:

```python
from cucim.skimage.feature import peak_local_max as cucim_peak_local_max
import cupy as cp

x = cp.random.uniform(10, 10, 10)
cucim_peak_local_max(x)
```

Error:

<details>

```
Traceback (most recent call last):
  File "/localhome/local-ngeneva/test/.venv/lib/python3.12/site-packages/cupy/cuda/compiler.py", line 786, in compile
    nvrtc.compileProgram(self.ptr, options)
  File "cupy_backends/cuda/libs/nvrtc.pyx", line 125, in cupy_backends.cuda.libs.nvrtc.compileProgram
  File "cupy_backends/cuda/libs/nvrtc.pyx", line 138, in cupy_backends.cuda.libs.nvrtc.compileProgram
  File "cupy_backends/cuda/libs/nvrtc.pyx", line 53, in cupy_backends.cuda.libs.nvrtc.check_status
cupy_backends.cuda.libs.nvrtc.NVRTCError: NVRTC_ERROR_COMPILATION (6)

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/localhome/local-ngeneva/test/main.py", line 5, in <module>
    cucim_peak_local_max(x)
  File "/localhome/local-ngeneva/test/.venv/lib/python3.12/site-packages/cucim/skimage/feature/peak.py", line 269, in peak_local_max
    mask = _get_peak_mask(image, footprint, threshold)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/localhome/local-ngeneva/test/.venv/lib/python3.12/site-packages/cucim/skimage/feature/peak.py", line 56, in _get_peak_mask
    image_max = ndi.maximum_filter(image, footprint=footprint, mode="nearest")
[...]

PR #1031, opened 2026-02-19
Drop CuPy pre-13.6.0 logic & fix CuPy 14 compatibility

jakirkham:
Fixes https://github.com/rapidsai/cucim/issues/1030

As part of the upgrade to CUDA 13.0, RAPIDS also updated to CuPy 13.6.0 as a minimum. Therefore the workarounds for CuPy 13.3.0 and earlier do not apply. So clean them up.

Should add this same cleanup is needed for CuPy 14 compatibility. So this fixes cuCIM to work with CuPy 14.

Lastly CuPy 14.0.0 included a bug that affected some `__warningregistry__` resetting logic in cuCIM: https://github.com/rapidsai/cucim/pull/1031#discussion_r2830243458

This issue will be fixed in CuPy 14.0.1. So simply excluded CuPy 14.0.0 from the CuPy versions installable with cuCIM.

Refs:

* https://github.com/rapidsai/build-planning/issues/208
* https://github.com/rapidsai/cucim/pull/926
* https://github.com/cupy/cupy/issues/9704

### DIFF
--- a/python/cucim/src/cucim/skimage/measure/_regionprops_gpu_utils.py
+++ b/python/cucim/src/cucim/skimage/measure/_regionprops_gpu_utils.py
@@ -1,4 +1,4 @@
-# SPDX-FileCopyrightText: Copyright (c) 2026, NVIDIA CORPORATION. All rights reserved.
+# SPDX-FileCopyrightText: Copyright (c) 2025-2026, NVIDIA CORPORATION. All rights reserved.
 # SPDX-License-Identifier: Apache-2.0
 
 import math


python/cucim/src/cucim/core/operations/morphology/_pba_2d.py
```diff
@@ -92,7 +92,6 @@ def _get_pack_kernel(int_type, marker=-32768):
         in_params="raw B arr",
         out_params="raw I out",
         operation=code,
-        options=("--std=c++11",),
```
jakirkham, line 95: Starting in CuPy 13.0.0, [the default is to provide `-std=c++11` if the `-std` option is unspecified]( https://github.com/cupy/cupy/blob/1bdaf31e6c66e2bfcf714504d257e91f65491d1c/cupy/_core/raw.pyx#L43-L45 )

Given we are using CuPy 13.6.0 or newer, this always the case for us. So we can simply drop these lines

python/cucim/src/cucim/skimage/measure/_regionprops_gpu_utils.py
```diff
@@ -9,19 +9,8 @@
...
 # Need some default includes so uint32_t, uint64_t, etc. are defined
 
-if CUPY_GTE_13_3_0:
-    _includes = r"""
-#include <cupy/cuda_workaround.h>  // provide std:: coverage
-"""
-else:
-    _includes = r"""
-#include <type_traits>  // let Jitify handle this
-"""
```
jakirkham, line 23: This was set as a global variable in this file. However it does not appear to be used in this file

Given it is prefixed with `_`, this shouldn't be used outside this file either

Have gone ahead and dropped it

jakirkham: Looks like this is actually used here

https://github.com/rapidsai/cucim/blob/6173a4cf723242cee1bde45cd2d36b5291086984/python/cucim/src/cucim/skimage/measure/_regionprops_gpu_basic_kernels.py#L16

So going to add this back with the top case only

jakirkham:
Am seeing [the following error repeated in CI]( https://github.com/rapidsai/cucim/actions/runs/22197333519/job/64202098064#step:13:1 ):

```python
_ test_median_hist_dtypes[algorithm_kwargs1-sorting-uint8-footprint_shape1-mirror] _
[gw2] linux -- Python 3.11.14 /pyenv/versions/3.11.14/bin/python
python/cucim/src/cucim/skimage/filters/tests/test_median.py:140: in test_median_hist_dtypes
    with expected_warnings(msg):
/pyenv/versions/3.11.14/lib/python3.11/contextlib.py:137: in __enter__
    return next(self.gen)
           ^^^^^^^^^^^^^^
/pyenv/versions/3.11.14/lib/python3.11/site-packages/cucim/skimage/_shared/_warnings.py:132: in expected_warnings
    with all_warnings() as w:
/pyenv/versions/3.11.14/lib/python3.11/contextlib.py:137: in __enter__
    return next(self.gen)
           ^^^^^^^^^^^^^^
/pyenv/versions/3.11.14/lib/python3.11/site-packages/cucim/skimage/_shared/_warnings.py:69: in all_warnings
    mod.__warningregistry__.clear()
    ^^^^^^^^^^^^^^^^^^^^^^^
/pyenv/versions/3.11.14/lib/python3.11/site-packages/cupy_backends/cuda/libs/__init__.py:36: in __getattr__
    return _importlib.import_module(f'cupy_backends.cuda.libs.{name}')
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
/pyenv/versions/3.11.14/lib/python3.11/importlib/__init__.py:126: in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
<frozen importlib._bootstrap>:1204: in _gcd_import
    ???
<frozen importlib._bootstrap>:1176: in _find_and_load
    ???
<frozen importlib._bootstrap>:1140: in _find_and_load_unlocked
    ???
E   ModuleNotFoundError: No module named 'cupy_backends.cuda.libs.__warningregistry__'
_ test_median_hist_dtypes[algorithm_kwargs1-sorting-uint8-footprint_shape1-nearest] _
```

It looks like this is due to the following code interacting with CuPy

https://github.com/rapidsai/cucim/blob/6173a4cf723242cee1bde45cd2d36b5291086984/python/cucim/src/cucim/skimage/_shared/_warnings.py#L67-L71

python/cucim/src/cucim/skimage/_shared/_warnings.py
```diff
@@ -65,10 +65,7 @@ def all_warnings():
     del frame
 
     for mod_name, mod in list(sys.modules.items()):
-        try:
-            mod.__warningregistry__.clear()
-        except AttributeError:
-            pass
+        getattr(mod, "__warningregistry__", {}).clear()
```
jakirkham: Am not entirely sure how we fix the issue above. However this does seem to be an important piece

The [other piece in CuPy itself]( https://github.com/cupy/cupy/blob/76bbe04a55266b33b39fb7cfe1c6b92d414bff3c/cupy_backends/cuda/libs/__init__.py#L40-L41 )

IIUC the cuCIM vendored scikit-image code here is working around a CPython bug ( https://github.com/python/cpython/issues/65923#issuecomment-1093657196 )

However it seems the attribute request is being interpreted by CuPy as an `import`, which fails. Think the logic in CuPy may be treating too many things as submodules

It is unclear whether we can workaround it here. Though we may also want a fix in CuPy as well

Would be interested in hearing other perspectives on this issue

cc @leofang @grlee77

jakirkham: Offline Leo pointed me to this fix that will be in CuPy 14.0.1

Ref:

* https://github.com/cupy/cupy/pull/9724
* https://github.com/cupy/cupy/pull/9707
* https://github.com/cupy/cupy/issues/9704

jakirkham: Given we can't seem to easily workaround this and it will be fixed upstream, will back out this change

Instead will exclude CuPy 14.0.0 from the allowed version. That way users won't be able to accidentally pick it up. They should just get CuPy 14.0.1 instead

Will keep this thread open so it can be found more easily

grlee77: This private `all_warnings` context manager is only used within the `expected_warnings` context manager. That one is also private and used only internally for checking warning messages in various test cases.

It is not modified from the upstream scikit-image function (aside from trivial style formatting changes)

jakirkham: Thanks Greg! 🙏

Yeah we just saw some funky test failures. Turns out this was due to something on the CuPy side, which occurs with CuPy 14.0.0, but will be fixed in CuPy 14.0.1

Have confirmed this by testing locally. So have just excluded CuPy 14.0.0

dependencies.yaml
```diff
@@ -241,7 +241,7 @@ dependencies:
           - scipy>=1.11.2
       - output_types: conda
         packages:
-          - &cupy_unsuffixed cupy>=13.6.0
+          - &cupy_unsuffixed cupy>=13.6.0,!=14.0.0
```
jakirkham, line 244: As discussed above in this thread ( https://github.com/rapidsai/cucim/pull/1031#discussion_r2830243458 ), CuPy 14.0.0 has a bug ( https://github.com/cupy/cupy/issues/9704 ), which cuCIM encountered. More details in thread

CuPy already includes a fix for this bug, which will be included in CuPy 14.0.1

So simply excluded CuPy 14.0.0. This way CuPy 14.0.1 can be picked up without issues

jameslamb: Thanks for the link, agree a `!=14.0.0` is a good choice here.

jakirkham:
Thanks Greg and James! 🙏

jakirkham:
Discussed with the RAPIDS team and we have decided to backport these changes and cut cuCIM 26.02.01 with the fixes

This work is happening in PR: https://github.com/rapidsai/cucim/pull/1036

python/cucim/src/cucim/core/operations/morphology/_pba_2d.py
jakirkham: Should have added beginning with CuPy 14.0.0, [leaving `options` unspecified results in using `-std=c++17`]( https://github.com/cupy/cupy/blob/e0faa4d1abb05132498acbcc9dd00c4d19fd958e/cupy/_core/raw.pyx#L43-L45 ), which includes all of the C++11 features we use and some. So whether CuPy 13 or 14 is used, we no longer need to specify this

PR #1032, opened 2026-02-19
add PR CI job checking that nightly tests are passing

jameslamb:
Contributes to https://github.com/rapidsai/shared-actions/issues/94

Proposes adding a CI job on PRs that will fail if nightly tests haven't run successfully at least once in the last 7 days.

### Why?

Blocking PR CI when nightlies have been failing for too long is intended to make nightly failures a bit louder, to ensure issues are noticed and addressed before the next release.

See https://github.com/rapidsai/build-planning/issues/127 for context.

A check like this is already in place in many RAPIDS repos, including cuDF.

### How long will we have to fix issues?

By default, 7 days.

But that's configurable and teams often extend that based on their current priorities.

### DIFF
--- a/conda/environments/all_cuda-129_arch-aarch64.yaml
+++ b/conda/environments/all_cuda-129_arch-aarch64.yaml
@@ -11,7 +11,7 @@ dependencies:
 - cuda-cudart-dev
 - cuda-nvcc
 - cuda-version=12.9
-- cupy>=13.6.0
+- cupy>=13.6.0,!=14.0.0
 - cxx-compiler
 - gcc_linux-aarch64=14.*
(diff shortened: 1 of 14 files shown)


jakirkham:
Thanks James! 🙏

Ironically the nightly CI jobs just started failing due to a few issues encountered with CuPy 14, which was recently released 😭

Have put together the needed fixes in PR: https://github.com/rapidsai/cucim/pull/1031

If you have time to review, that would be greatly appreciated 🙂

jakirkham:
Thanks James! 🙏

This seems sensible to me

Let's check with the rest of the cuCIM team to see if they have any more thoughts

PR #1034, opened 2026-03-03
Caching for cuslide2

cdinea:
Description

Implements tile-level caching for cuslide2's single-region read_region() path 

Key changes

- Tile-level caching (ifd.cpp)
- Decomposes each read_region() ROI into its constituent TIFF tiles, performs per-tile cache lookups via cuCIM's existing ImageCache API, decodes only cache-miss tiles via nvImageCodec, inserts decoded tiles into the cache, and assembles the final output raster.
- Reuses cuslide's cache infrastructure (create_key, find, lock/unlock, allocate, insert) — no new cache backend code.
- Fixed hash_value_ to include the file handle hash (file_hash ^ splitmix64(ifd_index)) for cross-file cache key uniqueness.
- Edge tiles are clipped to actual image bounds; decode failures fill with background_value.
- Tiles always cached in host memory; GPU output transferred via a single cudaMemcpy(H2D) after assembly.
- Falls back to direct ROI decode when caching is not applicable (strip-based images, out-of-bounds ROI, or no_cache mode).

### DIFF
--- a/README.md
+++ b/README.md
@@ -102,7 +102,7 @@ docker rm -v ${tmp_id}
 ## cuslide2 Plugin
 
 > **Deprecation Notice:** Starting with 26.04, the original `cuslide`
-> plugin will be deprecated in favor of `cuslide2`. `cuslide2` uses
+> plugin will be deprecated in favor of `cuslide2` with removal of `cuslide` planned for 26.08. `cuslide2` uses
 > [nvImageCodec](https://developer.nvidia.com/nvimagecodec) for GPU-accelerated
 > TIFF decoding and adds tile-level caching, async batch decoding, and broader
 > format support. Please plan to migrate existing workflows to `cuslide2`.


cpp/plugins/cucim.kit.cuslide2/src/cuslide/tiff/ifd.cpp
```diff
@@ -452,18 +453,243 @@ bool IFD::read([[maybe_unused]] const TIFF* tiff,
...
+        if (out_device.type() == cucim::io::DeviceType::kCUDA)
+        {
+            uint8_t* gpu_buf = nullptr;
+            cudaError_t err = cudaMalloc(reinterpret_cast<void**>(&gpu_buf), one_raster_size);
+            if (err != cudaSuccess)
+            {
+                cucim_free(host_raster);
+                throw std::runtime_error("Failed to allocate GPU buffer for tile-cached output");
+            }
+            cudaMemcpy(gpu_buf, host_raster, one_raster_size, cudaMemcpyHostToDevice);
```
jakirkham: In the future we may want to look at using stream to make this async

cpp/plugins/cucim.kit.cuslide2/src/cuslide/tiff/ifd.cpp
```diff
@@ -452,18 +453,243 @@ bool IFD::read([[maybe_unused]] const TIFF* tiff,
...
+            }
+
+            dest_row_ptr += static_cast<size_t>(copy_rows) * dest_row_stride;
+        }
+
+        // --- Move assembled raster to the requested output device ---
+        if (out_device.type() == cucim::io::DeviceType::kCUDA)
+        {
+            uint8_t* gpu_buf = nullptr;
+            cudaError_t err = cudaMalloc(reinterpret_cast<void**>(&gpu_buf), one_raster_size);
```
jakirkham: Similarly in the future we could look at making this allocation async

scripts/test_philips_tiff.py
```diff
@@ -350,6 +350,108 @@ def test_philips_tiff(file_path, plugin_lib):
...
+            # Enable per-process cache (256 MB) and stat recording.
+            # Re-open the image after configuring the cache so the
+            # IFD picks up the active cache manager.
+            CuImage.cache("per_process", memory_capacity=256)
+            CuImage.cache().record(True)
+            img_cached = CuImage(file_path)
+            print(f"  Cache type: {CuImage.cache().type}")
+            print(f"  Stat recording: {CuImage.cache().record()}")
+
+            import numpy as np
```
jakirkham: Can we move this `import` to the top-level?

scripts/test_philips_tiff.py
```diff
@@ -350,6 +350,108 @@ def test_philips_tiff(file_path, plugin_lib):
...
+            # --- Summary ---
+            print("\n  📊 Cache summary:")
+            print(f"     Total hits:   {CuImage.cache().hit_count}")
+            print(f"     Total misses: {CuImage.cache().miss_count}")
+            print(f"     Cache size:   {CuImage.cache().size} tiles")
+            print("  ✅ Tile-level caching test passed!")
+
+    except Exception as e:
+        print(f"  ❌ Caching test failed: {e}")
+        import traceback
```
jakirkham: Same with this `import`

scripts/test_aperio_svs.py
```diff
@@ -265,6 +265,107 @@ def test_batch_decode_performance(img, level_dimensions):
...
+
+    NOTE: Tile-level caching requires that the cuslide2 plugin
+    successfully extracts TileWidth/TileLength TIFF tags via
+    nvImageCodec (>= 0.7.0). If tile sizes are (0,0), the
+    caching code path is not entered and the test is skipped.
+
+    Raises:
+        RuntimeError: If caching assertions fail.
+    """
+    import numpy as np
```
jakirkham: Let's move this out of the function and to the top-level

cpp/plugins/cucim.kit.cuslide2/src/cuslide/tiff/ifd.cpp
cdinea: thank you for the feedback @jakirkham - I agree, cudaMemcpyAsync with a stream would allow overlapping the H2D transfer with other work. For now this only runs on the single-location tile-cached path (location_len == 1, batch_size == 1), so there's no pipeline to overlap with, but it would be a worthwhile optimization if we later add prefetching or pipelining to this path. The multi-location batch path already avoids this copy entirely via zero-copy OutputBufferProvider - keepign track of thsi RFE here https://github.com/rapidsai/cucim/issues/1064

cpp/plugins/cucim.kit.cuslide2/src/cuslide/tiff/ifd.cpp
cdinea: thank you for the feedback @jakirkam - do you suggest to add cudaMallocAsync here?the batch decode path already uses a custom CUDA stream (decode_stream_ in NvImageCodecProcessor, created with cudaStreamNonBlocking) for async operations. This single-location tile-cached path doesn't have a stream because it runs synchronously in IFD::read() outside the NvImageCodecProcessor pipeline. Pllumbing a stream into this path (or using cudaMallocAsync / cudaMemcpyAsync) would be a natural follow-up - tracking this here https://github.com/rapidsai/cucim/issues/1065

scripts/test_philips_tiff.py
cdinea: thank you @jakirkham - addressed this feedback here https://github.com/rapidsai/cucim/pull/1034/commits/fbb7e6e6de423451e6bb5f40e2f8e20678ba2751

scripts/test_philips_tiff.py
cdinea: thank you for the feedback @jakirkham - addressed this feedback here https://github.com/rapidsai/cucim/pull/1034/commits/a2cc7208134d1f872c794e1d5c883940920b7453

scripts/test_aperio_svs.py
cdinea: thanky ou for the feedback @jakirkham - feedback already addressed in commit fbb7e6e

cpp/plugins/cucim.kit.cuslide2/src/cuslide/nvimgcodec/nvimgcodec_decoder.h
```diff
@@ -173,6 +173,16 @@ std::vector<BatchDecodeResult> decode_batch_regions_nvimgcodec(
...
+/**
+ * Schedule batch decoding with optional caller-provided output buffers.
+ *
+ * When @p output_buffers is non-empty the decoder writes directly into those
+ * buffers (one per region, same order).  The caller retains ownership; the
+ * returned BatchDecodeState will NOT free them on destruction.
+ *
+ * When @p output_buffers is empty (default), the function allocates its own
+ * buffers, which are freed by wait_batch_decode() or ~BatchDecodeState.
+ */
```
gigony: I think now this block is redundant. Could you please remove these lines or refactor line 152-175 with the updated description?

cpp/plugins/cucim.kit.cuslide2/src/cuslide/tiff/ifd.cpp
```diff
@@ -452,18 +453,243 @@ bool IFD::read([[maybe_unused]] const TIFF* tiff,
...
+
+                        #ifdef DEBUG
+                        ::fmt::print("  📥 Cache MISS tile[{},{}] idx={} decoded {}x{}\n",
+                                     tile_col, tile_row, tile_index, actual_tw, actual_th);
+                        #endif
+                    }
+                    else
+                    {
+                        image_cache.unlock(index_hash);
+                        #ifdef DEBUG
```
gigony, line 659: After unlock, need to free `tile_buf` (from line 582) to avoid memory leak.

Please release tile_buf in the failure branch before continue

cpp/plugins/cucim.kit.cuslide2/src/cuslide/tiff/ifd.cpp
```diff
@@ -452,18 +453,243 @@ bool IFD::read([[maybe_unused]] const TIFF* tiff,
...
+        const uint32_t pixel_offset_sy = static_cast<uint32_t>(sy % th);
+        const uint32_t pixel_offset_ey = static_cast<uint32_t>(ey % th);
+
+        // Tiles per row in the IFD tile grid
+        const uint32_t stride_y = (width_ + tw - 1) / tw;
+
+        const uint64_t ifd_hash = hash_value_;
+
+        // Allocate host output buffer for tile assembly
+        uint8_t* host_raster = static_cast<uint8_t*>(cucim_malloc(one_raster_size));
```
jantonguirao: suggestion: use RAII for host_raster (std::unique_ptr<uint8_t, CucimFreeDeleter>) so any throw frees it.

cpp/plugins/cucim.kit.cuslide2/src/cuslide/tiff/ifd.cpp
```diff
@@ -452,18 +453,243 @@ bool IFD::read([[maybe_unused]] const TIFF* tiff,
...
+                        image_cache.unlock(index_hash);
+
+                        #ifdef DEBUG
+                        ::fmt::print("  📥 Cache MISS tile[{},{}] idx={} decoded {}x{}\n",
+                                     tile_col, tile_row, tile_index, actual_tw, actual_th);
+                        #endif
+                    }
+                    else
+                    {
+                        image_cache.unlock(index_hash);
```
jantonguirao: `tile_buf` was allocated via `image_cache.allocate()` on line 582 (which resolves to `cucim_malloc` for CPU caches) but is never freed on this failure path. It is neither inserted into the cache (so the cache destructor won't free it) nor explicitly freed before `continue`. Add a `cucim_free(tile_buf)` here. Or even better, use a RAII wrapper.

Note: `ImageCache` has no `deallocate()` counterpart to `allocate()` — adding one would make this pattern safer.

cpp/plugins/cucim.kit.cuslide2/src/cuslide/tiff/ifd.cpp
jantonguirao: The tile-caching path allocates its own `host_raster` here and ignores any pre-allocated buffer that may have been set in `output_buffer` at line 454. In practice `request->buf` is never populated today (the pybind11 binding hardcodes `nullptr`), so this is not currently triggered. However, if `buf` support is ever wired up for the single-region path, the tile-caching branch will silently write to the wrong buffer.

cpp/plugins/cucim.kit.cuslide2/src/cuslide/tiff/ifd.cpp
jantonguirao: `cudaMemcpy` return value is not checked. The `cudaMalloc` just above (line 643) is checked and throws on failure — this should be consistent. A silent failure here leaves `output_buffer` pointing at uninitialized GPU memory with no error surfaced to the caller.

cpp/plugins/cucim.kit.cuslide2/src/cuslide/tiff/ifd.cpp
```diff
@@ -452,18 +453,243 @@ bool IFD::read([[maybe_unused]] const TIFF* tiff,
...
+                    tile_data = static_cast<uint8_t*>(cached_value->data);
+
+                    #ifdef DEBUG
+                    ::fmt::print("  ♻️  Cache HIT  tile[{},{}] idx={}\n", tile_col, tile_row, tile_index);
+                    #endif
+                }
+                else
+                {
+                    // *** Cache miss — decode this tile via nvImageCodec ***
+                    uint8_t* tile_buf = static_cast<uint8_t*>(image_cache.allocate(tile_buf_size));
```
jantonguirao: `image_cache.allocate()` can return null on OOM (it calls `cucim_malloc` for CPU caches, which can fail). The null pointer is then passed directly to the decoder with no guard. Add a null check before the decode call.

scripts/test_philips_tiff.py
```diff
@@ -345,8 +344,104 @@ def test_philips_tiff(file_path, plugin_lib):
...
+    # Exercises the tile-level cache in ifd.cpp: ROI -> tile grid
+    # decomposition, per-tile cache lookup, miss-decode-insert,
+    # and warm-read cache hits.
+    #
+    # NOTE: Tile-level caching requires that the cuslide2 plugin
+    # successfully extracts TileWidth/TileLength TIFF tags via
+    # nvImageCodec (>= 0.7.0). If tile sizes are (0,0), the
+    # caching code path is not entered and the test is skipped.
+    # ==================================================================
+    print("💾 Testing tile-level caching...")
```
jantonguirao: The caching test in `test_aperio_svs.py` was extracted into a clean `test_tile_level_caching()` function. Here the same logic is copy-pasted inline inside a `try/except`. Consider moving the shared logic to a helper (e.g. in a `test_common.py`) so both scripts stay in sync.

cpp/plugins/cucim.kit.cuslide2/src/cuslide/tiff/ifd.cpp
```diff
@@ -452,18 +453,243 @@ bool IFD::read([[maybe_unused]] const TIFF* tiff,
...
+                const size_t tile_buf_size = static_cast<size_t>(actual_tw) * actual_th * samples;
+                const uint32_t tile_row_stride = actual_tw * samples;
+
+                // Hash for per-tile locking
+                const uint64_t index_hash = ifd_hash ^
+                    (static_cast<uint64_t>(tile_index) | (static_cast<uint64_t>(tile_index) << 32));
+
+                // --- Cache lookup ---
+                auto key = image_cache.create_key(ifd_hash, tile_index);
+                image_cache.lock(index_hash);
```
mkepa-nv: This interleaved usage of `index_hash` and `key` looks error prone. Maybe in a future MPR it would be better to use `key` for `lock()`?

cpp/plugins/cucim.kit.cuslide2/src/cuslide/tiff/ifd.cpp
```diff
@@ -452,18 +453,243 @@ bool IFD::read([[maybe_unused]] const TIFF* tiff,
...
+                    }
+                }
+
+                // --- Assembly: copy relevant sub-rectangle from tile into output ---
+                const uint32_t src_byte_offset =
+                    (tile_pixel_sy * actual_tw + tile_pixel_sx) * samples;
+
+                for (uint32_t r = 0; r < copy_rows; ++r)
+                {
+                    memcpy(dest_row_ptr + dest_col_byte_offset + r * dest_row_stride,
```
mkepa-nv, line 683: So this caching works only on the CPU, right? How often do you output to CPU memory in cuCIM? I am bit worried that for light codecs it may be faster to decode each time than do to pixel by pixel copy on the CPU. 
This isn't a big deal as user can always disable caching, but it would be good to have some benchmarks that show if there is speed if the caching is in the CPU

cpp/plugins/cucim.kit.cuslide2/src/cuslide/tiff/ifd.cpp
```diff
@@ -452,18 +453,243 @@ bool IFD::read([[maybe_unused]] const TIFF* tiff,
...
+            cudaError_t err = cudaMalloc(reinterpret_cast<void**>(&gpu_buf), one_raster_size);
+            if (err != cudaSuccess)
+            {
+                cucim_free(host_raster);
+                throw std::runtime_error("Failed to allocate GPU buffer for tile-cached output");
+            }
+            cudaMemcpy(gpu_buf, host_raster, one_raster_size, cudaMemcpyHostToDevice);
+            cucim_free(host_raster);
+
+            output_buffer = gpu_buf;
```
mkepa-nv: As Joaquin noted in https://github.com/rapidsai/cucim/pull/1034/changes#r3014427070, you ignore already allocated buffer from `out_buf->data`, it would be good to fix that to prevent future bugs

scripts/test_aperio_svs.py
```diff
@@ -265,6 +267,105 @@ def test_batch_decode_performance(img, level_dimensions):
...
+        f"     Hits: {warm_hits} (+{new_hits}), Misses: {warm_misses} (+{new_misses})"
+    )
+    print(f"     Time: {warm_time * 1000:.1f} ms")
+    assert new_hits > 0, "Expected cache hits on warm read"
+    assert new_misses == 0, "Expected zero new misses on warm read"
+
+    # --- Data correctness: cold vs warm must match ---
+    arr_cold = np.asarray(region_cold)
+    arr_warm = np.asarray(region_warm)
+    assert np.array_equal(arr_cold, arr_warm), "Cold and warm reads differ!"
```
mkepa-nv: I would also compare cached vs non-cached decode.

scripts/test_aperio_svs.py
```diff
@@ -265,6 +267,105 @@ def test_batch_decode_performance(img, level_dimensions):
...
+    print(f"     ✅ Data matches (shape={arr_cold.shape})")
+
+    # --- Overlapping read (partial hit: some tiles shared) ---
```
mkepa-nv: This doesn't really test anything in CI - I would add some asserts. And maybe make some of the image outside the cached region, so there are both cache hits and misses

cpp/plugins/cucim.kit.cuslide2/src/cuslide/nvimgcodec/nvimgcodec_decoder.h
cdinea: thank you for the review @gigony - addressed this in this commit https://github.com/rapidsai/cucim/pull/1034/commits/d784aff0040783e09939a993fe482896b57481e2

cpp/plugins/cucim.kit.cuslide2/src/cuslide/tiff/ifd.cpp
cdinea: thank you for the feedback @gigony - addressed this in this commit https://github.com/rapidsai/cucim/pull/1034/commits/ac9099cb4c474215b30414a838b67a1ec66a47cf

cpp/plugins/cucim.kit.cuslide2/src/cuslide/tiff/ifd.cpp
cdinea: thanky ou for the feedback @jantonguirao - wrapped host_raster in RAII in this commit https://github.com/rapidsai/cucim/pull/1034/commits/838ae2a9c514dffdee36d0bf9a3ef969a4ab5d44

cpp/plugins/cucim.kit.cuslide2/src/cuslide/tiff/ifd.cpp
cdinea: thank you for the feedback @jantonguirao - this was indeed a leak. Already fixed: cucim_free(tile_buf) is now called on the failure path  before continue. this was addressed in this commit ac9099c

cpp/plugins/cucim.kit.cuslide2/src/cuslide/tiff/ifd.cpp
cdinea: thank you for the feedback @jantonguirao - addressed this feedback in this commit https://github.com/rapidsai/cucim/pull/1034/commits/3442abb93d3038ca4633597fcd36bff79bff79aa

cpp/plugins/cucim.kit.cuslide2/src/cuslide/tiff/ifd.cpp
cdinea: thank you for the feedback @jantonguirao - address this in this commit https://github.com/rapidsai/cucim/pull/1034/commits/b4cd4af5f794601c337560e7081f45cd07d1ff33

cpp/plugins/cucim.kit.cuslide2/src/cuslide/tiff/ifd.cpp
cdinea: thank you for the feedback @jantonguirao - addressed this in this commit https://github.com/rapidsai/cucim/pull/1034/commits/c9a3d9d07cf2dfbe2f6bd5ca5dafaf2ed77c038e

scripts/test_philips_tiff.py
cdinea: thank you for the feedback @jantonguirao - addressed this in this commit https://github.com/rapidsai/cucim/pull/1034/commits/4d9462ab1eeeb84d7142365038c7e9045bc5154e

cpp/plugins/cucim.kit.cuslide2/src/cuslide/tiff/ifd.cpp
cdinea: thank you for the feedback @mkepa-nv - tracking this as a RFE here https://github.com/rapidsai/cucim/issues/1066

cpp/plugins/cucim.kit.cuslide2/src/cuslide/tiff/ifd.cpp
cdinea: thank you fr the question @mkepa-nv - this is a good question, the assembly isn't pixel-by-pixel — it's memcpy per row of the sub-rectangle from each cached tile (copy_bytes_per_row = copy_cols * samples). The cache stores decoded tiles in CPU memory; for GPU output, the final assembled raster is copied to the device via cudaMemcpy(H2D). For compressed WSI formats , ddecode cost dominates and warm cache reads avoid re-decompression entirely. For uncompressed/light codecs the benefit is marginal, but the cache defaults to disabled (kNoCache) so users only pay this cost when they explicitly enable it. Happy to add benchmarks comparing cached vs uncached warm reads across compression types

cpp/plugins/cucim.kit.cuslide2/src/cuslide/tiff/ifd.cpp
cdinea: i agree @mkepa-nv - this is fixed in this commit 3442abb — the tile-caching path now honors out_buf->data when provided

scripts/test_aperio_svs.py
cdinea: thank you for the feedback @mkepa-nv - this is a good idea, we now test cached vss non cached code : we capture a direct decode before enabling the cache, then compares pixel-for-pixel against the cached cold read- this should catches any bugs in the tile grid decomposition or row-wise assembly that produce different output from the direct nvImageCodec decode path

scripts/test_aperio_svs.py
cdinea: thank you for the feedback @mkepa-nv - I think that this is a good idea, the overlapping read test now 

- asserts cache hits > 0 — tiles shared with the previous (0,0) read must come from cache
- asserts cache misses > 0 — tiles outside the previous region must trigger new decodes
- pixel-for-pixel comparison — the cached overlapping result is compared against a non-cached direct decode of the same (128,128)→(512,512) region, with max pixel diff reported on failure

in this commit https://github.com/rapidsai/cucim/pull/1034/commits/fd54e0c698e6bb494e3bc58652aa50a77aef6bec

README.md
```diff
@@ -99,9 +99,13 @@ docker cp $tmp_id:/input notebooks
 docker rm -v ${tmp_id}
 ```
 
-## cuslide2 Plugin (Experimental)
+## cuslide2 Plugin
 
-cuCIM includes an experimental `cuslide2` plugin that uses [nvImageCodec](https://developer.nvidia.com/nvimagecodec) for GPU-accelerated TIFF decoding. This plugin provides enhanced performance for batch ROI decoding operations.
+> **Deprecation Notice:** Starting with the next release, the original `cuslide`
```
jakirkham: Think we should be explicit on the release of the deprecation to make this clear to users

```suggestion
> **Deprecation Notice:** Starting with 26.04, the original `cuslide`
```

jakirkham (approved):
Thanks Cristiana! 🙏

Only remaining suggestion from me is to put the version of the deprecation

Anything that remains unaddressed would make sure to file an issue so we can continue to iterate on the good feedback we have gotten here (thanks all for taking the time to look and share your thoughts 🙏)

jantonguirao (approved):

scripts/test_aperio_svs.py
mkepa-nv: @cdinea Thanks for the change!

scripts/test_aperio_svs.py
mkepa-nv: @cdinea thanks for the change!

I am not sure if such verbose checking for pixels correctness is needed. IMO something like this should be enough:
`assert np.array_equal(arr_cold, arr_arr_direct), "Cold and non cached reads differ!"`. But this is just a nitpick, it can stay how it is right now

cpp/plugins/cucim.kit.cuslide2/src/cuslide/tiff/ifd.cpp
mkepa-nv: @cdinea 
> the assembly isn't pixel-by-pixel

This was just a mental shortcut on my side - memcpy under the hood is doing byte by byte (or word by word) copy on a single threaded. So this could be slower that decode on the gpu (which for sure have more work to do, but is massively parallel).

If this is disabled by default, I don't have any more concerns. It is always good to have benchmarks, but there is no hurry to add them now, it can be done later

cpp/plugins/cucim.kit.cuslide2/src/cuslide/tiff/ifd.cpp
```diff
@@ -452,18 +454,298 @@ bool IFD::read([[maybe_unused]] const TIFF* tiff,
...
+
+        // If the caller provided a pre-allocated buffer, assemble directly into it;
+        // otherwise allocate our own host buffer (RAII-managed).
+        const bool caller_owns_buffer = is_buf_available && output_buffer != nullptr;
+        std::unique_ptr<uint8_t, decltype(cucim_free)*> host_raster_owner(nullptr, cucim_free);
+        uint8_t* host_raster = nullptr;
+
+        if (caller_owns_buffer)
+        {
+            host_raster = output_buffer;
```
mkepa-nv, line 532: @cdinea The user provided buffer can be on device right? Then this code will not work (it assumes `host_raster`  is in host memory, as it uses `memcpy` and `memset` to work with it

mkepa-nv (changes requested):
I think there is a bug now with how user provided buffer is used. IIRC Joaquin noted before that user provided buffer is never used currently? But it would be good to fix that anyway

cpp/plugins/cucim.kit.cuslide2/src/cuslide/tiff/ifd.cpp
mkepa-nv: @cdinea but from the perf perspective maybe it would be better to have caching tiles decoded in gpu memory? Or have two caches, one for CPU and one for GPU tiles?

cpp/plugins/cucim.kit.cuslide2/src/cuslide/tiff/ifd.cpp
cdinea: thank you for the feedback @mkepa-nv - i added a caller_buffer_on_device check: if the caller's buffer is on CUDA, a temporary host staging buffer is allocated for assembly, and the result is cudaMemcpy'd into the caller's device buffer at the end. Only when the caller's buffer is on the CPU does the code write into it directly .pushed this change in this commit https://github.com/rapidsai/cucim/pull/1034/commits/ccdb2b6f2c2e8f724676c00c22d099622e75908f

scripts/test_aperio_svs.py
cdinea: thank you for the feedback @mkepa-nv - simplified the assertions in this commit https://github.com/rapidsai/cucim/pull/1034/commits/f032db21e606d9b01880401352be5b8130f7c546

cpp/plugins/cucim.kit.cuslide2/src/cuslide/tiff/ifd.cpp
cdinea: thank you for the feedback @mkepa-nv -  A GPU tile cache would eliminate the cudaMemcpy overhead on warm reads and let the assembly happen as a device kernel. We'll track this as a follow-up optimization iin this RFE https://github.com/rapidsai/cucim/issues/1069

README.md
```diff
@@ -99,9 +99,13 @@ docker cp $tmp_id:/input notebooks
 docker rm -v ${tmp_id}
 ```
 
-## cuslide2 Plugin (Experimental)
+## cuslide2 Plugin
 
-cuCIM includes an experimental `cuslide2` plugin that uses [nvImageCodec](https://developer.nvidia.com/nvimagecodec) for GPU-accelerated TIFF decoding. This plugin provides enhanced performance for batch ROI decoding operations.
+> **Deprecation Notice:** Starting with 26.04, the original `cuslide`
+> plugin will be deprecated in favor of `cuslide2`. `cuslide2` uses
```
jakirkham:
```suggestion
> **Deprecation Notice:** Starting with 26.04, the original `cuslide`
> plugin will be deprecated in favor of `cuslide2` with removal of `cuslide` planned for 26.08. `cuslide2` uses
```

Edit: Changed removal version from 26.06 to 26.08 based on Gigon's suggestion offline

jakirkham: cc @gigony @cdinea (please share your feedback)

cdinea: looks good to me, thank you @jakirkham

PR #1035, opened 2026-03-03
make test_load_image_metadata test case insensitive to key order

grlee77:
One test case is currently comparing image metadata using a string comparison that is sensitive to key order.

I think we only care that the same keys and values are present, but not the order, so this MR uses `json.loads` to convert the string to a Python `dict`.

### DIFF
--- a/python/cucim/tests/unit/clara/test_load_image_metadata.py
+++ b/python/cucim/tests/unit/clara/test_load_image_metadata.py
@@ -1,8 +1,9 @@
 #
-# SPDX-FileCopyrightText: Copyright (c) 2021-2022, NVIDIA CORPORATION.
+# SPDX-FileCopyrightText: Copyright (c) 2021-2026, NVIDIA CORPORATION.
 # SPDX-License-Identifier: Apache-2.0
 #
 
+import json
 import math
 
 import pytest
@@ -65,7 +66,9 @@ def test_load_image_metadata(testimg_tiff_stripe_32x24_16):
     assert isinstance(metadata, dict)
     assert len(metadata) == 2  # 'cucim' and 'tiff'
     # A raw metadata string.
-    assert img.raw_metadata == '{"axes": "YXC", "shape": [24, 32, 3]}'
+    assert json.loads(img.raw_metadata) == json.loads(
+        '{"axes": "YXC", "shape": [24, 32, 3]}'
+    )
 
 
 def test_load_image_resolution_metadata(


python/cucim/tests/unit/clara/test_load_image_metadata.py
```diff
@@ -65,7 +66,9 @@ def test_load_image_metadata(testimg_tiff_stripe_32x24_16):
     assert isinstance(metadata, dict)
     assert len(metadata) == 2  # 'cucim' and 'tiff'
     # A raw metadata string.
-    assert img.raw_metadata == '{"axes": "YXC", "shape": [24, 32, 3]}'
+    assert json.loads(img.raw_metadata) == json.loads(
+        '{"axes": "YXC", "shape": [24, 32, 3]}'
+    )
```
jakirkham, line 71: For more context, we recently started seeing [this failure in the nightlies]( https://github.com/rapidsai/cucim/actions/runs/22563164914/job/65353826578#step:12:25117 ):

```
FAILED tests/unit/clara/test_load_image_metadata.py::test_load_image_metadata[testimg_tiff_stripe_32x24_16_jpeg] - assert '{"shape": [2...axes": "YXC"}' == '{"axes": "YX... [24, 32, 3]}'
  
  - {"axes": "YXC", "shape": [24, 32, 3]}
  + {"shape": [24, 32, 3], "axes": "YXC"}
FAILED tests/unit/clara/test_load_image_metadata.py::test_load_image_metadata[testimg_tiff_stripe_32x24_16_deflate] - assert '{"shape": [2...axes": "YXC"}' == '{"axes": "YX... [24, 32, 3]}'
  
  - {"axes": "YXC", "shape": [24, 32, 3]}
  + {"shape": [24, 32, 3], "axes": "YXC"}
FAILED tests/unit/clara/test_load_image_metadata.py::test_load_image_metadata[testimg_tiff_stripe_32x24_16_raw] - assert '{"shape": [2...axes": "YXC"}' == '{"axes": "YX... [24, 32, 3]}'
  
  - {"axes": "YXC", "shape": [24, 32, 3]}
  + {"shape": [24, 32, 3], "axes": "YXC"}
```

The results are the same except for the order of the keys-value pairs. The code here has been unchanged since July 2021

Comparing [the last passing nightly build]( https://github.com/rapidsai/cucim/actions/runs/22294539736/job/64547128969#step:13:232 ) and [the first failing nightly build]( https://github.com/rapidsai/cucim/actions/runs/22384202869/job/64791315395#step:13:232 ), it appears that [`tifffile` recently released `2026.2.24`]( https://github.com/cgohlke/tifffile/releases/tag/v2026.2.24 ), which may be the cause. Though have not dug further yet

In any event, it makes sense to compare these as `dict`s instead of as `str`s to avoid being sensitive to the ordering

jakirkham:
CI is failing due to GitHub network issues: https://www.githubstatus.com/incidents/n07yy1bk6kc4

Let's retry after that clears up

jakirkham:
Looks like the GitHub issue may be clearing up. Also CI jobs finished. Have restarted the failing ones

python/cucim/tests/unit/clara/test_load_image_metadata.py
jakirkham, line 71: Would it make sense to use a `dict` for the comparison?

```suggestion
    assert json.loads(img.raw_metadata) == {"axes": "YXC", "shape": [24, 32, 3]}'
```

jameslamb:
admin-merging this per offline conversation w/ @jakirkham , since it's passing everything else but `check-nightly-ci`.

PR #1038, opened 2026-03-04
Update cuCIM for consistency with scikit-image 0.26

grlee77:
This PR ports recent API and implementation changes from scikit-image 0.26.

### Key Changes
- There is one new keyword argument, `boundary_mode` added to `binary_blobs`.
- There was some minor refactoring in `regionprops` and addition of one new property, "intensity_median".
- There are some new deprecations / removals
  - Previously deprecated `shift_x` and `shift_y` arguments to `erosion` and `dilation` are now removed
  - For `remove_small_objects`, the parameter `min_size` is deprecated in favor of `max_size`
  - For `remove_small_holes`, the parameter `area_threshold` is deprecated in favor of `max_size`
- many misc docstring fixes
- large refactor of geometric transforms code to match the scikit-image 0.26 behavior/API
  - most changes to `_shared/utils.py` are additional helpers related to this refactor (also ported from scikit-image 0.26)
- `cucim.skimage.segmentation` also now has a separate `__init__.pyi` for lazy loading 
- some existing test cases now load the local copies of test data from the repository instead of relying on `fetch` (fetch no longer works for these files in scikit-image 0.26).

### Possibly follow-up tasks
- provide batch implementation for `intensity_median` in `_regionprops_gpu.py`
- make decision on deprecating `cucim.skimage.morphology.binary_*` functions
  - These are deprecated now in scikit-image 0.26.
  - For cuCIM, the binary variants are around 2x faster so I kept them for now. A follow-up task would be to update the grayscale variants to automatically dispatch to the faster binary kernel when possible. Then we could remove the separate `binary_*` functions without losing performance.

### Directory layout change in scikit-image 0.26
- scikit-image has switched to separate `src/skimage` and `tests/skimage` folder layout.
- I have currently kept the existing folder layout in cuCIM (`tests` subfolder within each `cucim.skimage` module)

### DIFF
--- a/python/cucim/src/cucim/skimage/transform/tests/test_geometric.py
+++ b/python/cucim/src/cucim/skimage/transform/tests/test_geometric.py
@@ -115,15 +115,16 @@ def test_euclidean_estimation():
     ((EuclideanTransform, False), (SimilarityTransform, True)),
 )
 def test_3d_euclidean_similarity_estimation(tform_class, has_scale):
-    src_points = cp.random.rand(1000, 3)
+    rng = np.random.default_rng(1234)
+    src_points = cp.asarray(rng.random((1000, 3)))
 
     # Random transformation for testing
-    angles = cp.random.random((3,)) * 2 * np.pi - np.pi
+    angles = cp.asarray(rng.random((3,))) * 2 * np.pi - np.pi
     rotation_matrix = cp.array(_euler_rotation_matrix(angles))
     if has_scale:
-        scale = cp.random.randint(0, 20)
+        scale = rng.integers(1, 20)
         rotation_matrix = rotation_matrix * scale
-    translation_vector = cp.random.random((3,))
+    translation_vector = cp.asarray(rng.random((3,)))
     dst_points = []
     for pt in src_points:
         pt_r = pt.reshape(3, 1)


python/cucim/src/cucim/skimage/_shared/tests/test_utils.py
```diff
@@ -1,7 +1,7 @@
 # SPDX-FileCopyrightText: 2009-2022 the scikit-image team
-# SPDX-FileCopyrightText: Copyright (c) 2021-2025, NVIDIA CORPORATION. All rights reserved.
+# SPDX-FileCopyrightText: Copyright (c) 2021-2026, NVIDIA CORPORATION. All rights reserved.
 # SPDX-License-Identifier: Apache-2.0 AND BSD-3-Clause
-
+import re
```
jakirkham, line 4: Thinks we should keep this spaced out from the license header

```suggestion

import re
```

python/cucim/src/cucim/skimage/_shared/compat.py
```diff
@@ -28,6 +28,7 @@
 # check CuPy instead of SciPy
 CUPY_LT_14 = parse(cp.__version__) < parse("14.0.0a1")
 
+
 # Starting in SciPy v1.12, 'scipy.sparse.linalg.cg' keyword argument `tol` is
```
jakirkham, line 32: Just out-of-curiosity was this to address a linter issue or something?

python/cucim/src/cucim/skimage/color/tests/test_delta_e.py
```diff
@@ -78,8 +82,7 @@ def load_ciede2000_data():
     ]
 
     # note: ciede_test_data.txt contains several intermediate quantities
-    path = fetch("color/tests/ciede2000_test_data.txt")
-    return np.loadtxt(path, dtype=dtype)
```
jakirkham, line 82: Why do we no longer need to fetch this test data? Is that done somewhere else?

python/cucim/src/cucim/skimage/exposure/exposure.py
```diff
@@ -74,7 +74,7 @@ def _bincount_histogram(image, source_range, bin_centers=None):
     ----------
     image : array
         Input image.
-    source_range : string
+    source_range : {'image', 'dtype'}
```
jakirkham, line 77: So is this a `dict` now?

python/cucim/src/cucim/skimage/exposure/exposure.py
```diff
@@ -697,9 +698,9 @@ def adjust_gamma(image, gamma=1, gain=1):
 
     Examples
     --------
-    >>> from skimage import data
+    >>> from skimage as ski
     >>> from cucim.skimage import exposure, img_as_float
-    >>> image = img_as_float(cp.array(data.moon()))
+    >>> image = img_as_float(cp.array(ski.data.moon()))
```
jakirkham, line 703: Was scikit-image's layout changed recently?

python/cucim/src/cucim/skimage/feature/_basic_features.py
```diff
@@ -48,7 +49,7 @@ def _mutiscale_basic_features_singlechannel(
     sigma_min=0.5,
     sigma_max=16,
     num_sigma=None,
-    num_workers=None,
+    workers=None,
```
jakirkham, line 52: Guessing this didn't need a deprecation as it was internal facing only

python/cucim/src/cucim/skimage/morphology/binary.py
```diff
@@ -48,11 +48,30 @@ def _iterate_binary_func(binary_func, image, footprint, out, border_value):
     return out
 
 
+# Note: scikit-image 0.26 deprecated the binary_* functions.
+#
+# For cuCIM, we need to consider whether or not to do the same. For
+# scikit-image it seems grayscale morphology is equal or faster to binary, but
+# for cuCIM performance is better with these binary variants.
```
jakirkham, line 55: Ah interesting! How do these functions compare to [the ones in CuPy]( https://docs.cupy.dev/en/stable/reference/scipy_ndimage.html#morphology )? Do they overlap? Are there perf or behavior differences?

python/cucim/src/cucim/skimage/morphology/gray.py
```diff
@@ -63,73 +60,6 @@ def _iterate_gray_func(gray_func, image, footprints, out, mode, cval):
     return out
 
 
-def _shift_footprint(footprint, shift_x, shift_y):
-    """Shift the binary image `footprint` in the left and/or up.
-
-    This only affects 2D footprints with even number of rows
-    or columns.
```
jakirkham, line 70: Is this part of the deprecation above?

python/cucim/src/cucim/skimage/measure/_regionprops_gpu_convex.py
```diff
@@ -137,7 +139,10 @@ def regionprops_feret_diameter_max(
     width less than the one computed by scikit-image.
     """
     if not isinstance(images_convex, Sequence):
-        raise ValueError("Expected `images_convex` to be a sequence of images")
+        raise ValueError(
+            "Expected `images_convex` to be a sequence of images",
+            stacklevel=2,
+        )
```
gigony: ValueError does not accept a stacklevel keyword argument. Python will silently pack stacklevel=2 into the exception's args tuple, producing a misleading repr like ValueError("Expected...", 2). This should just be:

```
raise ValueError("Expected `images_convex` to be a sequence of images")
```


```suggestion
        raise ValueError("Expected `images_convex` to be a sequence of images")
```

python/cucim/src/cucim/skimage/exposure/exposure.py
```diff
@@ -697,9 +698,9 @@ def adjust_gamma(image, gamma=1, gain=1):
 
     Examples
     --------
-    >>> from skimage import data
+    >>> from skimage as ski
```
gigony:
```suggestion
    >>> import skimage as ski
```

python/cucim/src/cucim/skimage/morphology/isotropic.py
```diff
@@ -94,6 +95,26 @@ def isotropic_erosion(image, radius, out=None, spacing=None):
         and thresholding of distance maps, Pattern Recognition Letters,
         Volume 13, Issue 3, 1992, Pages 161-166.
         :DOI:`10.1016/0167-8655(92)90055-5`
+
+    Examples
+    --------
+    Erosion shrinks bright regions
+
+    >>> import cumpy as cp
```
gigony:
```suggestion
    >>> import cupy as cp
```

python/cucim/src/cucim/skimage/measure/_regionprops.py
```diff
@@ -1269,68 +1338,70 @@ def regionprops(
...
     **feret_diameter_max** : float
         Maximum Feret's diameter computed as the longest distance between
         points around a region's convex hull contour as determined by
         ``find_contours``. [5]_
     **image** : (H, J) ndarray
-        Sliced binary region image which has the same size as bounding box.
+        Binary region image sliced by the bounding box.
     **image_convex** : (H, J) ndarray
-        Binary convex hull image which has the same size as bounding box.
+        Binary convex hull image slicd by bounding box.
```
gigony:
```suggestion
        Binary convex hull image sliced by bounding box.
```

python/cucim/src/cucim/skimage/measure/_regionprops.py
```diff
@@ -1269,68 +1338,70 @@ def regionprops(
...
     **image_filled** : (H, J) ndarray
-        Binary region image with filled holes which has the same size as
-        bounding box.
-    **image_intensity** : ndarray
-        Image inside region bounding box.
+        Binary region image with filled holes sliced by bounding box.
+    **image_intensity** : (H, J) ndarray
+        Intensity image isliced by bounding box.
```
gigony:
```suggestion
        Intensity image sliced by bounding box.
```

python/cucim/src/cucim/skimage/morphology/isotropic.py
```diff
@@ -158,6 +180,26 @@ def isotropic_dilation(image, radius, out=None, spacing=None):
         and thresholding of distance maps, Pattern Recognition Letters,
         Volume 13, Issue 3, 1992, Pages 161-166.
         :DOI:`10.1016/0167-8655(92)90055-5`
+
+    Examples
+    --------
+    Dilation enlarges bright regions
+
+    >>> import cupy as cp
+    >>> import skimage as ski
```
gigony:
```suggestion
    >>> import cucim.skimage as ski
```

python/cucim/src/cucim/skimage/transform/_geometric.py
```diff
@@ -249,6 +261,32 @@ def __add__(self, other):
...
+
+    @classmethod
+    def from_estimate(cls, src, dst, *args, **kwargs):
+        """Estimate transform from matching source/destination coordinates."""
+        return _from_estimate(cls, src, dst, *args, **kwargs)
+
+
+def _from_estimate(cls, src, dst, *args, **kwargs):
+    tf = cls.identity(src.shape[1])
+    estimate_func = getattr(tf, "_estimate", None)
```
gigony, line 278: From AI review:

> from_estimate() now forces a CuPy allocation before it knows whether the caller is using NumPy or CuPy. GeometricTransform._from_estimate() constructs cls.identity(...), and the default identity/constructor path allocates with xp=cp. That regresses the NumPy backend. I reproduced this directly: ProjectiveTransform.from_estimate(np_src, np_dst) and EuclideanTransform.from_estimate(np_src, np_dst) both raised CUDARuntimeError here, while the deprecated cls(xp=np).estimate(...) path still worked.

python/cucim/src/cucim/skimage/_shared/utils.py
gigony: ## From AI Review

the deprecated morphology parameters no longer preserve their old exclusive-threshold semantics. The new decorator rewrites the deprecated argument to `DEPRECATED` and copies its value into the new name before the function body runs. But `remove_small_objects()` tries to distinguish old behavior with `if min_size is not DEPRECATED`, so that branch is unreachable for normal deprecated calls. The result is that deprecated `min_size=` and `area_threshold=` silently behave like the new inclusive `max_size=` on equality-boundary cases.

`379:425:python/cucim/src/cucim/skimage/_shared/utils.py`

```379:425:python/cucim/src/cucim/skimage/_shared/utils.py
def fixed_func(*args, **kwargs):
    deprecated_value = DEPRECATED
    new_value = DEPRECATED

    if len(args) > deprecated_idx:
        deprecated_value = args[deprecated_idx]
        if self.new_name is not None:
            args = (
                args[:deprecated_idx]
                + (DEPRECATED,)
                + args[deprecated_idx + 1 :]
            )
    if self.deprecated_name in kwargs.keys():
        deprecated_value = kwargs[self.deprecated_name]
        if self.new_name is not None:
            kwargs[self.deprecated_name] = DEPRECATED
    ...
    elif self.new_name is not None:
        kwargs[self.new_name] = deprecated_value
```

`174:179:python/cucim/src/cucim/skimage/morphology/misc.py`

```174:179:python/cucim/src/cucim/skimage/morphology/misc.py
if min_size is not DEPRECATED:
    # Exclusive threshold is deprecated behavior
    too_small = component_sizes < min_size
else:
    # New behavior uses inclusive threshold
    too_small = component_sizes <= max_size
```


`283:296:python/cucim/src/cucim/skimage/morphology/misc.py`

```283:296:python/cucim/src/cucim/skimage/morphology/misc.py
# removing small objects from the inverse of ar
with warnings.catch_warnings():
    warnings.filterwarnings(
        "ignore",
        message="Parameter `min_size` is deprecated",
        category=FutureWarning,
    )
    out = remove_small_objects(
        out,
        min_size=area_threshold,
        max_size=max_size,
        connectivity=connectivity,
        out=out,
    )
```

The tests in `python/cucim/src/cucim/skimage/morphology/tests/test_misc.py` miss this because they do not cover `min_size=` at all, and the new `area_threshold` test does not exercise a component/hole whose size is exactly equal to the threshold.

python/cucim/src/cucim/skimage/measure/_regionprops.py
```diff
@@ -126,6 +123,7 @@
     "inertia_tensor_eigvals": float,
     "intensity_max": float,
     "intensity_mean": float,
+    # 'intensity_median': float,  # TODO: uncomment once supported
     "intensity_min": float,
```
gigony, line 127: ## From AI review

`regionprops_table(..., properties=("intensity_median",))` is documented but still unusable. The property was added on `RegionProperties`, but `COL_DTYPES` still excludes it. That makes the non-batch path fall into `_extra_properties` lookup for a built-in property, while the batch path explicitly raises `NotImplementedError`. The tests hide this gap by removing `intensity_median` from the doc-vs-table consistency check instead of exercising the public API.

`99:127:python/cucim/src/cucim/skimage/measure/_regionprops.py`

```99:127:python/cucim/src/cucim/skimage/measure/_regionprops.py
COL_DTYPES = {
    "area": float,
    ...
    "intensity_max": float,
    "intensity_mean": float,
    # 'intensity_median': float,  # TODO: uncomment once supported
    "intensity_min": float,
    "intensity_std": float,
    ...
}
```

`923:935:python/cucim/src/cucim/skimage/measure/_regionprops.py`
```923:935:python/cucim/src/cucim/skimage/measure/_regionprops.py
for prop in properties:
    r = regions[0]
    ...
    prop = PROPS.get(prop, prop)
    rp = getattr(r, prop)
    if prop in COL_DTYPES:
        dtype = COL_DTYPES[prop]
    else:
        func = r._extra_properties[prop]
        dtype = _infer_regionprop_dtype(
            func,
            intensity=r._intensity_image is not None,
            ndim=r.image.ndim,
        )
```

`319:324:python/cucim/src/cucim/skimage/measure/_regionprops_gpu.py`
```319:324:python/cucim/src/cucim/skimage/measure/_regionprops_gpu.py
# TODO(grelee): implement batch kernel for efficient intensity median
# computation.
if "intensity_median" in properties:
    raise NotImplementedError(
        "Batch computation of 'intensity_median' is not yet supported"
    )
```

`1398:1410:python/cucim/src/cucim/skimage/measure/tests/test_regionprops.py`
```1398:1410:python/cucim/src/cucim/skimage/measure/tests/test_regionprops.py
def test_all_documented_items_in_col_dtypes():
    ...
    # TODO(grelee): remove next line once `intensity_median` is fully enabled
    property_names.remove("intensity_median")
```

python/cucim/src/cucim/skimage/morphology/misc.py
```diff
@@ -152,23 +171,42 @@ def remove_small_objects(ar, min_size=64, connectivity=1, *, out=None):
             "Did you mean to use a boolean array?"
         )
 
-    too_small = component_sizes < min_size
+    if min_size is not DEPRECATED:
+        # Exclusive threshold is deprecated behavior
+        too_small = component_sizes < min_size
+    else:
+        # New behavior uses inclusive threshold
+        too_small = component_sizes <= max_size
```
gigony: ## From AI Review

### Issue with the change

The PR author clearly intended to preserve the old exclusive threshold behavior (`<`) when the deprecated parameter is used, as seen in `remove_small_objects`:
```python
    if min_size is not DEPRECATED:
        # Exclusive threshold is deprecated behavior
        too_small = component_sizes < min_size
    else:
        # New behavior uses inclusive threshold
        too_small = component_sizes <= max_size
```

**The Bug:** 
Because both functions use `new_name="max_size"` in their `@deprecate_parameter` decorators, the decorator automatically translates the old parameter to the new one and hides the original usage (setting the old parameter to `DEPRECATED`). 

As a result, `min_size is not DEPRECATED` will **never** evaluate to `True`. The old behavior (`< min_size`) is completely unreachable dead code. Users who pass `min_size=3` will silently get the new behavior (`<= 3`), which changes the mathematical output of their code.

*(Note: This is also why the test `test_remove_small_holes_deprecated_area_threshold` in the PR incorrectly asserts that `area_threshold=3` produces the exact same result as `max_size=3`. It should produce the result of `max_size=2`.)*

---

### How to Fix the PR Before Approval/Merge

The `new_name` feature of the decorator is only safe when the new parameter is a 1:1 drop-in replacement with identical behavior. Since the mathematical behavior changes (`<` vs `<=`), the translation must be handled manually.

**Step 1: Remove `new_name="max_size"` from the decorators**
Keep the template warning, but remove the automatic translation so the function bodies can actually see which parameter was used.
```python
@deprecate_parameter(
    deprecated_name="min_size",
    # REMOVE: new_name="max_size",
    ...
)
```

**Step 2: Manually handle the logic in `remove_small_holes`**
Implement the user's suggestion to only pass the provided parameter:
```python
def remove_small_holes(ar, area_threshold=DEPRECATED, connectivity=1, *, max_size=64, out=None):
    # Manual validation
    if area_threshold is not DEPRECATED and max_size != 64:
        raise ValueError("Both deprecated parameter `area_threshold` and new parameter `max_size` are used. Use only the latter to avoid conflicting values.")
    
    kwargs = {"connectivity": connectivity, "out": out}
    if area_threshold is not DEPRECATED:
        kwargs["min_size"] = area_threshold
    else:
        kwargs["max_size"] = max_size
        
    # (The `warnings.catch_warnings()` block can be removed, as `remove_small_objects` 
    # will handle its own warnings appropriately based on the kwargs passed)
    out = remove_small_objects(out, **kwargs)
```

**Step 3: Manually handle the logic in `remove_small_objects`**
```python
def remove_small_objects(ar, min_size=DEPRECATED, connectivity=1, *, max_size=64, out=None):
    # Manual validation
    if min_size is not DEPRECATED and max_size != 64:
        raise ValueError("Both deprecated parameter `min_size` and new parameter `max_size` are used. Use only the latter to avoid conflicting values.")
        
    # ...
    if min_size is not DEPRECATED:
        too_small = component_sizes < min_size
    else:
        too_small = component_sizes <= max_size
```

**Step 4: Fix the tests**
Update `test_remove_small_holes_deprecated_area_threshold` (and add one for `min_size`) to ensure they verify the old behavior is preserved (e.g., `min_size=3` acts identically to `max_size=2`).

gigony (approved):
Thanks @grlee77 , 
I found minor issues and typos.
Otherwise, it looks good to me.

python/cucim/src/cucim/skimage/morphology/misc.py
grlee77: Thanks, let me take a closer look at this. I may have made a mistake porting the change from scikit-image

grlee77: This was a bug in the scikit-image branch I was working from. It should be fixed in b8df259

Looking at `Blame` in the UI in the scikit-image repo, I found it has also been fixed there:
https://github.com/scikit-image/scikit-image/commit/dc020276c192d1b2679e758992063a6ae78173aa

It is a slightly different implementation, but the net result should be the same

python/cucim/src/cucim/skimage/measure/_regionprops.py
grlee77: Right, it is implemented in the `RegionProps` when called on a per-ROI basis, but I didn't have time yet to do an efficient batch implementation (computing for all regions in a single kernel call like other intensity-based measures). Let me open a new issue to follow-up on that in the future

grlee77: created https://github.com/rapidsai/cucim/issues/1045

python/cucim/src/cucim/skimage/transform/_geometric.py
grlee77: good catch, let me update it

grlee77: Fixed in [daf3a51](https://github.com/rapidsai/cucim/pull/1038/commits/daf3a51c2b5e975c1acf98fb089b534a92b47e9e)

python/cucim/src/cucim/skimage/_shared/utils.py
grlee77: Fixed in [b8df259](https://github.com/rapidsai/cucim/pull/1038/commits/b8df2593912c4d1c9499244d54e7e122ea221b33)
(this was also incorrect in the commit of scikit-image I was working from but was since resolved there as well)

python/cucim/src/cucim/skimage/morphology/tests/test_binary.py
```diff
@@ -10,13 +10,26 @@
...
+#
+# pytestmark = pytest.mark.filterwarnings(
+#     "ignore:"
+#     "`binary_(dilation|erosion|opening|closing)` is deprecated.*"
+#     "Use `skimage.morphology.(dilation|erosion|opening|closing)` instead"
+#     ":FutureWarning"
+# )
+
+
 def test_non_square_image():
```
jakirkham, line 33: Should we raise an issue to track and add that to this comment?

python/cucim/src/cucim/skimage/morphology/tests/test_binary.py
```diff
@@ -378,3 +391,20 @@ def test_tuple_as_footprint(function, ndim):
...
+# )
+# def test_deprecation_warning(func_name):
+#     func = getattr(binary, func_name)
+#     footprint = footprint_rectangle((3, 3))
+
+#     regex = f"`{func_name}` is deprecated"
+#     with pytest.warns(FutureWarning, match=regex) as record:
+#         func(bw_img, footprint)
+#     assert_stacklevel(record)
+#     assert len(record) == 1
```
jakirkham, line 410: Should add to the comment the same issue we would have opened above?

python/cucim/src/cucim/skimage/morphology/tests/test_gray.py
```diff
@@ -544,16 +543,3 @@ def test_tuple_as_footprint(function, ndim, odd_only):
...
-@pytest.mark.parametrize("name", ["shift_x", "shift_y"])
-@pytest.mark.parametrize("value", [True, False, None])
-def test_deprecated_shift(func, name, value):
-    img = cp.ones(10)
-    func(img)  # Shouldn't warn
-
-    regex = "`shift_x` and `shift_y` are deprecated"
-    with pytest.warns(FutureWarning, match=regex) as record:
-        func(img, **{name: value})
-    assert_stacklevel(record)
```
jakirkham, line 559: Guessing the deprecation happened hence we can drop this test

python/cucim/src/cucim/skimage/morphology/gray.py
```diff
@@ -265,19 +188,6 @@ def erosion(
...
-    #     if len(footprint) != image.ndim:
-    #         raise ValueError(
-    #             "footprint.ndim={len(footprint)}, image.ndim={image.ndim}"
-    #         )
-    #     if image.ndim == 2 and any(s % 2 == 0 for s in footprint):
-    #         # only odd-shaped footprints are properly handled for tuples
-    #         footprint = cp.ones(footprint, dtype=bool)
-    #     else:
-    #         ndi.grey_erosion(image, size=footprint, output=out)
-    #         return out
```
jakirkham, line 280: This is dropped because it has been unused?

Why did we not need to use it previously?

python/cucim/src/cucim/skimage/transform/tests/test_geometric.py
```diff
@@ -157,45 +185,11 @@ def test_similarity_estimation():
...
+    tform3 = SimilarityTransform.from_estimate(SRC, DST)
     assert_array_almost_equal(tform3.params, tform2.params)
 
 
-def test_3d_similarity_estimation():
-    # Note: some choices of seed will produce poorly conditioned values,
-    #       leading to NaN in the estimate. We fix the seed here to avoid this.
-    seed = 10
-    rng = cp.random.default_rng(seed)
-    src_points = rng.random(size=(1000, 3))
```
jakirkham, line 171: Why was this test dropped?

python/cucim/src/cucim/skimage/transform/tests/test_geometric.py
```diff
@@ -893,35 +1068,42 @@ def test_euler_rotation():
...
 
 
-@pytest.mark.parametrize("xp", [np, cp])
-def test_similarity_transform_params(xp):
-    with pytest.raises(ValueError):
-        _ = SimilarityTransform(
-            translation=(4, 5, 6, 7), dimensionality=4, xp=xp
-        )
-    tf = SimilarityTransform(scale=4, dimensionality=3, xp=xp)
-    assert_array_equal(tf(xp.asarray([[1, 1, 1]])), xp.asarray([[4, 4, 4]]))
```
jakirkham, line 922: Why was this test dropped?

conda/environments/all_cuda-131_arch-x86_64.yaml
```diff
@@ -40,7 +40,7 @@ dependencies:
 - python>=3.11,<3.14
 - pywavelets>=1.6
 - recommonmark
-- scikit-image>=0.19.0,<0.26.0
+- scikit-image>=0.23.2,<0.27.0
```
jakirkham, line 43: Looks like this was missed in the Conda recipe. We are handling this in PR: https://github.com/rapidsai/cucim/pull/1049

xref: https://github.com/rapidsai/cucim/pull/1049#discussion_r2948102658

PR #1040, opened 2026-03-09
Add `cucim.skimage.restoration.rolling_ball` (useful for background subtraction)

grlee77:
The implementation here is equivalent to the scikit-image implementation. There was no need to try adapting the custom Cython code from scikit-image as the core operation amounts to greyscale morphology which has already been implemented.

We can support the `nansafe=True` option in a follow-up if we add `mask` support to vendored ndimage code

The new kernels added here are just for efficient creation of the `ball_kernel` or `ellipse_kernel` tensors used as inputs to the grayscale morphology function.

Large footprints degrade performance for any morphology operations. For binary morphology we avoid much degradation by decomposing into a sequence of 3x3 footprints instead, but such a decomposition is not available for the non-flat structuring elements used by this algorithm. The large footprint size will become the limiting factor in processing higher resolution images. This implementation adds a `downsample` option that is not present in `scikit-image` to deal with this scenario. It is simple conceptually for a user to implement this outside of `rolling_ball`, but in practice making sure the grayscale kernel `structure` remains consistent requires scaling both the magnitude and size of the structure accordingly. Given that this is prone to error and using even modest downsampling provides huge performance benefit I think it makes sense to include the option. The number of operations for a downsampling factor `f` will be reduced by a factor of `f ** (2*d)` in `d` dimensions (e.g. 16-fold fewer operations for a downsampling factor of 2 in 2D or 64-fold fewer in 3D).

# Benchmark results:  shape (1920, 1080), no downsampling

| footprint radius | duration: skimage (s) | duration: cuCIM (s) | acceleration | 
|------------------|-----------------------|---------------------|--------------|
| 5                |  0.905                |  0.00134            |    674x      |
| 10               |  1.93                 |  0.00528            |    366x      |
| 20               |  5.42                 |  0.02029            |    267x      |
| 40               |  17.62                |  0.08104            |    217x      |

### DIFF
--- a/python/cucim/src/cucim/skimage/restoration/tests/test_rolling_ball.py
+++ b/python/cucim/src/cucim/skimage/restoration/tests/test_rolling_ball.py
@@ -40,14 +40,17 @@ def _ball_kernel_reference(
         raise ValueError("dtype must be a floating-point type")
 
     half_size = math.ceil(radius)
-    coords = np.meshgrid(
-        *[
-            np.arange(-half_size, half_size + 1, dtype=dtype)
-            for _ in range(ndim)
-        ],
-        indexing="ij",
+    coords = np.stack(
+        np.meshgrid(
+            *[
+                np.arange(-half_size, half_size + 1, dtype=dtype)
+                for _ in range(ndim)
+            ],
+            indexing="ij",
+        ),
+        axis=-1,
     )
-    sum_of_squares = sum(c**2 for c in coords)
+    sum_of_squares = np.sum(coords * coords, axis=-1)
     distance_from_center = np.sqrt(sum_of_squares)
     kernel = np.sqrt(np.clip(radius**2 - sum_of_squares, 0, None))
     if structure_and_footprint:


python/cucim/src/cucim/skimage/restoration/_rolling_ball.py
```diff
@@ -0,0 +1,535 @@
...
+    The result matches the scikit-image implementation exactly.
+    """
+
+    image = cp.asarray(image)
+    float_type = cp.float32 if image.dtype.itemsize <= 4 else cp.float64
+    img = image.astype(float_type, copy=False)
+
+    if downscale is not None:
+        # Scale the radius proportionally (minimum 1)
+        new_radius = max(1, int(round(radius / downscale)))
```
gigony, line 329: AI-generated review note: because `downscale=None` gets normalized to `1` in `rolling_ball()`, this branch rounds non-integer radii even on the default path (`radius=2.5` becomes `new_radius=2`). Upstream keeps the float radius and builds support with `ceil(radius)`, so this changes the footprint and output for float radii. Could we preserve the no-downscale path as float-valued instead of quantizing it?

benchmarks/skimage/cucim_restoration_bench.py
```diff
@@ -171,6 +180,44 @@ def main(args):
...
+            elif ndim == 2:
+                if run_cpu:
+                    radius = [5, 10, 20, 40]
+                else:
+                    radius = [5, 10, 20, 40, 80, 160]
+            var_kwargs["radius"] = radius
+
+            # algorithm kwarg is not available in scikit-image
+            if not run_cpu:
+                var_kwargs["algorithm"] = ["exact", "layered", "tophat"]
```
gigony: AI-generated review note: `ImageBench` forwards `var_kwargs` directly into `cucim.skimage.restoration.rolling_ball()`, but this PR does not add an `algorithm` parameter to that function. That means the GPU benchmark path should fail with `TypeError` instead of producing results. Should this axis be dropped until `algorithm=` exists?

python/cucim/src/cucim/skimage/restoration/_rolling_ball.py
```diff
@@ -0,0 +1,535 @@
...
+    return background.astype(image.dtype, copy=False)
+
+
+def rolling_ball(
+    image,
+    *,
+    radius=100,
+    kernel=None,
+    nansafe=False,
+    workers=None,
```
gigony, line 390: AI-generated review note: upstream `skimage.restoration.rolling_ball()` uses `num_threads`, not `workers`. As written, code ported from scikit-image that passes `num_threads=...` will raise `TypeError` on cuCIM even though the PR description says the implementation is equivalent. Could we accept `num_threads` (even as an ignored alias) for compatibility?

python/cucim/src/cucim/skimage/restoration/tests/test_rolling_ball.py
```diff
@@ -0,0 +1,315 @@
...
+    )
+    cp.cuda.Stream.null.synchronize()
+    time_down_ms = 1000 * (time.perf_counter() - t0)
+
+    nrmse = normalized_root_mse(background_full, background_down)
+
+    # result with and without downscaling should be similar
+    assert nrmse < 0.5
+
+    assert time_down_ms < time_full_ms, (
```
gigony: AI-generated review note: this asserts a wall-clock performance property inside the unit suite. On shared or busy CI GPUs, `perf_counter()` noise can easily flip `time_down_ms < time_full_ms` even when the implementation is correct, so this looks flaky. I would keep the shape, dtype, and NRMSE checks here and leave the speed expectation to the benchmark suite.

python/cucim/src/cucim/skimage/restoration/_rolling_ball.py
gigony: AI-generated fix suggestion: keep `downscale=None` as `None` when calling `_rolling_ball_exact()` so the default path skips the quantization branch, and change `ball_kernel()` to size its support from `ceil(radius)` rather than `int(radius)`. That would preserve scikit-image's float-radius behavior while still letting the downscaled path rescale an integer radius when approximation is explicitly requested.

benchmarks/skimage/cucim_restoration_bench.py
gigony: AI-generated fix suggestion: the minimal-risk fix is to remove the `algorithm` benchmark axis and the matching docstring example in this PR. If you do want multiple algorithms, then `rolling_ball()` needs to grow a real `algorithm=` kwarg before the benchmark starts passing it through.

python/cucim/src/cucim/skimage/restoration/_rolling_ball.py
gigony: AI-generated fix suggestion: accept `num_threads=None` in the public signature for skimage compatibility and treat it as an ignored alias for the existing no-op threading parameter. If you want to keep `workers`, you could accept both and raise only when both are provided with conflicting non-`None` values.

python/cucim/src/cucim/skimage/restoration/tests/test_rolling_ball.py
gigony: AI-generated fix suggestion: keep the deterministic assertions here (`nrmse`, shape, and dtype), but drop `assert time_down_ms < time_full_ms` from the test. The speed claim is better covered by the benchmark harness that this PR already updates, where timing regressions can be observed without making CI pass/fail depend on wall-clock noise.

python/cucim/src/cucim/skimage/restoration/tests/test_rolling_ball.py
```diff
@@ -0,0 +1,315 @@
...
+    """Simple numpy-based reference implementation of the ball kernel.
+
+    used to validate ball_kernel elementwise kernel implementation
+    """
+    dtype = np.dtype(dtype)
+    if dtype.kind != "f":
+        raise ValueError("dtype must be a floating-point type")
+
+    coords = np.meshgrid(
+        *[np.arange(-radius, radius + 1, dtype=dtype) for _ in range(ndim)],
```
jakirkham: Noticing that the radius is `+1`on one side and not the other. Is this because we are doing some check like `-radius <= x < radius + 1`?

python/cucim/src/cucim/skimage/restoration/tests/test_rolling_ball.py
```diff
@@ -0,0 +1,315 @@
...
+        indexing="ij",
+    )
+    sum_of_squares = sum(c**2 for c in coords)
```
jakirkham: Wonder if we could use NumPy's `sum` over an axis

```suggestion
    sum_of_squares = (coords**2).sum(axis=0)
```

python/cucim/src/cucim/skimage/restoration/tests/test_rolling_ball.py
```diff
@@ -0,0 +1,315 @@
...
+    if structure_and_footprint:
+        center_height = radius
+        kernel -= center_height
+        footprint = np.zeros(kernel.shape, dtype=bool)
+        footprint[distance_from_center <= radius] = True
+        kernel[~footprint] = np.inf
+        return cp.asarray(kernel), cp.asarray(footprint)
+    else:
+        kernel[distance_from_center > radius] = np.inf
+        return cp.asarray(kernel), None
```
jakirkham, line 65: Would we want to use CuPy or the Array API throughout this function to avoid the transfer?

benchmarks/skimage/cucim_restoration_bench.py
```diff
@@ -171,6 +180,44 @@ def main(args):
...
+                if run_cpu:
+                    radius = [4, 8]
+                else:
+                    radius = [4, 8]
+                    if min(shape) > 16:
+                        radius += [16]
+                    if min(shape) > 32:
+                        radius += [32]
+                    if min(shape) > 64:
+                        radius += [64]
```
jakirkham, line 194: Perhaps we could build the list once and then slice it based on the constraint

Alternatively wonder if there is a comprehension that would do what we want

Though ok leaving if we think this is easier for readers to understand

python/cucim/src/cucim/skimage/restoration/tests/test_rolling_ball.py
```diff
@@ -0,0 +1,315 @@
...
+    radius = 60
+    for downscale in [None, downscale_factor]:
+        rolling_ball(img, radius=radius, downscale=downscale)
+
+    # Baseline: no downscaling
+    cp.cuda.Stream.null.synchronize()
+    t0 = time.perf_counter()
+    background_full = rolling_ball(img, radius=radius, downscale=None)
+    cp.cuda.Stream.null.synchronize()
+    time_full_ms = 1000 * (time.perf_counter() - t0)
```
jakirkham: Should we use a context manager for this?

Perhaps CuPy has something we can use

python/cucim/src/cucim/skimage/restoration/tests/test_rolling_ball.py
```diff
@@ -0,0 +1,315 @@
...
+        footprint = np.zeros(kernel.shape, dtype=bool)
+        footprint[distance_from_center <= radius] = True
+        kernel[~footprint] = np.inf
+        return cp.asarray(kernel), cp.asarray(footprint)
+    else:
+        kernel[distance_from_center > radius] = np.inf
+        return cp.asarray(kernel), None
+
+
+def _ellipsoid_kernel_reference(
```
jakirkham, line 68: Similar feedback from `_ball_kernel_reference` applies here

python/cucim/src/cucim/skimage/restoration/_rolling_ball.py
grlee77: Thanks, I will update `ball_kernel` to round up using `math.ceil(radius)` as suggested

benchmarks/skimage/cucim_restoration_bench.py
grlee77: This is a good catch. In an early revision I did have an `algorithm` option that had an approximation that could be a bit faster, but I dropped to keep things a bit simpler and then missed removing that option from the benchmark.

python/cucim/src/cucim/skimage/restoration/_rolling_ball.py
grlee77: I will add the `@deprecate_parameter` decorator so the old argument name can also be used until at least release 26.12.

python/cucim/src/cucim/skimage/restoration/tests/test_rolling_ball.py
grlee77: Removed as suggested. Given the large downsampling factor the timing difference should be pretty robust, but I agree that it isn't really needed in the test suite.

python/cucim/src/cucim/skimage/restoration/tests/test_rolling_ball.py
grlee77: This is just because `arange` is like Python's range and excludes the upper value. This gives a kernel that is always an odd shape of size (2 * radius + 1) on each axis with symmetric coords like [-3, -2, -1, 0, 1, 2, 3]

python/cucim/src/cucim/skimage/restoration/tests/test_rolling_ball.py
grlee77: For the test functions we don't need to care much about any overhead. Since the sizes are relatively smaller it may be faster to just do all operations on CPU and transfer at the end but I didn't measure it.

python/cucim/src/cucim/skimage/restoration/tests/test_rolling_ball.py
grlee77: Sure, `coords` is short (length `ndim`) so the current code isn't terrible, but I think this would be more efficient

python/cucim/src/cucim/skimage/restoration/tests/test_rolling_ball.py
grlee77: There is a `cupyx.performance.benchmark` helper that is more robust (has `num_warmup`, `max_duration`, etc. arguments), but in any case I had just removed the timing assertions from this test following [this suggestion](https://github.com/rapidsai/cucim/pull/1040#discussion_r3012730462) as I think we don't really need to test performance differences in the test suite.

python/cucim/src/cucim/skimage/restoration/tests/test_rolling_ball.py
grlee77: Oh, nevermind. The output of `np.meshgrid` is a `tuple of ndarray` so to make this change we would first have to use `np.stack` to stack the separate coordinate arrays.

grlee77: I went ahead and used stack and then this suggestion since that was the approach already being used in the other ellipsoid kernel reference implementation.

jakirkham:
Thanks all! 🙏

PR #1054, opened 2026-03-18
build wheels with CUDA 13.0.x, test wheels against mix of CTK versions

jameslamb:
Contributes to https://github.com/rapidsai/build-planning/issues/257

* builds CUDA 13 wheels with the 13.0 CTK

Contributes to https://github.com/rapidsai/build-planning/issues/256

* updates wheel tests to cover a range of CTK versions (we previously, accidentally, were only testing the latest 12.x and 13.x)


jakirkham:
Thanks James! 🙏

PR #1055, opened 2026-03-18
Update vendored interpolation kernels with precision fix

grlee77:
cuCIM's `cucim.skimage.transform` functions `rotate`, `rescale`, `resize`, `warp`, etc. rely on interpolation routines from `cupyx.scipy.ndimage`. I recently found that cases that should have run in float32 were actually using a float64 code path which results in substantial performance loss.

See the benchmarking results reported here:
https://github.com/cupy/cupy/pull/9769

This PR is just a port of those changes into cuCIM's vendored copy of ndimage

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -224,11 +224,13 @@ repos:
             ^python/cucim/src/cucim/skimage/restoration/__init__[.]py$|
             ^python/cucim/src/cucim/skimage/restoration/__init__[.]pyi$|
             ^python/cucim/src/cucim/skimage/restoration/_denoise[.]py$|
+            ^python/cucim/src/cucim/skimage/restoration/_rolling_ball[.]py$|
             ^python/cucim/src/cucim/skimage/restoration/deconvolution[.]py$|
             ^python/cucim/src/cucim/skimage/restoration/j_invariant[.]py$|
             ^python/cucim/src/cucim/skimage/restoration/tests/test_denoise[.]py$|
             ^python/cucim/src/cucim/skimage/restoration/tests/test_j_invariant[.]py$|
             ^python/cucim/src/cucim/skimage/restoration/tests/test_restoration[.]py$|
+            ^python/cucim/src/cucim/skimage/restoration/tests/test_rolling_ball[.]py$|
             ^python/cucim/src/cucim/skimage/restoration/uft[.]py$|
             ^python/cucim/src/cucim/skimage/segmentation/__init__[.]py$|
             ^python/cucim/src/cucim/skimage/segmentation/__init__[.]pyi$|
@@ -503,11 +505,13 @@ repos:
             python/cucim/src/cucim/skimage/restoration/__init__[.]py$|
             python/cucim/src/cucim/skimage/restoration/__init__[.]pyi$|
             python/cucim/src/cucim/skimage/restoration/_denoise[.]py$|
+            python/cucim/src/cucim/skimage/restoration/_rolling_ball[.]py$|
             python/cucim/src/cucim/skimage/restoration/deconvolution[.]py$|
             python/cucim/src/cucim/skimage/restoration/j_invariant[.]py$|
             python/cucim/src/cucim/skimage/restoration/tests/test_denoise[.]py$|
             python/cucim/src/cucim/skimage/restoration/tests/test_j_invariant[.]py$|
             python/cucim/src/cucim/skimage/restoration/tests/test_restoration[.]py$|
+            python/cucim/src/cucim/skimage/restoration/tests/test_rolling_ball[.]py$|
             python/cucim/src/cucim/skimage/restoration/uft[.]py$|
             python/cucim/src/cucim/skimage/segmentation/__init__[.]py$|
             python/cucim/src/cucim/skimage/segmentation/__init__[.]pyi$|
--- a/benchmarks/skimage/cucim_restoration_bench.py
+++ b/benchmarks/skimage/cucim_restoration_bench.py
@@ -121,6 +121,15 @@ def main(args):
         ("wiener", dict(balance=100.0), dict(), False, False),
         ("unsupervised_wiener", dict(), dict(), False, False),
         ("richardson_lucy", dict(), dict(num_iter=[5]), False, True),
+        (
+            "rolling_ball",
+            dict(nansafe=False),
+            dict(
+                # radius=[5, 10, 20, 40],  # appropriate radius range set below
+            ),
+            False,  # could support color with some modification to the benchmark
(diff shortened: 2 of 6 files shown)


python/cucim/src/cucim/skimage/_vendored/_ndimage_interp_kernels.py
```diff
@@ -263,6 +265,7 @@ def _generate_interp_custom(
     name="",
     integer_output=False,
     nprepad=0,
+    float_dtype=cupy.float64,
```
jakirkham, line 288: Should we also make this `None` and then fill it in later if unspecified?

Also have the same question for other functions below

python/cucim/src/cucim/skimage/_vendored/_ndimage_interp_kernels.py
```diff
@@ -20,7 +22,7 @@
 spline_weights_inline = _spline_kernel_weights.spline_weights_inline
 
 
-def _get_coord_map(ndim, nprepad=0):
+def _get_coord_map(ndim, nprepad=0, float_type=None):
```
jakirkham, line 25: In cases where `float_type` is unused, maybe we could add something to the docstring to call this out

python/cucim/src/cucim/skimage/_vendored/_ndimage_interpolation.py
```diff
@@ -477,19 +483,22 @@ def affine_transform(
     if output_shape is None:
         output_shape = input.shape
 
+    float_dtype = cupy.float64
+    if not float64_coords:
+        float_dtype = cupy.promote_types(input.real.dtype, cupy.float32)
     if mode == "opencv" or mode == "_opencv_edge":
         if matrix.ndim == 1:
             matrix = cupy.diag(matrix)
         coordinates = cupy.indices(output_shape, dtype=cupy.float64)
```
jakirkham: Should we use `float_dtype` here?

```suggestion
        coordinates = cupy.indices(output_shape, dtype=float_dtype)
```

python/cucim/src/cucim/skimage/_vendored/_ndimage_interpolation.py
```diff
@@ -362,6 +365,7 @@ def affine_transform(
     prefilter=True,
     *,
     texture_memory=False,
+    float64_coords=False,
```
gigony, line 368: AI-generated review note: If the intent here is to stay aligned with the CuPy #9769 port, defaulting `float64_coords` to `False` broadens the behavior change from the explicit `rotate()` fast path to every existing `affine_transform()` caller. Public call sites like `transform._warps.warp()` will now silently lose the legacy float64-coordinate path for float32 inputs unless they discover and override this new kwarg.

Suggested fix: keep the default on the legacy path and opt specific call sites into the relaxed-precision path only after validating them. For example:
```python
def affine_transform(..., float64_coords=True):
    ...

return affine_transform(
    input, matrix, offset, output_shape, output, order, mode, cval,
    prefilter, float64_coords=False,
)
```

python/cucim/src/cucim/skimage/_vendored/_ndimage_interpolation.py
```diff
@@ -743,6 +758,7 @@ def shift(
             order=order,
             integer_output=integer_output,
             nprepad=nprepad,
+            float_dtype=float_dtype,
         )
         shift = cupy.asarray(shift, dtype=cupy.float64, order="C")
```
gigony: AI-generated review note: This path still materializes `shift` as `float64`, and the same pattern appears again in `zoom()` below. Because the generated kernels keep coordinate and weight temporaries typed as `W` (`raw W shift` / `raw W zoom`, `W c_j`, `W weights_j`), both entry points still execute with double-precision coordinates here, so the precision fix and expected speedup are incomplete.

Suggested fix: cast these control arrays to `float_dtype` before launching the kernel, or change the generated kernel so the coordinate/weight temporaries use `float_type` instead of `W`. The minimal call-site fix would be:
```python
shift = cupy.asarray(shift, dtype=float_dtype, order="C")
...
zoom = cupy.asarray(zoom, dtype=float_dtype)
```

python/cucim/src/cucim/skimage/_vendored/_ndimage_interpolation.py
grlee77: I think I was trying to improve precision of the coordinate computation, but I think it is probably not necessary to use double here. It was merged this way upstream to CuPy, though. Let me make the change here and we can see about possibly porting the same change upstream as well.

python/cucim/src/cucim/skimage/_vendored/_ndimage_interpolation.py
grlee77: I think it was a mistake in the CuPy PR to set this to True by default. I should probably submit a small patch there to update CuPy to match the implementation here.

python/cucim/src/cucim/skimage/_vendored/_ndimage_interpolation.py
grlee77: Good point, I need to look more closely to see if it is safe to change from W to float_dtype. It might not be that simple if `W` is allowed to be complex-valued.

grlee77: I updated the `zoom` and `shift` allocations to use the correct floating-point type and left `W` in the helper functions

gigony: Hi @grlee77 , I see the following error on https://github.com/rapidsai/cucim/actions/runs/23821142337/job/69434943992?pr=1055
```
FAILED python/cucim/src/cucim/skimage/transform/tests/test_warps.py::test_downsize[float32] - NameError: name 'float_type' is not defined
```

grlee77: This was a typo. Fixed by changing to `float_dtype`

python/cucim/src/cucim/skimage/_vendored/_ndimage_interp_kernels.py
grlee77: added docstring notes

python/cucim/src/cucim/skimage/_vendored/_ndimage_interpolation.py
jakirkham: Sounds good. Thanks Greg! 🙏

Please link the upstream PR here when you have it

jakirkham:
Thanks all! 🙏

PR #1056, opened 2026-03-18
fix(conda): specify `python-gil` for conda recipe

gforsyth:
Builds are failing (e.g. https://github.com/rapidsai/cucim/actions/runs/23230625912/job/67601928465) on the `main` branch because `conda` is trying to use a free-threaded Python variant.  Adding an explicit `python-gil` to the conda recipe to stop this.

Example solver error:

```
WARNING: failed to get package records, retrying.  exception was: Unsatisfiable dependencies for platform linux-aarch64: {MatchSpec("python==3.14.0rc1=h8f01225_2_cp314t"), MatchSpec("_python_rc")}
Encountered problems while solving:
  - nothing provides _python_rc needed by python-3.14.0rc1-h8f01225_2_cp314t

Could not solve for environment specs
The following packages are incompatible
├─ cuda-version =12.9 * is requested and can be installed;
├─ cupy >=13.6.0,!=14.0.0 * is installable with the potential options
│  ├─ cupy 13.6.0 would require
│  │  └─ cupy-core ==13.6.0 py39he4d69f8_0, which requires
│  │     └─ python >=3.9,<3.10.0a0 *_cpython, which can be installed;
│  ├─ cupy [13.6.0|14.0.1] would require
│  │  └─ cupy-core [==13.6.0 py310h967c7ba_0|==13.6.0 py310h967c7ba_1|==13.6.0 py310h967c7ba_2|==14.0.1 py310h967c7ba_0], which requires
│  │     └─ python >=3.10,<3.11.0a0 *, which can be installed;
│  ├─ cupy [13.6.0|14.0.1] would require
│  │  └─ cuda-version >=13,<14.0a0 *, which conflicts with any installable versions previously reported;
│  ├─ cupy 13.6.0 would require
│  │  └─ cuda-version >=11,<12.0a0 *, which conflicts with any installable versions previously reported;
│  ├─ cupy [13.6.0|14.0.1] would require
│  │  └─ cupy-core [==13.6.0 py311h6a7bbfe_0|==13.6.0 py311h6a7bbfe_1|==13.6.0 py311h6a7bbfe_2|==14.0.1 py311h6a7bbfe_0], which requires
│  │     └─ python >=3.11,<3.12.0a0 *, which can be installed;
│  ├─ cupy [13.6.0|14.0.1] would require
[...]

PR #1058, opened 2026-03-19
fix(conda): specify `python-gil` for conda recipe (#1056)

gforsyth:
Backport of #1056 to ensure that conda/mamba don't try to grab free-threaded Python builds when solving.


jakirkham:
Thanks Gil and James! 🙏

PR #1077, opened 2026-04-27
Fixes for heap overflow

cdinea:

### DIFF
--- a/cpp/include/cucim/util/checked_math.h
+++ b/cpp/include/cucim/util/checked_math.h
@@ -7,14 +7,49 @@
 
 #include <cstddef>
 #include <cstdint>
+#include <cstdlib>
 #include <stdexcept>
+#include <string>
 #include <type_traits>
 
 namespace cucim::util
 {
 
-constexpr size_t kMaxTileBytes = 256ULL * 1024 * 1024; // 256 MiB
-constexpr size_t kMaxRasterBytes = 1ULL * 1024 * 1024 * 1024; // 1 GiB
+constexpr size_t kDefaultMaxTileBytes = 2ULL * 1024 * 1024 * 1024; // 2 GiB
+constexpr size_t kDefaultMaxRasterBytes = 16ULL * 1024 * 1024 * 1024; // 16 GiB
+
+namespace detail
+{
+
+inline size_t parse_env_size(const char* env_var, size_t default_value)
+{
+    const char* val = std::getenv(env_var);
+    if (val == nullptr || val[0] == '\0')
+    {
+        return default_value;
+    }
+    char* end = nullptr;
+    unsigned long long parsed = std::strtoull(val, &end, 10);
+    if (end == val || parsed == 0)
+    {
+        return default_value;
+    }
+    return static_cast<size_t>(parsed);
+}
+
+} // namespace detail
+
+inline size_t max_tile_bytes()
+{
+    static const size_t value = detail::parse_env_size("CUCIM_MAX_TILE_BYTES", kDefaultMaxTileBytes);
+    return value;
+}
+
+inline size_t max_raster_bytes()
+{
+    static const size_t value = detail::parse_env_size("CUCIM_MAX_RASTER_BYTES", kDefaultMaxRasterBytes);
+    return value;
+}
 
 template <typename T>
 inline T checked_mul(T a, T b)
@@ -36,11 +71,13 @@ inline size_t checked_mul3(size_t a, size_t b, size_t c)
 inline size_t checked_tile_size(size_t width, size_t height, size_t pixel_bytes)
 {
     size_t result = checked_mul3(width, height, pixel_bytes);
-    if (result > kMaxTileBytes)
+    size_t limit = max_tile_bytes();
+    if (result > limit)
     {
         throw std::overflow_error("Tile size exceeds maximum allowed (" +
                                   std::to_string(result) + " > " +
-                                  std::to_string(kMaxTileBytes) + " bytes)");
+                                  std::to_string(limit) + " bytes). "
+                                  "Override with CUCIM_MAX_TILE_BYTES env var.");
     }
     return result;
 }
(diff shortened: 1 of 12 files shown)


cpp/include/cucim/util/checked_math.h
```diff
@@ -0,0 +1,62 @@
...
+#include <cstddef>
+#include <cstdint>
+#include <stdexcept>
+#include <type_traits>
+
+namespace cucim::util
+{
+
+constexpr size_t kMaxTileBytes = 256ULL * 1024 * 1024; // 256 MiB
+constexpr size_t kMaxRasterBytes = 1ULL * 1024 * 1024 * 1024; // 1 GiB
```
grlee77: @gigony , @cdinea, @jakirkham  : I don't think we should just hard-code these limits here. 

This `kMaxRasterBytes` is much smaller than many real-world microscopy images (and the memory limits of typical workstations), so it would likely prevent valid use cases?

I am not sure if the 256 MB tile size limit is also problematic in practice, but may not want to also hard-code an arbitrary value.

cpp/plugins/cucim.kit.cuslide/src/cuslide/tiff/ifd.cpp
```diff
@@ -147,7 +148,8 @@ bool IFD::read(const TIFF* tiff,
     int32_t n_ch = samples_per_pixel_; // number of channels
     int ndim = 3;
 
-    size_t raster_size = w * h * samples_per_pixel_;
+    size_t raster_size = cucim::util::checked_raster_size(
```
grlee77: The width and height here come from user-supplied values not potentially malicious file metadata so I don't think we need to check and reject here?

cpp/include/cucim/util/checked_math.h
grlee77: One option could be to have a bit larger default like 16 GB for `kMaxRasterBytes` but allow user-override via environment variable like `CUCIM_MAX_RASTER_BYTES`?

grlee77 (commented):
I commented on an issue with hard-coded size limits that may be too restrictive in practice.

I also had a local agent review and it pointed out one remaining related issue:

> High: J2K validation still does not validate component dimensions used by the SYCC converters.
>     cpp/plugins/cucim.kit.cuslide/src/cuslide/jpeg2k/libopenjpeg.cpp:231 validates output bytes from component 0 only. But cpp/plugins/
>     cucim.kit.cuslide/src/cuslide/jpeg2k/color_conversion.cpp:36 and cpp/plugins/cucim.kit.cuslide/src/cuslide/jpeg2k/color_conversion.cpp:122
>     read Cb/Cr based on component 0 geometry and subsampling assumptions. A malformed codestream with non-null but undersized chroma components
>     can still drive out-of-bounds reads. Add per-component dimension checks for RGB, 4:4:4, 4:2:2, and 4:2:0 before conversion.

cpp/include/cucim/util/checked_math.h
cdinea: thank you @grlee77 - i pushed the changes in [this](https://github.com/rapidsai/cucim/pull/1077/commits/3a7ba9672132502f7e24ae1acd7158d2390d5754l) commit to raise defaults to 2 GiB (tiles) and 16 GiB (rasters) - I think these should accommodate even large microscopy tiles and full-resolution region reads on typical workstations?
Also added environment variable overrides: CUCIM_MAX_TILE_BYTES and CUCIM_MAX_RASTER_BYTES so users with unusual image sizes or memory-constrained environments can adjust without recompiling

cpp/plugins/cucim.kit.cuslide/src/cuslide/tiff/ifd.cpp
cdinea: thank you for the feedback @grlee77 - updated this in [this](https://github.com/rapidsai/cucim/pull/1077/commits/6de1ddaee67c0358563f0a1dcd4dd8b20cd8afcb) commit to use checked_mul3 (overflow detection only) instead of checked_raster_size (overflow + upper-bound limit),this way we still catch silent integer wraparound

cdinea:
> I commented on an issue with hard-coded size limits that may be too restrictive in practice.
> 
> I also had a local agent review and it pointed out one remaining related issue:
> 
> > High: J2K validation still does not validate component dimensions used by the SYCC converters.
> > cpp/plugins/cucim.kit.cuslide/src/cuslide/jpeg2k/libopenjpeg.cpp:231 validates output bytes from component 0 only. But cpp/plugins/
> > cucim.kit.cuslide/src/cuslide/jpeg2k/color_conversion.cpp:36 and cpp/plugins/cucim.kit.cuslide/src/cuslide/jpeg2k/color_conversion.cpp:122
> > read Cb/Cr based on component 0 geometry and subsampling assumptions. A malformed codestream with non-null but undersized chroma components
> > can still drive out-of-bounds reads. Add per-component dimension checks for RGB, 4:4:4, 4:2:2, and 4:2:0 before conversion.

thank you for the feedback @grlee77 -  addressed the agent review in the [latest](https://github.com/rapidsai/cucim/pull/1077/commits/ccd27e8e81b29fb1539f30333ee9ec8443e2d630) commit . Added per-component dimension validation in libopenjpeg.cpp that computes the minimum expected chroma size as ceil(luma_dim / subsampling_factor) using each component's dx/dy values, and rejects the image if any chroma plane is undersized

cdinea:
/ok to test https://github.com/rapidsai/cucim/pull/1077/commits/ccd27e8

grlee77 (approved):
Thanks @cdinea, looks good to me (once CI passes)

gigony (approved):
Thanks @cdinea for handling the issues! It looks good to me! Please remove the 'draft' tag from the PR title now.

cdinea:
thank you for the (re)review and approval @grlee77 @gigony

jakirkham (approved):
Thanks Cristiana for working on this! Also thanks Gigon and Greg for reviewing 🙏

PR #1079, opened 2026-05-11
vendor metadata hardening

cdinea:
Harden vendor-specific TIFF metadata parsing in cuslide and cuslide2
- Bounds-check pixel_spacings vector before access (empty vector guard + spacing_index bound)
- Replace hardcoded +34 pointer arithmetic with bounds-checked scan for XML associated images
- Guard against division by zero when downsample factor rounds to 0
- Wrap std::stod/stoul in try-catch for untrusted Philips XML numeric data
- Add string length check before std::mismatch for Aperio/Philips/Macro/Label prefix detection
- Cap recursive XML parsing at depth 32 to prevent stack exhaustion
- Limit Aperio SVS ImageDescription processing to 1 MB to prevent memory exhaustion


gigony (approved):
Thanks @cdinea for addressing the issues!
It looks good to me!

PR #1081, opened 2026-05-12
fix(ci): add explicit `actions: write` permission for `telemetry-summarize`


gforsyth:
## Description
As part of https://github.com/rapidsai/build-planning/issues/275 we removed the default set of permissions from all workflow jobs.

The shared-action that handles telemetry cleanup is now failing because it is trying to clean up artifacts that shouldn't remain after a given workflow run.

This PR adds an explicit `actions: write` permission, to allow that action to clean up those artifacts.

xref https://github.com/rapidsai/shared-actions/issues/106

### DIFF
--- a/.github/workflows/pr.yaml
+++ b/.github/workflows/pr.yaml
@@ -6,9 +6,7 @@ on:
 concurrency:
   group: ${{ github.workflow }}-${{ github.ref }}
   cancel-in-progress: true
-
 permissions: {}
-
 jobs:
   pr-builder:
     needs:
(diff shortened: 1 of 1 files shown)


.github/workflows/pr.yaml
```diff
@@ -6,9 +6,7 @@ on:
 concurrency:
   group: ${{ github.workflow }}-${{ github.ref }}
   cancel-in-progress: true
-
 permissions: {}
-
 jobs:
   pr-builder:
     needs:
```
jakirkham, line 12: Could we keep the spacing here?

jakirkham (commented):
Thanks Gil! 🙏

One style question below

Otherwise looks good

.github/workflows/pr.yaml
gforsyth: `yq` is always going to remove the extraneous spaces, and I'm inclined to let it, since there will be more programmatic updates in the future as we tighten permission blocks

jakirkham: Searched a bit and didn't come up with anything either

Just a bit unfortunate that this will turn into one giant block of YAML

jakirkham (approved):

jameslamb (approved):

PR #1085, opened 2026-05-19
Upgrade nvImageCodec dependency from v0.7.0 to v0.8.0

cdinea:
## Summary

- Bumps all nvImageCodec version constraints from `>=0.7.0,<0.8.0` to `>=0.8.0,<0.9.0` across `dependencies.yaml`, conda environments, and `conda_build_config.yaml`
- Adapts the cuslide2 dynlink layer to the **v0.8.0 breaking API change**: `nvimgcodecCodeStreamCreateFromFile()` and `nvimgcodecCodeStreamCreateFromHostMem()` now require an additional nullable `code_stream_view` parameter
- Adds new v0.8.0 API types to the dynlink header:
  - `nvimgcodecCodeStreamInfoTiffExt_t` (TIFF pagination via `next_bitstream_offset`, SubIFD offsets)
  - `bitstream_offset` and `limit_images` fields in `nvimgcodecCodeStreamView_t`
  - `NVIMGCODEC_METADATA_KIND_TIFF_TAG_LIST` enum value
  - `NVIMGCODEC_STRUCTURE_TYPE_TIFF_CODE_STREAM_INFO` enum value
- Updates `nvimgcodec_version.h` from 0.7.0 to 0.8.0
- Cleans up stale v0.6.0/v0.7.0 workaround comments now that the extension auto-discovery bug is fixed in v0.8.0

### Key v0.8.0 features unlocked by this upgrade

| Feature | Benefit |
|---------|---------|
| TIFF CodeStream offset parsing (`bitstream_offset`) | Parse from arbitrary IFD offset — enables future NDPI/SCN support |
| TIFF pagination (`limit_images`, `next_bitstream_offset`) | Efficiently page through large multi-image TIFFs |
| SubIFD support (`subifd_offsets`) | Access sub-resolution layers embedded in SubIFDs |
| `TIFF_TAG_LIST` metadata kind | Enumerate available TIFF tags without trial-and-error queries |
| Hamamatsu NDPI > 4 GB support | Fixed upstream parser for large NDPI files |
| Extension auto-discovery fix | Conda installs now find codec plugins without workaround |
| CUDA stream pool (round-robin) | Reduced GPU stream synchronization overhead |

note: test scripts for additional file formats like ndpi,scn are WIP and will be submitted shortly.

## Test plan

- [ ] CI build passes (conda and pip) with nvImageCodec 0.8.0
- [ ] Existing cuslide2 unit tests pass (single-tile, batch, caching)
- [ ] Verify Aperio SVS, Philips TIFF, and Generic TIFF open and decode correctly
- [ ] Verify `nullptr` code_stream_view preserves existing behavior (no regression)
- [ ] Spot-check that the extension path workaround still works as a safety net on non-standard layouts


cdinea:
@jolorunyomi is it possible to  have this as part of 26.06 release? we are targeting this as part of the 26.06 release

jakirkham:
Not Joseph, but yes cuCIM is still in burndown. So we can do this as long as it is done before code freeze next week (so by Wednesday May 27th)

It looks like this has conflicts. Could we please resolve those and then enable CI to run?

cpp/plugins/cucim.kit.cuslide2/src/cuslide/tiff/types.h
```diff
@@ -20,6 +20,12 @@ enum class TiffType : uint32_t
     Generic = 0,
     Philips = 1,
     Aperio = 2,
+    Hamamatsu = 3,
+    Leica = 4,
+    Ventana = 5,
+    Trestle = 6,
+    OmeTiff = 7,
+    Qptiff = 8,
```
gigony: Minor issue, but it might be better to use `QpTiff` instead of `Qptiff`, since we already use `OmeTiff`. 
This would make the naming more readable and meaningful, while keeping it consistent with `OmeTiff`.
That said, both options are consistent from certain perspectives, so I am curious how you feel about it. ;)

gigony (approved):
Thanks @cdinea for the changes!
It looks great to me!

cpp/plugins/cucim.kit.cuslide2/src/cuslide/tiff/types.h
cdinea: thank you for the prompt review @gigony  - i think you are right -renamed Qptiff → QpTiff across all three files to match the OmeTiff convention. Pushed in this commit https://github.com/rapidsai/cucim/pull/1085/commits/9d0eb4e

cdinea:
/ok to test https://github.com/rapidsai/cucim/pull/1085/commits/9d0eb4e

cpp/plugins/cucim.kit.cuslide2/src/cuslide/tiff/tiff.cpp
```diff
@@ -542,15 +542,15 @@ void TIFF::resolve_vendor_format()
         }
 
         // Method 2: Check metadata_blobs for Aperio (kind=1)
```
grlee77: Please consider the attached path so this case and the other cases below can rely on NVIMGCODEC_METADATA_KIND_MED_APERIO values, etc. when available rather than keeping these numeric value mappings up to date (e.g. the comment kind=1 above is outdated here)

[metadata_kind.patch](https://github.com/user-attachments/files/28285076/metadata_kind.patch)

grlee77 (approved):
Thanks, @cdinea, I attached a patch with a suggestion for `cpp/plugins/cucim.kit.cuslide2/src/cuslide/tiff/tiff.cpp` to use something lke

```cpp
+#ifdef CUCIM_HAS_NVIMGCODEC
+constexpr int kMetadataKindMedAperio = NVIMGCODEC_METADATA_KIND_MED_APERIO;
+constexpr int kMetadataKindMedPhilips = NVIMGCODEC_METADATA_KIND_MED_PHILIPS;
+constexpr int kMetadataKindMedVentana = NVIMGCODEC_METADATA_KIND_MED_VENTANA;
+constexpr int kMetadataKindMedLeica = NVIMGCODEC_METADATA_KIND_MED_LEICA;
+constexpr int kMetadataKindMedTrestle = NVIMGCODEC_METADATA_KIND_MED_TRESTLE;
+#else
+constexpr int kMetadataKindMedAperio = 5;
+constexpr int kMetadataKindMedPhilips = 6;
+constexpr int kMetadataKindMedVentana = 7;
+constexpr int kMetadataKindMedLeica = 8;
+constexpr int kMetadataKindMedTrestle = 9;
+#endif
```

so descriptive `kMetadataKindMedAperio`, etc. constants can be used instead of the numeric values in the code below. This seems less error prone when making future changes

cpp/plugins/cucim.kit.cuslide2/src/cuslide/nvimgcodec/nvimgcodec_tiff_parser.cpp
```diff
@@ -274,7 +350,7 @@ NvImageCodecTiffParserManager::NvImageCodecTiffParserManager()
         exec_params.struct_next = nullptr;
         exec_params.device_allocator = nullptr;
         exec_params.pinned_allocator = nullptr;
-        exec_params.max_num_cpu_threads = 0;
+        exec_params.max_num_cpu_threads = compute_max_decoder_threads();
```
jantonguirao, line 354: **[Suggestion]** Worth being explicit in release notes that this changes the effective default thread count for `cuslide2`. nvImageCodec's default is `num_cores`; the new heuristic gives `min(hw_threads/4, 8)`. On a 16-core workstation that's 16 → 4 threads (~4× reduction); on a 64-core node it's 64 → 8. Users running single-process workflows (notebooks, scripts) will see a different baseline than before.

The author already evaluated this and added `CUCIM_MAX_DECODER_THREADS=0` as an escape hatch back to the old behavior, so this isn't a blocker — just please surface the new default loudly in the release notes / changelog so single-process users aren't surprised.

cpp/plugins/cucim.kit.cuslide2/src/cuslide/nvimgcodec/nvimgcodec_tiff_parser.cpp
```diff
@@ -143,9 +146,82 @@ static std::string detect_nvimgcodec_extensions_path()
...
+//
+// Heuristic:  min(hardware_concurrency / 4, kMaxDefaultThreads)
+//
+// The CUDA stream count is left at nvImageCodec's default because
+// empirically it has limited performance impact and over-allocating
+// streams can itself cause oversubscription.
+//
+// Overridable via CUCIM_MAX_DECODER_THREADS environment variable.
+//   - value > 0 : use that exact thread count
+//   - value == 0: fall back to nvImageCodec default (num_cores - 1)
```
jantonguirao: **[Minor]** Doc inaccuracy: per `nvimgcodec.h` (`max_num_cpu_threads`: "0 means default value equal to number of cpu cores"), the nvImageCodec default is `num_cores`, not `num_cores - 1`. Suggest: `// - value == 0: fall back to nvImageCodec default (= num cpu cores)`.

cpp/plugins/cucim.kit.cuslide2/src/cuslide/nvimgcodec/nvimgcodec_tiff_parser.cpp
```diff
@@ -143,9 +146,82 @@ static std::string detect_nvimgcodec_extensions_path()
...
+//   - malformed : warn and fall through to heuristic
+//
+static constexpr int kMaxDefaultThreads = 8;
```
jantonguirao: **[Nit]** Name is slightly misleading — `kMaxDefaultThreads` reads as "the default value of the max", but the constant is used as a hard cap on the heuristic's fair-share calculation, not as a default. Something like `kThreadCountCap` or `kHeuristicMaxThreads` would describe the role more precisely.

cpp/plugins/cucim.kit.cuslide2/src/cuslide/nvimgcodec/nvimgcodec_tiff_parser.cpp
```diff
@@ -143,9 +146,82 @@ static std::string detect_nvimgcodec_extensions_path()
...
+        if (end == env_val || *end != '\0')
+        {
+            fmt::print(stderr,
+                "[cuslide2] WARNING: CUCIM_MAX_DECODER_THREADS='{}' is not a "
+                "valid integer — ignoring and using heuristic\n", env_val);
+        }
+        else if (val > 0)
+        {
+            fmt::print(stderr,
+                "[cuslide2] CUCIM_MAX_DECODER_THREADS={}\n", val);
```
jantonguirao: **[Minor]** This prints to stderr unconditionally whenever `CUCIM_MAX_DECODER_THREADS` is set to a valid positive value. That's a permanent banner on every `cuslide2` plugin load for anyone using the override. The warnings on malformed/negative are useful, but the success path probably wants to be `#ifdef DEBUG` only (matching the heuristic-debug print on line 213), or routed through a logger instead of always-on stderr.

cpp/plugins/cucim.kit.cuslide2/src/cuslide/tiff/tiff.cpp
```diff
@@ -616,6 +614,68 @@ void TIFF::resolve_vendor_format()
...
+
+        // Hamamatsu NDPI: extension .ndpi or MODEL contains "Hamamatsu"
+        if (file_ext == ".ndpi" || model.find("Hamamatsu") != std::string::npos)
+        {
+            tiff_type_ = TiffType::Hamamatsu;
+        }
+        // Ventana BIF: extension .bif or nvImageCodec MED_VENTANA blob (kind=7)
+        else if (file_ext == ".bif" ||
+                 (nvimgcodec_parser_ &&
+                  nvimgcodec_parser_->get_metadata_blobs(0).find(7) != nvimgcodec_parser_->get_metadata_blobs(0).end()))
```
jantonguirao: **[Nit]** `nvimgcodec_parser_->get_metadata_blobs(0)` is called twice per branch on lines 640/647/653. If `get_metadata_blobs` returns a `const std::map&` this is cheap, but readability suffers from the long expressions. A single local at the top of the `if (tiff_type_ == TiffType::Generic)` block (e.g. `const auto* blobs = nvimgcodec_parser_ ? &nvimgcodec_parser_->get_metadata_blobs(0) : nullptr;`) would make each branch a one-liner.

cpp/plugins/cucim.kit.cuslide2/src/cuslide/tiff/tiff.cpp
cdinea: thanks @grlee77  — applied your awesome suggestion in this commit https://github.com/rapidsai/cucim/pull/1085/commits/67fd09b3256a3dbcbc879aea0bfe1eaa026018b6. Added kMetadataKindMed{Aperio,Philips,Ventana,Leica,Trestle} constants that resolve to NVIMGCODEC_METADATA_KIND_MED_* enums when CUCIM_HAS_NVIMGCODEC is defined, with hardcoded fallbacks otherwise.

cpp/plugins/cucim.kit.cuslide2/src/cuslide/nvimgcodec/nvimgcodec_tiff_parser.cpp
cdinea: thank you @jantonguirao - good point, I missed this - will add a changelog entry calling this out explicitly. 

Something like:

Changed: cuslide2 now applies a decoder thread heuristic (min(hw_threads/4, 8)) instead of using nvImageCodec's default of num_cores - 1. This reduces the per-instance CPU threadpool to prevent oversubscription in multi-worker data loader deployments. Single-process workflows (notebooks, scripts) may see lower decode throughput with the new default. Set CUCIM_MAX_DECODER_THREADS=0 to restore the previous behavior, or set it to a specific value to tune for your workload.

cpp/plugins/cucim.kit.cuslide2/src/cuslide/nvimgcodec/nvimgcodec_tiff_parser.cpp
cdinea: thank you for the suggestion @jantonguirao - The comment now correctly says num cpu cores per the nvimgcodec.h documentation for max_num_cpu_threads in thsi commit https://github.com/rapidsai/cucim/pull/1085/commits/c0f2b465408feef5059a416227745f5c2b54b65f

cpp/plugins/cucim.kit.cuslide2/src/cuslide/cuslide.cpp
```diff
@@ -99,14 +98,14 @@ static bool CUCIM_ABI parser_parse(CuCIMFileHandle_ptr handle_ptr, cucim::io::fo
...
-    // Try ImageDescription first (works with nvImageCodec 0.7.0+)
+    // Try ImageDescription first
     bool is_aperio_svs = (tif->ifd(0)->image_description().rfind("Aperio", 0) == 0);
 
     // Detect if this is a Philips TIFF file
     // Philips TIFF also has multiple SubfileType=0 (by design)
     bool is_philips_tiff = (tif->tiff_type() == cuslide::tiff::TiffType::Philips);
 
-    // Fallback detection for nvImageCodec 0.6.0: check for multiple resolution levels
+    // Fallback detection: check for multiple resolution levels
```
jakirkham: Is this still needed for more recent nvImageCodec versions?

cpp/plugins/cucim.kit.cuslide2/src/cuslide/nvimgcodec/nvimgcodec_decoder.cpp
```diff
@@ -347,7 +347,7 @@ bool decode_ifd_region_nvimgcodec(const IfdInfo& ifd_info,
         output_image_info.plane_info[0].num_channels = num_channels;
         output_image_info.plane_info[0].row_stride = row_stride;
         output_image_info.plane_info[0].sample_type = NVIMGCODEC_SAMPLE_DATA_TYPE_UINT8;
-        // Note: buffer_size removed in nvImageCodec v0.7.0 - size is inferred from plane_info
+        // Note: size is inferred from plane_info (no explicit buffer_size field)
```
jakirkham: How confident are we of this inference? Should we discuss a different approach with the nvimgcodec team?

cpp/plugins/cucim.kit.cuslide2/src/cuslide/nvimgcodec/nvimgcodec_tiff_parser.cpp
```diff
@@ -63,8 +66,8 @@ namespace cuslide2::nvimgcodec
 //      NvimgcodecLoadSymbol() + dladdr() (not the stub address)
 //   3. Probes <lib_dir>/extensions/ and <lib_dir>/../extensions/
 //
-// TODO: Remove this workaround once cuCIM upgrades to nvImageCodec >= 0.8.0,
-//       which fixes the extension search path for conda installs.
+// NOTE: nvImageCodec 0.8.0 fixes the extension search path for conda installs.
+// This workaround is kept as a safety net for non-standard install layouts.
```
jakirkham, line 70: Could we raise a GH issue to track?

cpp/plugins/cucim.kit.cuslide2/src/cuslide/tiff/tiff.cpp
```diff
@@ -541,16 +555,16 @@ void TIFF::resolve_vendor_format()
             is_aperio = true;
         }
 
-        // Method 2: Check metadata_blobs for Aperio (kind=1)
-        // This includes the workaround for nvImageCodec 0.6.0 misclassifying Aperio as Leica
+        // Method 2: Check metadata_blobs for Aperio (MED_APERIO)
+        // This includes a workaround for nvImageCodec misclassifying Aperio as Leica
```
jakirkham: Has this been raised with the nvImageCodec team?

cpp/plugins/cucim.kit.cuslide2/src/cuslide/tiff/tiff.cpp
```diff
@@ -562,8 +576,7 @@ void TIFF::resolve_vendor_format()
     }
 
     // Detect Philips TIFF
-    // NOTE: nvImageCodec 0.6.0 doesn't expose individual TIFF tags (like SOFTWARE)
-    // Workaround: Check for Philips XML in ImageDescription or use nvImageCodec metadata kind
+    // Check for Philips TIFF using SOFTWARE tag, ImageDescription XML, or nvImageCodec metadata kind
```
jakirkham, line 579: Previously this was noted as a workaround. Is that still accurate?

If so, let's leave that note in here and raise it with the nvImageCodec team

cpp/plugins/cucim.kit.cuslide2/src/cuslide/tiff/tiff.cpp
```diff
@@ -574,8 +587,7 @@ void TIFF::resolve_vendor_format()
             is_philips = true;
         }
 
-        // Method 2: Check for Philips XML structure in ImageDescription
-        // (Workaround for nvImageCodec 0.6.0 where SOFTWARE tag is not available)
+        // Method 2: Check for Philips XML structure in ImageDescription (fallback)
```
jakirkham, line 590: Is this still a workaround? If so, have we raised this with the nvImageCodec team?

jakirkham (commented):
Thanks Cristiana! 🙏

Had a few questions below.

In particular it looks like there are cases where workarounds are now becoming part of the codebase going forward. Would like some more clarity on those and what the plan is regarding resolving them upstream in nvImageCodec.

cpp/plugins/cucim.kit.cuslide2/src/cuslide/nvimgcodec/nvimgcodec_tiff_parser.cpp
cdinea: thank you for the suggestion @jantonguirao- renamed to kHeuristicMaxThreads which better describes its role as a cap on the fair-share calculation rather than a default value. Updated in this commit https://github.com/rapidsai/cucim/pull/1085/commits/76eb4c4b8dad76667fbb682d04a8f7273db6afce  along with the num_cores doc fix

cpp/plugins/cucim.kit.cuslide2/src/cuslide/nvimgcodec/nvimgcodec_tiff_parser.cpp
cdinea: thank you @jantonguirao - moved the success-path print to #ifdef DEBUG only, consistent with the heuristic debug print below. Warnings for malformed/negative values remain unconditional on stderr since those indicate misconfiguration. Fixed in this https://github.com/rapidsai/cucim/pull/1085/commits/968fdb7f8d79da31e525b5c8ae10efd95a94d587

cpp/plugins/cucim.kit.cuslide2/src/cuslide/tiff/tiff.cpp
cdinea: thank you @jantonguirao - extracted the blob map into a local pointer at the top of the Generic block. Each branch now reads as blobs->find(kMetadataKindMedVentana) != blobs->end() instead of the repeated long expression. Fixed in this commit https://github.com/rapidsai/cucim/pull/1085/commits/3899d8a2ca6b7dbf41d7e2e38a941e6b8e284bfc

cpp/plugins/cucim.kit.cuslide2/src/cuslide/cuslide.cpp
cdinea: Thank you for the review @jakirkham - the primary Aperio detection (ImageDescription prefix + MED_APERIO metadata blob) is reliable on nvImageCodec 0.8.0, so the fallback is no longer needed for Aperio specifically. However, it still catches a legitimate edge case: generic pyramidal TIFFs that aren't any recognized vendor format but have a valid multi-resolution structure. Without it, those files would hit the SubfileType=0 validation error.

That said, the comment is misleading — it says "treat as Aperio/SVS" when really it means "treat as a valid pyramid." I'll update the comment to reflect the actual intent, and I'll add a TODO to revisit whether this heuristic can be removed once we're confident all pyramid formats are covered by the `is_known_multi_ifd_format check`

cpp/plugins/cucim.kit.cuslide2/src/cuslide/nvimgcodec/nvimgcodec_decoder.cpp
cdinea: thank you for the prompt review @jakirkham - this isn't an inference on our part — it's an intentional API change. In nvImageCodec v0.7.0 the buffer_size field was removed from nvimgcodecImageInfo_t because it was redundant with the plane geometry. The required buffer size is deterministic: row_stride * height per plane, where both values are set explicitly in plane_info. We allocate exactly that and nvImageCodec validates the buffer against plane_info at decode time, so there's no ambiguity or guesswork involved.

I've updated the comment to make this clearer in this commit https://github.com/rapidsai/cucim/pull/1085/commits/c8ba2f30e8eb00cf6b4098a860649442e7040b64 — the old wording ("inferred") was misleading

cdinea: fyi @jantonguirao @mkepa-nv

cpp/plugins/cucim.kit.cuslide2/src/cuslide/tiff/tiff.cpp
cdinea: thank you for the feedback @jakirkham - on more investigation, the "misclassifying Aperio as Leica" claim is not accurate. What actually happened: the original v0.6.0 code used hardcoded magic numbers (find(1) for Aperio, find(2) for Philips), but the nvimgcodecMetadataKind_t enum shifted between v0.6.0 and v0.8.0 — new entries (TIFF_TAG, ICC_PROFILE) were inserted before the MED_* values, changing their integer values. In v0.6.0, MED_APERIO was 3 (not 1), so the original magic numbers were wrong regardless. This has since been fixed by using the named constants (NVIMGCODEC_METADATA_KIND_MED_APERIO, etc.)

cpp/plugins/cucim.kit.cuslide2/src/cuslide/tiff/tiff.cpp
cdinea: thank you for the feedback @jakirkham  -no, it's no longer a workaround. The old comment referencing v0.6.0's limitation was stale. With v0.8.0, SOFTWARE tags are exposed directly

cpp/plugins/cucim.kit.cuslide2/src/cuslide/tiff/tiff.cpp
cdinea: thank you for the feedback @jakirkham - not a workaround for an nvImageCodec limitation - the Philips XML check in ImageDescription is a legitimate defense-in-depth fallback: some Philips TIFF files don't carry "Philips" in the SOFTWARE tag but do embed a characteristic XML structure (<DataObject ObjectType="DPUfsImport">) in ImageDescription

cpp/plugins/cucim.kit.cuslide2/src/cuslide/tiff/tiff.cpp
```diff
@@ -616,6 +626,68 @@ void TIFF::resolve_vendor_format()
...
+            auto dot_pos = path_str.rfind('.');
+            if (dot_pos != std::string::npos)
+            {
+                file_ext = path_str.substr(dot_pos);
+                std::transform(file_ext.begin(), file_ext.end(), file_ext.begin(), ::tolower);
+            }
+        }
+
+        const decltype(nvimgcodec_parser_->get_metadata_blobs(0))* blobs =
+            nvimgcodec_parser_ ? &nvimgcodec_parser_->get_metadata_blobs(0) : nullptr;
```
jakirkham: Seeing [the following error on CI]( https://github.com/rapidsai/cucim/actions/runs/26547146206/job/78201495122?pr=1085#step:13:3102 ):

```cpp
/__w/cucim/cucim/cpp/plugins/cucim.kit.cuslide2/src/cuslide/tiff/tiff.cpp: In member function 'void cuslide::tiff::TIFF::resolve_vendor_format()':
/__w/cucim/cucim/cpp/plugins/cucim.kit.cuslide2/src/cuslide/tiff/tiff.cpp:644:68: error: cannot declare pointer to 'const class std::map<int, cuslide2::nvimgcodec::IfdInfo::MetadataBlob>&'
  644 |         const decltype(nvimgcodec_parser_->get_metadata_blobs(0))* blobs =
      |                                                                    ^~~~~
```

cdinea: Thanks for flagging this @jakirkham -  get_metadata_blobs() returns a reference, so decltype givees us a reference type and you can't form a pointer to that. Wrapped it with std::remove_reference_t and added the missing <type_traits> include. Should be good now, pused the fix and restarted CI

msarahan (approved):

cpp/plugins/cucim.kit.cuslide2/src/cuslide/cuslide.cpp
```diff
@@ -99,24 +98,27 @@ static bool CUCIM_ABI parser_parse(CuCIMFileHandle_ptr handle_ptr, cucim::io::fo
...
-    // Aperio SVS files typically have 3-6 IFDs with multiple resolution levels
-    // If we have multiple IFDs and they look like a pyramid, treat as Aperio/SVS
+    // Fallback: detect unrecognised pyramidal TIFFs that were not caught by
+    // ImageDescription or nvImageCodec metadata-blob detection above.
+    // If the file has >=3 IFDs forming a resolution pyramid, assume it is a
+    // valid multi-resolution WSI so it is not rejected by the SubfileType=0
+    // single-main-image constraint below.
+    // TODO: revisit once all known pyramid formats are covered by the
+    //       is_known_multi_ifd_format check — this heuristic may become unnecessary.
     if (!is_aperio_svs && ifd_count >= 3 && level_count >= 3)
```
jakirkham, line 115: Let's raise a GH issue to track this

jakirkham (approved):
Thanks Cristiana! 🙏

The cleanups look good. Am happy to move ahead with these changes.

Let's continue to follow up with upstream on improvements.

Also let's file GH issues for any TODOs or other notes that have action items.

cpp/plugins/cucim.kit.cuslide2/src/cuslide/cuslide.cpp
cdinea: thank you for the feedback and for the prompt review @jakirkham - tracking issue here https://github.com/rapidsai/cucim/issues/1094

PR #1087, opened 2026-05-22
add large image support to `cucim.skimage.measure.label` (support 64-bit indexing)

### LINKED ISSUE #1076: [FEA] adopt ndimage.label large array support from CuPy
`cupyx.scipy.ndimage.label` is being updated to support large arrays (>= 2**32 elements)
https://github.com/cupy/cupy/pull/9859

We should port that fix to the vendored version here and update the variant under `cucim.skimage.measure.label` accordingly.

grlee77:
closes #1076

CuPy recently updated `cupyx.scipy.ndimage.label` to support 64-bit indexing when necessary to support large image sizes. See a more detailed description for the changes made in the upstream PR:
https://github.com/cupy/cupy/pull/9859

This MR makes the exact same change to the very similar kernels underlying `cucim.skimage.measure.label`. The reason we don't use CuPy kernels directly for cuCIM's `label` is that the CuPy API is for binary image labeling only (i.e. all foreground pixels are treated the same regardless of intensity). The version of `label` in the scikit-image API also supports a grayscale mode where neighboring pixels are only connected if the grayscale intensity is actually equal. Thus we have modified kernels that support that logic. The indexing concern and solution is the same as in the upstream PR, though.


python/cucim/src/cucim/skimage/measure/tests/test_ccomp.py
```diff
@@ -1,15 +1,19 @@
 # SPDX-FileCopyrightText: 2009-2022 the scikit-image team
-# SPDX-FileCopyrightText: Copyright (c) 2021-2025, NVIDIA CORPORATION. All rights reserved.
+# SPDX-FileCopyrightText: Copyright (c) 2021-2026, NVIDIA CORPORATION. All rights reserved.
 # SPDX-License-Identifier: Apache-2.0 AND BSD-3-Clause
 
 # Note: These test cases originated in skimage/morphology/tests/test_ccomp.py
 
 import cupy as cp
-
-# import numpy as np
```
jakirkham, line 9: Guessing we just never needed NumPy here

python/cucim/src/cucim/skimage/measure/tests/test_ccomp.py
jakirkham: Can we drop these too?

jakirkham (commented):
Thanks Greg! 🙏

Generally looks reasonable

Had a couple minor questions on cleanups

python/cucim/src/cucim/skimage/measure/tests/test_ccomp.py
grlee77: yes

python/cucim/src/cucim/skimage/measure/tests/test_ccomp.py
grlee77: done, thanks

grlee77:
Thanks @jakirkham, please take another look when you have a chance

jakirkham (approved):
Thanks Greg! 🙏

LGTM

PR #1105, opened 2026-06-24
Ignore CuPy NumPy shape deprecation warning

bdice:
## Summary

Fixes NumPy 2.5 `DeprecationWarning` failures triggered when pytest promotes deprecations to errors.

This PR:

- updates cucim's vendored pad helper to use `reshape` instead of assigning to `ndarray.shape`
- filters the same NumPy 2.5 warning when it originates from upstream CuPy
- filters the same NumPy 2.5 warning when it originates from upstream scikit-image marching-cubes lookup-table setup

## Root Cause

The failing CI job used Python 3.12, NumPy 2.5.0, and CuPy 14.1.1. NumPy 2.5 warns when code assigns directly to `ndarray.shape`; cucim's pytest config treats `DeprecationWarning` as an error.

The extracted failures came from three locations:

- `cucim/skimage/_vendored/pad.py::_as_pairs`
- `cupy/_padding/pad.py::_as_pairs`
- `skimage/measure/_marching_cubes_lewiner.py`

The cucim-owned vendored helper is fixed directly. The CuPy and scikit-image occurrences are upstream dependency warnings, so they are filtered narrowly by module.

## Validation

Created an isolated conda environment at `.conda-cucim-cupy-np25` with Python 3.12.13, NumPy 2.5.0, CuPy 14.1.1, and the cucim test dependencies.

Extracted the failing node IDs from the CI log for `actions/runs/28078321887/job/83129420085`:

```text
1062 unique failing node IDs
```

Ran the exact extracted subset locally with GPU access outside the sandbox:

```bash
PYTHONPATH=src /home/bdice/code/cucim/.conda-cucim-cupy-np25/bin/python -m pytest -s -q -c pyproject.toml --tb=short -x <extracted-node-ids>
```

Result:

```text
1062 passed, 8 warnings in 133.77s (0:02:13)
```

Also ran:

```bash
git diff --check
```

### DIFF
--- a/python/cucim/pyproject.toml
+++ b/python/cucim/pyproject.toml
@@ -148,6 +148,10 @@ filterwarnings = [
     "error::DeprecationWarning",
     # https://github.com/cupy/cupy/blob/main/cupy/testing/_helper.py#L56
     "ignore:pkg_resources is deprecated as an API:DeprecationWarning",
+    # https://github.com/cupy/cupy/blob/v14.1.1/cupy/_padding/pad.py#L408
+    "ignore:Setting the shape on a NumPy array:DeprecationWarning:cupy",
+    # https://github.com/scikit-image/scikit-image/blob/v0.26.0/skimage/measure/_marching_cubes_lewiner.py#L237
+    "ignore:Setting the shape on a NumPy array:DeprecationWarning:skimage",
 ]
 
 [tool.ruff]
--- a/python/cucim/src/cucim/skimage/_vendored/pad.py
+++ b/python/cucim/src/cucim/skimage/_vendored/pad.py
@@ -1,6 +1,6 @@
 # SPDX-FileCopyrightText: Copyright (c) 2015 Preferred Infrastructure, Inc.
 # SPDX-FileCopyrightText: Copyright (c) 2015 Preferred Networks, Inc.
-# SPDX-FileCopyrightText: Copyright (c) 2023-2025, NVIDIA CORPORATION. All rights reserved.
+# SPDX-FileCopyrightText: Copyright (c) 2023-2026, NVIDIA CORPORATION. All rights reserved.
 # SPDX-License-Identifier: Apache-2.0 AND MIT
 
 """version of cupy.pad that dispatches to elementwise kernels for some boundary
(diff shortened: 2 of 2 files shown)


gforsyth (approved):
:shipit: 

This will be fixed upstream in `cupy`

seberg:
Hmmmm, seems there is a copy inside cucim, which needs fixing.  Maybe one could even avoid `cupy.pad` as well then by using that throughout, although to me ignoring seems totally fine as well.

```diff
diff --git a/python/cucim/src/cucim/skimage/_vendored/pad.py b/python/cucim/src/cucim/skimage/_vendored/pad.py
index 28bf916..dfd3be4 100644
--- a/python/cucim/src/cucim/skimage/_vendored/pad.py
+++ b/python/cucim/src/cucim/skimage/_vendored/pad.py
@@ -408,8 +408,7 @@ def _as_pairs(x, ndim, as_index=False):
 
     # Converting the array with `tolist` seems to improve performance
     # when iterating and indexing the result (see usage in `pad`)
-    x_view = x.view()
-    x_view.shape = (ndim, 2)
+    x_view = x.reshape(ndim, 2)
     return x_view.tolist()
```

jakirkham:
Thanks Bradley! 🙏

@grlee77 could you please take a look?

grlee77:
@seberg, I need to check if we still need this vendored `pad` version.  There are some enhancements in some of the vendored ndimage code so we definitely need to keep those, but for `pad` I am not sure until I take a closer look.

grlee77:
Thanks @bdice, please also make the fix suggested by @seberg for the vendored copy of `pad`.

The vendored version was based on the CuPy one, but has a specialized element-wise kernel path for a subset of the boundary modes that is automatically selected for performance depending on the internal `_use_elementwise_kernel` helper function.

jakirkham:
Thanks Greg! 🙏

> ...a specialized element-wise kernel path for a subset of the boundary modes that is automatically selected for performance depending on the internal `_use_elementwise_kernel` helper function.

Independent of this PR, but do we see a path to upstreaming some (or all) of this logic to CuPy?

Admittedly this may be complicated if logic higher up in an particular image operation is influencing how the padding is done

jakirkham:
Looked through the cases where we are doing `shape` assignment and agree with Sebastian we should be able to just use `reshape`

Have done this in PR: https://github.com/rapidsai/cucim/pull/1106

Would appreciate if all of you took a look 🙂

grlee77 (approved):
Thanks, this warning ignore looks good to me

It looks like @jakirkham is addressing other shape assignment issues in #1106

jakirkham:
Think we need both fixes in the same PR. Have seen some issues with CI if both are not present

Went ahead and merged this PR into PR: https://github.com/rapidsai/cucim/pull/1106

Hope that is ok

bdice:
Thanks @jakirkham! Closing this in favor of #1106.

PR #1106, opened 2026-06-24
Replace `shape` assignment with `reshape` calls

jakirkham:
NumPy is deprecating direct `shape` assignment. The underlying code path simply goes through `reshape`. So just switch to `reshape` to eliminate the deprecation warning and future proof the code.

ref: https://github.com/numpy/numpy/blob/2a674fc909ef47631e4f6550e25817731534987f/numpy/_core/src/multiarray/getset.c#L64

### DIFF
--- a/examples/python/gds_whole_slide/demo_implementation.py
+++ b/examples/python/gds_whole_slide/demo_implementation.py
@@ -92,7 +92,7 @@ def _decode(data, tile_shape, truncation_slices):
     """
     if not hasattr(data, "__cuda_array_interface__"):
         data = np.frombuffer(data, np.uint8)
-    data = data.reshape(tile_shape, copy=False)
+    data = data.reshape(tile_shape)
     # truncate any tiles that extend past the image boundary
     if truncation_slices:
         data = data[truncation_slices]
--- a/examples/python/gds_whole_slide/demo_implementation.py
+++ b/examples/python/gds_whole_slide/demo_implementation.py
@@ -92,7 +92,7 @@ def _decode(data, tile_shape, truncation_slices):
     """
     if not hasattr(data, "__cuda_array_interface__"):
         data = np.frombuffer(data, np.uint8)
-    data = data.reshape(tile_shape)
+    data = data.reshape(tile_shape, copy=False)
     # truncate any tiles that extend past the image boundary
     if truncation_slices:
         data = data[truncation_slices]
--- a/examples/python/gds_whole_slide/demo_implementation.py
+++ b/examples/python/gds_whole_slide/demo_implementation.py
@@ -1,4 +1,4 @@
-# SPDX-FileCopyrightText: Copyright (c) 2022-2025, NVIDIA CORPORATION. All rights reserved.
+# SPDX-FileCopyrightText: Copyright (c) 2022-2026, NVIDIA CORPORATION. All rights reserved.
 # SPDX-License-Identifier: Apache-2.0
 
 import os
--- a/examples/python/gds_whole_slide/demo_implementation.py
+++ b/examples/python/gds_whole_slide/demo_implementation.py
@@ -92,7 +92,7 @@ def _decode(data, tile_shape, truncation_slices):
     """
     if not hasattr(data, "__cuda_array_interface__"):
         data = np.frombuffer(data, np.uint8)
-    data.shape = tile_shape
+    data = data.reshape(tile_shape)
     # truncate any tiles that extend past the image boundary
     if truncation_slices:
         data = data[truncation_slices]
--- a/python/cucim/pyproject.toml
+++ b/python/cucim/pyproject.toml
@@ -148,6 +148,10 @@ filterwarnings = [
     "error::DeprecationWarning",
     # https://github.com/cupy/cupy/blob/main/cupy/testing/_helper.py#L56
     "ignore:pkg_resources is deprecated as an API:DeprecationWarning",
+    # https://github.com/cupy/cupy/blob/v14.1.1/cupy/_padding/pad.py#L408
+    "ignore:Setting the shape on a NumPy array:DeprecationWarning:cupy",
+    # https://github.com/scikit-image/scikit-image/blob/v0.26.0/skimage/measure/_marching_cubes_lewiner.py#L237
+    "ignore:Setting the shape on a NumPy array:DeprecationWarning:skimage",
 ]
 
 [tool.ruff]
--- a/python/cucim/src/cucim/skimage/_vendored/pad.py
+++ b/python/cucim/src/cucim/skimage/_vendored/pad.py
@@ -1,6 +1,6 @@
 # SPDX-FileCopyrightText: Copyright (c) 2015 Preferred Infrastructure, Inc.
 # SPDX-FileCopyrightText: Copyright (c) 2015 Preferred Networks, Inc.
-# SPDX-FileCopyrightText: Copyright (c) 2023-2025, NVIDIA CORPORATION. All rights reserved.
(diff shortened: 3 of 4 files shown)


examples/python/gds_whole_slide/demo_implementation.py
```diff
@@ -92,7 +92,7 @@ def _decode(data, tile_shape, truncation_slices):
     """
     if not hasattr(data, "__cuda_array_interface__"):
         data = np.frombuffer(data, np.uint8)
-    data.shape = tile_shape
+    data = data.reshape(tile_shape, copy=False)
```
jakirkham: [The current behavior of `.shape` assignment]( https://github.com/numpy/numpy/blob/2a674fc909ef47631e4f6550e25817731534987f/numpy/_core/src/multiarray/getset.c#L63-L74 ) is to call `reshape` and error if a copy occurs. Equivalent behavior can be implemented by calling [`.reshape(...)`]( https://numpy.org/doc/stable/reference/generated/numpy.ndarray.reshape.html#numpy.ndarray.reshape ) with the `copy` keyword argument as `False`

grlee77 (approved):
Thanks, @jakirkham this changes to avoid shape assignment look good to me

seberg (approved):
Thanks, looks all better and safe.  Not sure if the last can see NumPy arrays but it hardly matters.
(The first feels like it might be a latent fix, but probably not in practice.)

python/cucim/src/cucim/skimage/_vendored/pad.py
```diff
@@ -408,8 +408,7 @@ def _as_pairs(x, ndim, as_index=False):
 
     # Converting the array with `tolist` seems to improve performance
     # when iterating and indexing the result (see usage in `pad`)
-    x_view = x.view()
-    x_view.shape = (ndim, 2)
+    x_view = x.reshape((ndim, 2), copy=False)
     return x_view.tolist()
```
jakirkham: Given we call `.tolist(...)` here, avoiding a copy probably doesn't matter much. After all this is an array that is already created as C-contiguous above, so the copy shouldn't happen anyways.

This would better match the same upstream change: https://github.com/cupy/cupy/pull/10037

```suggestion
    x_view = x.reshape((ndim, 2))
    return x_view.tolist()
```

jakirkham: Included in commit: [`9c75cbb`]( https://github.com/rapidsai/cucim/pull/1106/changes/9c75cbb0f9d5923b5ee4a1074452a7aa1c6e46bb )

seberg: Yeah.  Reshapes are always no-copy by default if possible, so unless any of these could error it doesn't matter.
(But yes, it adds the guarantee, which may makes sense for the other cases)

examples/python/gds_whole_slide/demo_implementation.py
seberg: Hmmm, this one could be a problem, though?  Because `cupy` doesn't have `copy=False` (should probably add that).
And the `hasattr()` above looks like the input may or may not be a CuPy array?

jakirkham: Good point. Looking at [CuPy's reshape logic]( https://github.com/cupy/cupy/blob/92471d86165a936ff168d570928de4629ef76316/cupy/_core/_routines_manipulation.pyx#L359-L363 ) it appears to behave similar to NumPy's `copy=None`. Besides the NumPy case is already starting from a 1-D contiguous array so shouldn't copy anyways. So perhaps we can just drop this keyword argument

```suggestion
    data = data.reshape(tile_shape)
```

jakirkham: Included in commit: [`bfb7cd0`]( https://github.com/rapidsai/cucim/pull/1106/changes/bfb7cd0b4f1263d33542fe318e51e1ef5ba07f79 )

python/cucim/pyproject.toml
```diff
@@ -148,6 +148,10 @@ filterwarnings = [
     "error::DeprecationWarning",
     # https://github.com/cupy/cupy/blob/main/cupy/testing/_helper.py#L56
     "ignore:pkg_resources is deprecated as an API:DeprecationWarning",
+    # https://github.com/cupy/cupy/blob/v14.1.1/cupy/_padding/pad.py#L408
+    "ignore:Setting the shape on a NumPy array:DeprecationWarning:cupy",
+    # https://github.com/scikit-image/scikit-image/blob/v0.26.0/skimage/measure/_marching_cubes_lewiner.py#L237
+    "ignore:Setting the shape on a NumPy array:DeprecationWarning:skimage",
```
jakirkham, line 154: Have picked up @bdice fixes from PR: https://github.com/rapidsai/cucim/pull/1105

Was still seeing some test failures without them. Hoping by combining them we can get a clean CI run

python/cucim/src/cucim/skimage/util/_map_array.py
```diff
@@ -85,8 +85,7 @@ def map_array(input_arr, input_vals, output_vals, out=None):
             f"output array has shape {out.shape}."
         )
     try:
-        out_view = out.view()
-        out_view.shape = (-1,)  # no-copy reshape/ravel
+        out_view = out.reshape((-1,), copy=False)  # no-copy reshape/ravel
```
jakirkham: As `out` is a CuPy array and [CuPy's `reshape`]( https://docs.cupy.dev/en/stable/reference/generated/cupy.ndarray.html#cupy.ndarray.reshape ) doesn't accept the `copy` keyword argument, this fails. As we need to be sure a copy doesn't happen and there isn't a good alternative to assigning to `.shape`, this will need to revert to the original syntax.

```suggestion
        out_view = out.view()
        out_view.shape = (-1,)  # no-copy reshape/ravel
```

examples/python/gds_whole_slide/demo_implementation.py
jakirkham: Based on the discussion below, all of these were dropped because CuPy's `reshape` doesn't support the `copy` keyword argument and they were either unneeded in that context or the intended behavior could be captured other ways.

python/cucim/src/cucim/skimage/util/_map_array.py
```diff
@@ -85,7 +85,8 @@ def map_array(input_arr, input_vals, output_vals, out=None):
             f"output array has shape {out.shape}."
         )
     try:
-        out_view = out.reshape((-1,), copy=False)  # no-copy reshape/ravel
+        out_view = out.view()
+        out_view.shape = (-1,)  # no-copy reshape/ravel
```
jakirkham, line 89: Have reverted to the original syntax here as [CuPy's `reshape`]( https://docs.cupy.dev/en/stable/reference/generated/cupy.ndarray.html#cupy.ndarray.reshape ) doesn't support the `copy` keyword argument and we really must avoid a copy here. So the best way to do this with CuPy is to assign to `.shape`. This is unfortunate, but it should work as intended.

Edit - Filed an upstream request to add the `copy` keyword argument to `reshape`: https://github.com/cupy/cupy/issues/10039

python/cucim/pyproject.toml
jakirkham: Also the warnings in CuPy are being fixed by @seberg in PR: https://github.com/cupy/cupy/pull/10037

The warnings in scikit-image will be fixed in 0.27 with PR: https://github.com/scikit-image/scikit-image/pull/8020

bdice (approved):
Great, thanks!

jakirkham:
Thanks all! 🙏

PR #1115, opened 2026-07-08
Add new cucim.skimage.filters.rank module

### LINKED ISSUE #235: [FEA] cucim skimage local entropy
I wish I could use cuCIM for skimage local entropy function:
https://scikit-image.org/docs/dev/api/skimage.filters.rank.html#skimage.filters.rank.entropy

grlee77:
## Summary

This PR adds the `cucim.skimage.filters.rank` module, providing
GPU-accelerated rank filters for CuPy arrays with an API modeled after
`skimage.filters.rank`. It resolves https://github.com/rapidsai/cucim/issues/235.

The existing `cucim.skimage.filters.median` wrapper is also extended with
`behavior='rank'`, which delegates to the new rank median implementation.
`behavior='ndimage'` remains the default.

The implementation is intended to be largely compatible with scikit-image for
common rank-filter workflows. The most visible behavioral difference is
boundary handling: cuCIM uses reflected boundary extension, while scikit-image
rank filters crop neighborhoods at image boundaries. cuCIM also supports N-D
inputs and optional native processing of integer and real floating-point
dtypes, with some output-dtype and percentile edge-case behavior differing
from scikit-image. `otsu` and `windowed_histogram` are not currently
implemented.

See the detailed compatibility overview in
[`cucim.skimage.filters.rank`](python/cucim/src/cucim/skimage/filters/rank/__init__.py).

## Implementation Approach

The module uses multiple backend implementations because no single algorithm is
best across all filters, dtypes, dimensionalities, and footprint sizes.

- The elementwise backend handles the broadest set of cases: N-dimensional
  inputs, arbitrary footprint shapes, masks, shifts, and native integer and
  real floating-point dtypes. Operations that do not require sorted values use
  streaming reductions within this backend. It avoids requiring a fixed
  histogram range and is the general fallback.
- The histogram-based implementation targets high-value 2-D uint8 cases with
  fully populated, odd-sized rectangular footprints. It maintains a local
  sliding histogram and can reuse work between adjacent pixels, which is much
  more favorable for larger rectangular windows and filters that can be
  expressed from histogram counts, order statistics, prefix counts, or prefix
  sums.

By default, non-uint8 inputs are converted to uint8 with `img_as_ubyte` before
backend selection, matching scikit-image rank-filter behavior more closely and
allowing eligible inputs to use the uint8 histogram fast path. Users can opt
out with `cast_to_uint8=False` when they need native-dtype elementwise
behavior.

## Testing

The test suite provides several independent checks of correctness and backend
dispatch:

- All rank-filter APIs are compared with stored scikit-image reference results
  for 2-D inputs, with generic filters also covered in 3-D. Comparisons account
  for the documented boundary behavior and explicitly identify the few known
  algorithmic differences.
- Independent brute-force reference implementations validate the elementwise
  streaming operations and the histogram percentile, prefix-count, prefix-sum,
  bilateral, modal, and entropy operations. Forced elementwise and histogram
  results are also compared directly.
- Parameterized coverage exercises integer and floating-point dtypes, default
  uint8 conversion, native-dtype processing, output dtypes and scaling,
  overflow-sensitive sums, masks, shifts, arbitrary footprints, and trivial or
  empty neighborhoods.
- Backend tests cover automatic cutoff selection, explicit overrides,
  compatibility rejection and fallback, histogram counter widths, scratch
  partition limits, invalid inputs, and overlapping output aliases.
- Integration tests verify that `cucim.skimage.filters.median` with
  `behavior='rank'` matches `rank.median` for array and tuple footprints,
  supports caller-provided output arrays, and retains the existing
  `behavior='ndimage'` default and warning behavior.

The complete rank-filter test module currently reports 281 passed tests and 3
documented skips for known algorithmic differences.

## Reviewer Notes

The rank module includes support for generic rank filters, percentile-range
filters, bilateral rank filters, entropy, modal/majority, and related
operations. Backend selection is automatic by default, but a `backend` keyword
can force `elementwise` or `histogram` for testing, profiling, and threshold
tuning.

## cuCIM Rank Filter Benchmark Summary

Timings are from `cucim_filters_rank_results.csv` files generated by `run-nv-bench-filters-rank.sh` or `cucim_filters_rank_bench.py`.

```bash
CUCIM_BENCHMARK_MAX_DURATION=3 FOOTPRINT_SIZES=3,7,15,25,51,75 ./run-nv-bench-filters-rank.sh

python summarize_rank_benchmark_results.py \
	cucim_filters_rank_results.csv \
	--shape 1920x1080 \
	--backend auto \
	--footprint 3x3,7x7,15x15,25x25,51x51,75x75 \
	-o rank_filter_pr_benchmarks.md
```

### Speedup vs scikit-image CPU: uint8, rectangular footprint with "auto" algorithm tuning

| filter                      | image     | dtype   | backend   | 3x3     | 7x7    | 15x15   | 25x25   | 51x51   | 75x75   |
|:----------------------------|:----------|:--------|:----------|:--------|:-------|:--------|:--------|:--------|:--------|
| autolevel                   | 1920x1080 | uint8   | auto      | 860.3x  | 993.1x | 796.7x  | 171.8x  | 63.9x   | 39.2x   |
| autolevel_percentile        | 1920x1080 | uint8   | auto      | 809.7x  | 106.9x | 17.7x   | 11.7x   | 16.5x   | 20.5x   |
| enhance_contrast            | 1920x1080 | uint8   | auto      | 1172.1x | 988.7x | 691.8x  | 155.6x  | 59.4x   | 36.8x   |
| enhance_contrast_percentile | 1920x1080 | uint8   | auto      | 819.1x  | 106.9x | 17.8x   | 21.0x   | 29.2x   | 36.1x   |
| entropy                     | 1920x1080 | uint8   | auto      | 247.4x  | 108.5x | 47.9x   | 20.6x   | 24.8x   | 27.7x   |
| equalize                    | 1920x1080 | uint8   | auto      | 566.3x  | 703.1x | 585.9x  | 152.1x  | 65.6x   | 42.9x   |
| geometric_mean              | 1920x1080 | uint8   | auto      | 130.7x  | 45.4x  | 19.8x   | 26.1x   | 40.1x   | 50.5x   |
| gradient                    | 1920x1080 | uint8   | auto      | 1223.0x | 978.4x | 703.6x  | 159.0x  | 59.6x   | 37.9x   |
| gradient_percentile         | 1920x1080 | uint8   | auto      | 767.9x  | 100.6x | 17.0x   | 20.3x   | 28.9x   | 35.8x   |
| maximum                     | 1920x1080 | uint8   | auto      | 549.6x  | 570.5x | 524.5x  | 134.9x  | 59.6x   | 39.6x   |
| mean                        | 1920x1080 | uint8   | auto      | 690.5x  | 336.7x | 119.7x  | 60.3x   | 21.9x   | 14.8x   |
| mean_bilateral              | 1920x1080 | uint8   | auto      | 504.5x  | 152.5x | 44.5x   | 20.1x   | 17.7x   | 21.9x   |
| mean_percentile             | 1920x1080 | uint8   | auto      | 1062.4x | 151.5x | 25.1x   | 19.6x   | 26.5x   | 32.2x   |
| median                      | 1920x1080 | uint8   | auto      | 558.2x  | 87.1x  | 16.3x   | 19.4x   | 28.8x   | 36.0x   |
| modal (majority)            | 1920x1080 | uint8   | auto      | 803.3x  | 113.8x | 32.4x   | 41.9x   | 56.6x   | 70.0x   |
| minimum                     | 1920x1080 | uint8   | auto      | 706.5x  | 674.6x | 594.0x  | 150.1x  | 59.6x   | 45.0x   |
| noise_filter                | 1920x1080 | uint8   | auto      | 435.2x  | 417.2x | 332.6x  | 108.0x  | 51.3x   | 33.0x   |
| percentile                  | 1920x1080 | uint8   | auto      | 532.8x  | 75.8x  | 14.0x   | 17.9x   | 27.1x   | 34.9x   |
| pop                         | 1920x1080 | uint8   | auto      | 340.3x  | 574.6x | 1047.3x | 1760.2x | 3286.8x | 4824.1x |
| pop_bilateral               | 1920x1080 | uint8   | auto      | 847.5x  | 252.0x | 66.4x   | 29.2x   | 30.9x   | 37.1x   |
| pop_percentile              | 1920x1080 | uint8   | auto      | 1633.2x | 204.0x | 31.4x   | 27.7x   | 30.4x   | 33.7x   |
| subtract_mean               | 1920x1080 | uint8   | auto      | 723.3x  | 344.7x | 120.2x  | 61.6x   | 22.6x   | 13.7x   |
| subtract_mean_percentile    | 1920x1080 | uint8   | auto      | 1192.4x | 154.1x | 25.2x   | 17.8x   | 23.9x   | 29.0x   |
| sum                         | 1920x1080 | uint8   | auto      | 892.4x  | 376.2x | 122.2x  | 61.1x   | 22.4x   | 13.7x   |
| sum_bilateral               | 1920x1080 | uint8   | auto      | 551.8x  | 154.0x | 47.2x   | 20.4x   | 26.7x   | 33.3x   |
| sum_percentile              | 1920x1080 | uint8   | auto      | 1187.0x | 153.1x | 25.9x   | 31.9x   | 41.5x   | 50.8x   |
| threshold                   | 1920x1080 | uint8   | auto      | 781.0x  | 357.7x | 122.6x  | 60.4x   | 22.3x   | 13.8x   |
| threshold_percentile        | 1920x1080 | uint8   | auto      | 754.6x  | 101.8x | 17.7x   | 20.9x   | 29.0x   | 36.1x   |

We can see that the histogram-based kernels are consistently much faster than their scikit-image counterparts.

### Speedup vs scikit-image CPU: uint8, disk footprint (so only the elementwise backend can be used)

The benchmark was invoked with `backend='auto'`, but disk footprints are not
compatible with the histogram backend, so every row dispatches to elementwise.

| filter                      | image     | dtype   | backend     | 3x3     | 7x7    | 15x15   | 25x25   | 51x51   | 75x75   |
|:----------------------------|:----------|:--------|:------------|:--------|:-------|:--------|:--------|:--------|:--------|
| autolevel                   | 1920x1080 | uint8   | elementwise | 778.1x  | 803.6x | 308.7x  | 154.0x  | 55.3x   | 26.5x   |
| autolevel_percentile        | 1920x1080 | uint8   | elementwise | 774.0x  | 221.7x | 28.6x   | 10.1x   | 2.1x    | 0.9x    |
| enhance_contrast            | 1920x1080 | uint8   | elementwise | 911.9x  | 683.9x | 269.0x  | 139.1x  | 52.7x   | 25.7x   |
| enhance_contrast_percentile | 1920x1080 | uint8   | elementwise | 1056.6x | 214.2x | 28.5x   | 10.2x   | 2.2x    | 0.9x    |
| entropy                     | 1920x1080 | uint8   | elementwise | 332.3x  | 141.3x | 64.7x   | 26.5x   | 5.2x    | 2.2x    |
| equalize                    | 1920x1080 | uint8   | elementwise | 466.1x  | 567.7x | 238.3x  | 139.9x  | 57.8x   | 30.8x   |
| geometric_mean              | 1920x1080 | uint8   | elementwise | 176.4x  | 69.2x  | 21.8x   | 10.5x   | 3.7x    | 2.0x    |
| gradient                    | 1920x1080 | uint8   | elementwise | 933.4x  | 697.1x | 274.2x  | 142.1x  | 52.8x   | 25.5x   |
| gradient_percentile         | 1920x1080 | uint8   | elementwise | 1027.5x | 204.3x | 26.9x   | 9.7x    | 2.1x    | 0.8x    |
| majority                    | 1920x1080 | uint8   | elementwise | 1351.5x | 276.1x | 36.8x   | 12.6x   | 2.5x    | 1.1x    |
| maximum                     | 1920x1080 | uint8   | elementwise | 428.5x  | 427.2x | 201.7x  | 118.7x  | 50.2x   | 25.7x   |
| mean                        | 1920x1080 | uint8   | elementwise | 605.4x  | 320.7x | 170.6x  | 83.6x   | 29.6x   | 13.5x   |
| mean_bilateral              | 1920x1080 | uint8   | elementwise | 485.0x  | 154.3x | 68.4x   | 30.3x   | 10.0x   | 5.3x    |
| mean_percentile             | 1920x1080 | uint8   | elementwise | 1316.9x | 298.4x | 39.3x   | 13.4x   | 2.7x    | 1.0x    |
| median                      | 1920x1080 | uint8   | elementwise | 695.6x  | 171.7x | 25.7x   | 9.4x    | 2.0x    | 0.8x    |
| minimum                     | 1920x1080 | uint8   | elementwise | 574.4x  | 513.3x | 230.3x  | 131.0x  | 52.8x   | 26.5x   |
| modal                       | 1920x1080 | uint8   | elementwise | 1030.2x | 217.4x | 30.0x   | 10.8x   | 2.3x    | 1.1x    |
| noise_filter                | 1920x1080 | uint8   | elementwise | 392.8x  | 355.9x | 158.0x  | 98.8x   | 45.4x   | 19.6x   |
| percentile                  | 1920x1080 | uint8   | elementwise | 689.3x  | 151.9x | 22.4x   | 8.2x    | 1.9x    | 0.8x    |
| pop                         | 1920x1080 | uint8   | elementwise | 263.2x  | 392.8x | 472.4x  | 450.7x  | 260.2x  | 137.6x  |
| pop_bilateral               | 1920x1080 | uint8   | elementwise | 794.5x  | 255.6x | 101.4x  | 44.1x   | 13.7x   | 6.1x    |
| pop_percentile              | 1920x1080 | uint8   | elementwise | 2010.9x | 411.1x | 48.5x   | 15.0x   | 2.6x    | 0.9x    |
| subtract_mean               | 1920x1080 | uint8   | elementwise | 630.4x  | 322.9x | 169.8x  | 83.8x   | 29.3x   | 13.5x   |
| subtract_mean_percentile    | 1920x1080 | uint8   | elementwise | 1428.5x | 309.2x | 38.4x   | 12.9x   | 2.6x    | 1.0x    |
| sum                         | 1920x1080 | uint8   | elementwise | 776.5x  | 353.8x | 181.1x  | 84.2x   | 29.9x   | 13.5x   |
| sum_bilateral               | 1920x1080 | uint8   | elementwise | 520.8x  | 154.6x | 64.6x   | 28.7x   | 9.6x    | 5.2x    |
| sum_percentile              | 1920x1080 | uint8   | elementwise | 1437.4x | 304.8x | 37.2x   | 12.8x   | 2.5x    | 1.0x    |
| threshold                   | 1920x1080 | uint8   | elementwise | 685.3x  | 342.3x | 175.5x  | 83.5x   | 29.1x   | 13.3x   |
| threshold_percentile        | 1920x1080 | uint8   | elementwise | 946.8x  | 203.8x | 27.9x   | 10.1x   | 2.1x    | 0.8x    |

We can see that at large enough footprint size, the "elementwise" backend eventually becomes slower than scikit-image's histogram-based approach.


benchmarks/skimage/run-nv-bench-filters-rank.sh
```diff
@@ -0,0 +1,23 @@
+#!/bin/bash
+# SPDX-FileCopyrightText: Copyright (c) 2026, NVIDIA CORPORATION. All rights reserved.
```
jakirkham:
```suggestion
# SPDX-FileCopyrightText: Copyright (c) 2026, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
```

benchmarks/skimage/cucim_filters_rank_bench.py
```diff
@@ -0,0 +1,213 @@
+# SPDX-FileCopyrightText: Copyright (c) 2026, NVIDIA CORPORATION. All rights reserved.
```
jakirkham:
```suggestion
# SPDX-FileCopyrightText: Copyright (c) 2026, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
```

python/cucim/src/cucim/skimage/filters/rank/cuda/histogram_rank.cu
```diff
@@ -0,0 +1,755 @@
+/*
+ * SPDX-FileCopyrightText: Copyright (c) 2026, NVIDIA CORPORATION. All rights reserved.
```
jakirkham:
```suggestion
 * SPDX-FileCopyrightText: Copyright (c) 2026, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
```

python/cucim/src/cucim/skimage/filters/rank/tests/test_rank.py
```diff
@@ -0,0 +1,1913 @@
+# SPDX-FileCopyrightText: 2009-2022 the scikit-image team
+# SPDX-FileCopyrightText: Copyright (c) 2026, NVIDIA CORPORATION. All rights reserved.
```
jakirkham:
```suggestion
# SPDX-FileCopyrightText: Copyright (c) 2026, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
```

python/cucim/src/cucim/skimage/filters/rank/__init__.py
```diff
@@ -0,0 +1,222 @@
+# SPDX-FileCopyrightText: 2009-2022 the scikit-image team
+# SPDX-FileCopyrightText: Copyright (c) 2026, NVIDIA CORPORATION. All rights reserved.
```
jakirkham:
```suggestion
# SPDX-FileCopyrightText: Copyright (c) 2026, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
```

python/cucim/src/cucim/skimage/filters/rank/_bilateral.py
```diff
@@ -0,0 +1,215 @@
+# SPDX-FileCopyrightText: 2009-2022 the scikit-image team
+# SPDX-FileCopyrightText: Copyright (c) 2026, NVIDIA CORPORATION. All rights reserved.
```
jakirkham:
```suggestion
# SPDX-FileCopyrightText: Copyright (c) 2026, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
```

python/cucim/src/cucim/skimage/filters/rank/_generic.py
```diff
@@ -0,0 +1,860 @@
+# SPDX-FileCopyrightText: 2009-2022 the scikit-image team
+# SPDX-FileCopyrightText: Copyright (c) 2026, NVIDIA CORPORATION. All rights reserved.
```
jakirkham:
```suggestion
# SPDX-FileCopyrightText: Copyright (c) 2026, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
```

python/cucim/src/cucim/skimage/filters/rank/_histogram.py
```diff
@@ -0,0 +1,275 @@
+# SPDX-FileCopyrightText: Copyright (c) 2026, NVIDIA CORPORATION. All rights reserved.
```
jakirkham:
```suggestion
# SPDX-FileCopyrightText: Copyright (c) 2026, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
```

python/cucim/src/cucim/skimage/filters/rank/_percentile.py
```diff
@@ -0,0 +1,762 @@
+# SPDX-FileCopyrightText: 2009-2022 the scikit-image team
+# SPDX-FileCopyrightText: Copyright (c) 2026, NVIDIA CORPORATION. All rights reserved.
```
jakirkham:
```suggestion
# SPDX-FileCopyrightText: Copyright (c) 2026, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
```

python/cucim/src/cucim/skimage/filters/rank/_rank_filter.py
```diff
@@ -0,0 +1,1133 @@
+# SPDX-FileCopyrightText: 2009-2022 the scikit-image team
+# SPDX-FileCopyrightText: Copyright (c) 2026, NVIDIA CORPORATION. All rights reserved.
```
jakirkham:
```suggestion
# SPDX-FileCopyrightText: Copyright (c) 2026, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
```

python/cucim/src/cucim/skimage/filters/tests/test_median.py
```diff
@@ -1,5 +1,5 @@
 # SPDX-FileCopyrightText: 2009-2022 the scikit-image team
-# SPDX-FileCopyrightText: Copyright (c) 2021-2025, NVIDIA CORPORATION. All rights reserved.
+# SPDX-FileCopyrightText: Copyright (c) 2021-2026, NVIDIA CORPORATION. All rights reserved.
```
jakirkham:
```suggestion
# SPDX-FileCopyrightText: Copyright (c) 2021-2026, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
```

python/cucim/src/cucim/skimage/filters/__init__.pyi
```diff
@@ -1,5 +1,5 @@
 # SPDX-FileCopyrightText: 2009-2022 the scikit-image team
-# SPDX-FileCopyrightText: Copyright (c) 2023-2025, NVIDIA CORPORATION. All rights reserved.
+# SPDX-FileCopyrightText: Copyright (c) 2023-2026, NVIDIA CORPORATION. All rights reserved.
```
jakirkham:
```suggestion
# SPDX-FileCopyrightText: Copyright (c) 2023-2026, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
```

python/cucim/src/cucim/skimage/filters/_median.py
```diff
@@ -1,5 +1,5 @@
 # SPDX-FileCopyrightText: 2009-2022 the scikit-image team
-# SPDX-FileCopyrightText: Copyright (c) 2021-2025, NVIDIA CORPORATION. All rights reserved.
+# SPDX-FileCopyrightText: Copyright (c) 2021-2026, NVIDIA CORPORATION. All rights reserved.
```
jakirkham:
```suggestion
# SPDX-FileCopyrightText: Copyright (c) 2021-2026, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
```

jakirkham (commented):
Thanks Greg! 🙏

The style checker is failing. Can you please apply these changes or run `pre-commit` locally to address these issues?

That should unblock CI to test the code further

python/cucim/src/cucim/skimage/filters/rank/_rank_filter.py
```diff
@@ -0,0 +1,1133 @@
...
+                int pop_n = 0;
+                int pop_cumsum = 0;
+                int pop_j = 0;
+                while (pop_j < {filter_size}) {{
+                    X pop_val = values[pop_j];
+                    int pop_gs = 0;
+                    while (pop_j < {filter_size} && values[pop_j] == pop_val) {{
+                        pop_j++;
+                        pop_gs++;
+                    }}
```
gigony, line 676: ### [P1] Included NaNs can cause a non-terminating GPU kernel

**AI-assisted review comment (finding 1/5).** Based on static analysis; the CUDA reproduction below has not been run.

This affects `rank.pop_percentile` with a non-full percentile interval, an included NaN, and `cast_to_uint8=False`. The full-range streaming count does not enter this loop.

For a NaN, `values[pop_j] == pop_val` is false even when both expressions reference the same element. The inner loop therefore consumes nothing. Neither `pop_j` nor the outer loop's termination condition changes, so the GPU thread runs indefinitely.

Kernel launch can return asynchronously; the hang becomes visible on a subsequent synchronization, such as `.get()`. A watchdog may terminate the kernel instead, but that is not normal filter completion.

<details>
<summary>Reproduction / regression example (not run on CUDA)</summary>

**Run this only in an isolated GPU process with an external timeout or watchdog. The unpatched kernel can stall the worker; do not add the unguarded reproduction directly to a shared test run.**

```python
import cupy as cp
from cucim.skimage.filters import rank

image = cp.full((3, 3), cp.nan, dtype=cp.float32)
footprint = cp.ones((3, 3), dtype=bool)
result = rank.pop_percentile(
    image,
    footprint,
    p0=0.1,
    p1=0.9,
    backend="elementwise",
    cast_to_uint8=False,
)
result.get()
```

The code path shows that synchronization cannot finish normally on the unpatched implementation.

</details>

<details>
<summary>Proposed fix and regression coverage</summary>

Consume the first element of every group unconditionally, then extend the group while subsequent values compare equal. This guarantees progress and preserves the existing grouping/counting behavior for finite values and infinities. Apply the change to both generated variants.

**NaN policy:** This minimal termination fix treats each encountered NaN as a singleton group. It does not establish an ignore-NaN or propagate-NaN statistical contract. If the intended API instead rejects NaNs, add explicit validation and documentation rather than relying on an equality loop that cannot terminate.

```diff
diff --git a/python/cucim/src/cucim/skimage/filters/rank/_rank_filter.py b/python/cucim/src/cucim/skimage/filters/rank/_rank_filter.py
--- a/python/cucim/src/cucim/skimage/filters/rank/_rank_filter.py
+++ b/python/cucim/src/cucim/skimage/filters/rank/_rank_filter.py
@@ -641,7 +641,8 @@
                 int pop_j = 0;
                 while (pop_j < iv) {{
                     X pop_val = values[pop_j];
-                    int pop_gs = 0;
+                    int pop_gs = 1;
+                    pop_j++;
                     while (pop_j < iv && values[pop_j] == pop_val) {{
                         pop_j++;
                         pop_gs++;
@@ -661,7 +662,8 @@
                 int pop_j = 0;
                 while (pop_j < {filter_size}) {{
                     X pop_val = values[pop_j];
-                    int pop_gs = 0;
+                    int pop_gs = 1;
+                    pop_j++;
                     while (pop_j < {filter_size} && values[pop_j] == pop_val) {{
                         pop_j++;
                         pop_gs++;
```

#### Regression coverage

- Verify bounded completion for float32 and float64 with included NaNs, both with and without a mask.
- Cover an all-NaN neighborhood, mixed finite/NaN values, and a one-element neighborhood.
- Verify that masked-out NaNs do not contribute.
- Compare finite repeated-value groups before and after the change to ensure population semantics remain unchanged.
- Keep the termination test isolated until the patch is installed; assert the chosen NaN output semantics separately once that policy is agreed.

</details>

python/cucim/src/cucim/skimage/filters/rank/_generic.py
```diff
@@ -0,0 +1,860 @@
...
+    elif shift_z != 0:
+        if image.ndim < 3:
+            raise ValueError(
+                "shift_z is only valid for 3D or higher dimensional images"
+            )
+        shifts_list = [0] * image.ndim
+        shifts_list[0] = shift_z
+        shifts_list[1] = shift_y
+        shifts_list[2] = shift_x
+        shifts = tuple(shifts_list)
```
gigony, line 165: ### [P2] Nonzero `shift_z` changes the meaning of the X/Y shifts

**AI-assisted review comment (finding 2/5).** Based on static analysis; the CUDA reproduction below has not been run.

The nonzero-Z branch disagrees with the normal X/Y mapping in `_percentile.py:282-288`. The public docstring assigns Y to axis 0 and X to axis 1.

For a 3-D input:

| Call | Current origin |
| --- | --- |
| `shift_x=1, shift_y=0, shift_z=0` | `(0, 1, 0)` |
| `shift_x=1, shift_y=0, shift_z=1` | `(1, 0, 1)` |

Adding a Z shift also removes the shift from axis 1 and moves the X/Y interpretation to different axes. This silently changes the selected neighborhood; with anisotropic footprints, it can also cause an otherwise unrelated shift to fail origin validation.

<details>
<summary>Reproduction / regression example (not run on CUDA)</summary>

The volume below varies only along axis 1. A shift along either other axis cannot change its values, so adding a Z shift should leave the interior result unchanged.

```python
import cupy as cp
from cucim.skimage.filters import rank

image = cp.broadcast_to(
    cp.arange(5, dtype=cp.uint8)[None, :, None], (5, 5, 5)
).copy()
footprint = cp.ones((3, 3, 3), dtype=bool)

without_z = rank.mean(image, footprint, shift_x=1)
with_z = rank.mean(image, footprint, shift_x=1, shift_z=1)
explicit = rank.mean(image, footprint, shifts=(0, 1, 1))

assert without_z[2, 2, 2].item() == 1
assert with_z[2, 2, 2].item() == 1
cp.testing.assert_array_equal(with_z, explicit)
```

On the current code path, `with_z[2, 2, 2]` evaluates to 2 instead of 1.

</details>

<details>
<summary>Proposed fix and regression coverage</summary>

Preserve the PR's documented convention: Y uses axis 0, X uses axis 1, and Z extends it on axis 2. Remaining axes stay at zero unless `shifts` is explicitly supplied. This fixes the inconsistency without changing the documented X/Y behavior of the shared percentile/bilateral wrappers.

```diff
diff --git a/python/cucim/src/cucim/skimage/filters/rank/_generic.py b/python/cucim/src/cucim/skimage/filters/rank/_generic.py
--- a/python/cucim/src/cucim/skimage/filters/rank/_generic.py
+++ b/python/cucim/src/cucim/skimage/filters/rank/_generic.py
@@ -145,5 +145,5 @@
         shifts_list = [0] * image.ndim
-        shifts_list[0] = shift_z
-        shifts_list[1] = shift_y
-        shifts_list[2] = shift_x
+        shifts_list[0] = shift_y
+        shifts_list[1] = shift_x
+        shifts_list[2] = shift_z
         shifts = tuple(shifts_list)
```

#### Regression coverage

- Compare legacy X/Y/Z keywords with equivalent explicit `shifts` for zero, positive, and negative shifts.
- Cover both 3-D and 4-D inputs and anisotropic footprints.
- Retain errors for nonzero `shift_z` on a 2-D image and for combining nonzero legacy keywords with explicit `shifts`.
- Clarify axis 2 in the `shift_z` documentation when integrating the fix.

</details>

python/cucim/src/cucim/skimage/filters/rank/_rank_filter.py
```diff
@@ -0,0 +1,1133 @@
...
+            """
+    elif operation == "pop":
+        pre = "int n_vals = 0;"
+        update = "n_vals++;"
+        post = """
+            if (n_vals == 0) {
+                y = cast<Y>(x[i]);
+                return;
+            }
+            y = cast<Y>(n_vals);
```
gigony, line 173: ### [P2] Empty masked populations return image intensities instead of zero

**AI-assisted review comment (finding 3/5).** Based on static analysis; the CUDA reproduction below has not been run.

This concerns a valid footprint whose mask excludes every neighborhood sample, not an all-zero footprint. The bilateral and partial-percentile population variants have analogous empty-count guards.

The empty-neighborhood fallback copies the center pixel. That may be a deliberate fallback for some image-valued operations, but it is incorrect for a population: a neighborhood containing no eligible samples has population zero, independently of image intensity.

For a 3x3 footprint, an image filled with 255 and an all-false mask currently produces a population of 255 rather than zero. This is not a boundary-extension difference; it also occurs in interior pixels. The same issue affects bilateral and percentile population variants.

The brute-force helper also copies the input whenever its sample list is empty. An oracle using that fallback can agree with the kernel while repeating the same semantic error.

<details>
<summary>Reproduction / regression example (not run on CUDA)</summary>

```python
import cupy as cp
from cucim.skimage.filters import rank

image = cp.full((5, 5), 255, dtype=cp.uint8)
footprint = cp.ones((3, 3), dtype=bool)
mask = cp.zeros_like(image, dtype=bool)

cases = [
    (rank.pop, {}),
    (rank.pop_percentile, {"p0": 0.0, "p1": 1.0}),
    (rank.pop_percentile, {"p0": 0.1, "p1": 0.9}),
    (rank.pop_bilateral, {"s0": 10, "s1": 10}),
]
for function, kwargs in cases:
    out = cp.empty_like(image)
    result = function(image, footprint, mask=mask, out=out, **kwargs)
    assert result is out
    cp.testing.assert_array_equal(result, cp.zeros_like(image))
```

The assertions express the expected post-fix behavior; the current empty-count guards return the center intensity instead.

</details>

<details>
<summary>Proposed fix and regression coverage</summary>

Use the already-zero counters directly in the streaming population operations. Give masked percentile populations their own zero-valued empty case, leaving non-count fallbacks unchanged. Correct the brute-force oracle for the corresponding count operations as well.

```diff
diff --git a/python/cucim/src/cucim/skimage/filters/rank/_rank_filter.py b/python/cucim/src/cucim/skimage/filters/rank/_rank_filter.py
--- a/python/cucim/src/cucim/skimage/filters/rank/_rank_filter.py
+++ b/python/cucim/src/cucim/skimage/filters/rank/_rank_filter.py
@@ -169,9 +169,5 @@
         pre = "int n_vals = 0;"
         update = "n_vals++;"
         post = """
-            if (n_vals == 0) {
-                y = cast<Y>(x[i]);
-                return;
-            }
             y = cast<Y>(n_vals);
         """
@@ -260,8 +256,4 @@
         elif operation == "bilateral_pop":
             post = """
-                if (n_vals == 0) {
-                    y = cast<Y>(x[i]);
-                    return;
-                }
                 y = cast<Y>(bilat_pop);
             """
@@ -473,6 +465,13 @@
     if has_mask:
         # Runtime calculation of indices based on actual count
-        if operation in ("percentile", "threshold", "pop") or _bilateral_op:
+        if operation == "pop":
+            post = """
+                if (iv == 0) {{
+                    y = cast<Y>(0);
+                    return;
+                }}
+                sort(values, iv);"""
+        elif operation in ("percentile", "threshold") or _bilateral_op:
             post = """
                 if (iv == 0) {{
                     y = cast<Y>(x[i]);  // No valid values, keep original
diff --git a/python/cucim/src/cucim/skimage/filters/rank/tests/test_rank.py b/python/cucim/src/cucim/skimage/filters/rank/tests/test_rank.py
--- a/python/cucim/src/cucim/skimage/filters/rank/tests/test_rank.py
+++ b/python/cucim/src/cucim/skimage/filters/rank/tests/test_rank.py
@@ -78,6 +78,10 @@
             center = int(image[row, col])
             if not values:
-                out[row, col] = image[row, col]
+                out[row, col] = (
+                    0
+                    if operation in {"pop", "pop_bilateral"}
+                    else image[row, col]
+                )
             elif operation == "minimum":
                 out[row, col] = min(values)
             elif operation == "maximum":
```

#### Regression coverage

- Use nonzero images; a zero-valued input hides the current fallback error.
- Cover all-false masks and masks with holes large enough to leave some neighborhoods empty.
- Cover full and partial percentile intervals, bilateral populations, native floating-point images, and wider output arrays.
- Verify that nonempty counts are unchanged and caller-provided output arrays are still respected.
- Treat all-zero footprints separately: their validation is a different code path from a valid footprint with an empty masked neighborhood.

</details>

python/cucim/src/cucim/skimage/filters/rank/_rank_filter.py
```diff
@@ -0,0 +1,1133 @@
...
+            """
+        else:
+            post += f"""
+                X min_val = values[{idx_start}];
+                X max_val = values[{idx_end - 1}];
+                X g = x[i];
+                if (max_val - g < g - min_val) {{
+                    y = cast<Y>(max_val);
+                }} else {{
+                    y = cast<Y>(min_val);
```
gigony, line 601: ### [P2] Native unsigned contrast filtering can select the farther endpoint

**AI-assisted review comment (finding 4/5).** Based on static analysis; the CUDA reproduction below has not been run.

This affects native `uint32`/`uint64` processing with `cast_to_uint8=False`; the default uint8 conversion hides it. The same comparison also appears in the masked percentile and streaming contrast kernels.

The subtractions are evaluated in the unsigned input type before the comparison. Unlike uint8/uint16, uint32/uint64 do not promote to a signed `int` capable of representing the negative differences involved here.

If the center is above the selected maximum, `max_val - g` wraps to a large positive number and can select the minimum. If the center is below the selected minimum, `g - min_val` wraps and can select the maximum. Both outcomes choose the farther endpoint, contrary to the filter's definition.

<details>
<summary>Reproduction / regression example (not run on CUDA)</summary>

```python
import cupy as cp
from cucim.skimage.filters import rank

for dtype in (cp.uint32, cp.uint64):
    image = cp.asarray(
        [[0, 10, 20], [30, 255, 40], [50, 60, 70]], dtype=dtype
    )
    footprint = cp.ones((3, 3), dtype=bool)
    result = rank.enhance_contrast_percentile(
        image,
        footprint,
        p0=0.1,
        p1=0.9,
        backend="elementwise",
        cast_to_uint8=False,
    )
    assert result[1, 1].item() == 70
```

At the center, the selected samples are `0, 10, ..., 70`, so the endpoints are 0 and 70. The center value is 255 and should map to 70. In uint32 arithmetic, `70 - 255` becomes `4,294,967,111`; the current comparison instead selects 0. The uint64 path has the analogous wraparound.

</details>

<details>
<summary>Proposed fix and regression coverage</summary>

Handle centers outside the selected interval before subtracting. Short-circuit evaluation then reaches the distance comparison only when both unsigned differences are nonnegative and representable. Preserve the existing lower-endpoint tie break.

Apply this to all three generated comparisons. Avoid fixing it merely by casting the original values to `double`, which can lose integer precision for uint64 values above `2**53`.

```diff
diff --git a/python/cucim/src/cucim/skimage/filters/rank/_rank_filter.py b/python/cucim/src/cucim/skimage/filters/rank/_rank_filter.py
--- a/python/cucim/src/cucim/skimage/filters/rank/_rank_filter.py
+++ b/python/cucim/src/cucim/skimage/filters/rank/_rank_filter.py
@@ -99,5 +99,6 @@
                 X g = x[i];
-                if (max_val - g < g - min_val) {
+                if (g >= max_val ||
+                    (g > min_val && max_val - g < g - min_val)) {
                     y = cast<Y>(max_val);
                 } else {
                     y = cast<Y>(min_val);
@@ -582,7 +583,8 @@
                 X max_val = values[actual_end - 1];
                 X g = x[i];
                 // Replace with whichever extreme is closer
-                if (max_val - g < g - min_val) {{
+                if (g >= max_val ||
+                    (g > min_val && max_val - g < g - min_val)) {{
                     y = cast<Y>(max_val);
                 }} else {{
                     y = cast<Y>(min_val);
@@ -593,7 +595,8 @@
                 X min_val = values[{idx_start}];
                 X max_val = values[{idx_end - 1}];
                 X g = x[i];
-                if (max_val - g < g - min_val) {{
+                if (g >= max_val ||
+                    (g > min_val && max_val - g < g - min_val)) {{
                     y = cast<Y>(max_val);
                 }} else {{
                     y = cast<Y>(min_val);
```

#### Regression coverage

- Cover centers above the selected maximum and below the selected minimum for both uint32 and uint64.
- Exercise masked and unmasked percentile kernels, plus generic contrast with a center-excluding footprint or mask.
- Check centers inside the interval, exact endpoints, equal endpoints, and midpoint ties.
- Include uint64 values above `2**53`, not only small values stored in uint64 arrays.
- Retain the uint8/uint16 and floating-point regression cases.

</details>

python/cucim/src/cucim/skimage/filters/rank/cuda/histogram_rank.cu
```diff
@@ -0,0 +1,755 @@
...
+        __syncthreads();
+    }
+}
+
+__device__ void histogramWeightedPrefixScan256(int* hist, int* scan)
+{
+    int tx = threadIdx.x;
+    if (tx < 256)
+    {
+        scan[tx] = hist[tx] * tx;
```
gigony: ### [P2] Weighted histogram sums overflow before wider output conversion

**AI-assisted review comment (finding 5/5).** Based on static analysis; the CUDA reproduction below has not been run.

This concerns internal weighted-sum overflow, not the documented overflow from choosing a narrow output dtype. The demonstrated large footprint passes histogram compatibility checks, including with a float64 output.

The histogram population can fit safely in int32 while the weighted sum does not. Widening the output only converts an already-overflowed intermediate.

For a constant uint8 image containing 255 and a 3001x3001 footprint:

| Quantity | Value |
| --- | --- |
| Neighborhood population | `3001 * 3001 = 9,006,001` |
| Exact weighted sum | `9,006,001 * 255 = 2,296,530,255` |
| Maximum signed int32 | `2,147,483,647` |
| Two's-complement int32 wraparound value | `-1,998,437,041` |
| Correct mean | `255` |

The exact sum exceeds the storage range even though the population does not. Signed overflow is not valid arithmetic; the wraparound value illustrates the resulting corruption rather than an API guarantee.

This example is eligible for the histogram backend: a 1536x1536 image is 2-D uint8, the full footprint is odd-sized, its radius 1500 does not exceed either image extent, and a float64 output is supported. With no mask or shifts, automatic selection also chooses the histogram path for this large footprint.

Affected weighted-result operations include `mean_percentile`, `sum_percentile`, `subtract_mean_percentile`, `mean_bilateral`, and `sum_bilateral`. Switching to a wider `out` does not fix their internal arithmetic.

<details>
<summary>Reproduction / regression example (not run on CUDA)</summary>

This is a deliberately large integration case. Run it on a suitably provisioned GPU, separately from fast unit tests if necessary.

```python
import cupy as cp
from cucim.skimage.filters import rank

image = cp.full((1536, 1536), 255, dtype=cp.uint8)
footprint = cp.ones((3001, 3001), dtype=bool)
out = cp.empty(image.shape, dtype=cp.float64)

result = rank.mean_bilateral(
    image,
    footprint,
    s0=10,
    s1=10,
    out=out,
    backend="histogram",
)
assert result is out
cp.testing.assert_array_equal(result, cp.full_like(out, 255))
```

Every included sample is 255, so the expected result is independent of boundary handling. The current weighted path cannot represent the required intermediate sum.

</details>

<details>
<summary>Proposed fix and regression coverage</summary>

Keep population counts and bin indices as `int`, but widen the entire weighted-sum path to `long long`:

1. Promote before multiplication, not after an int32 product has overflowed.
2. Use a 64-bit weighted scan buffer and scan temporary.
3. Widen the shared range sums and percentile endpoint reconstruction.
4. Keep the selected sum wide until the final output conversion.

A separate buffer avoids changing `tmp1`, which other operation specializations use for integer bins, extrema, and mode selection. The buffer is declared only for the weighted-operation specializations. It adds 2048 bytes of shared scan storage, plus the larger endpoint scalars; benchmark the affected specializations after integration.

```diff
diff --git a/python/cucim/src/cucim/skimage/filters/rank/cuda/histogram_rank.cu b/python/cucim/src/cucim/skimage/filters/rank/cuda/histogram_rank.cu
--- a/python/cucim/src/cucim/skimage/filters/rank/cuda/histogram_rank.cu
+++ b/python/cucim/src/cucim/skimage/filters/rank/cuda/histogram_rank.cu
@@ -328,15 +328,15 @@
-__device__ void histogramWeightedPrefixScan256(int* hist, int* scan)
+__device__ void histogramWeightedPrefixScan256(int* hist, long long* scan)
 {
     int tx = threadIdx.x;
     if (tx < 256)
     {
-        scan[tx] = hist[tx] * tx;
+        scan[tx] = static_cast<long long>(hist[tx]) * tx;
     }
     __syncthreads();

     for (int offset = 1; offset < 256; offset <<= 1)
     {
-        int v = 0;
+        long long v = 0;
         if (tx >= offset && tx < 256)
         {
             v = scan[tx - offset];
@@ -371,8 +371,9 @@
     // clang-format off
 #if RANK_HIST_OP == OP_MEAN || RANK_HIST_OP == OP_SUM || RANK_HIST_OP == OP_SUBTRACT_MEAN || RANK_HIST_OP == OP_BILATERAL_MEAN || RANK_HIST_OP == OP_BILATERAL_POP || RANK_HIST_OP == OP_BILATERAL_SUM
     // clang-format on
-    __shared__ int range_start_sum;
-    __shared__ int range_end_sum;
+    __shared__ long long weighted_scan[256];
+    __shared__ long long range_start_sum;
+    __shared__ long long range_end_sum;
 #endif

 #if RANK_HIST_OP == OP_ENTROPY
@@ -492,7 +493,7 @@
     // clang-format off
 #if RANK_HIST_OP == OP_MEAN || RANK_HIST_OP == OP_SUM || RANK_HIST_OP == OP_SUBTRACT_MEAN || RANK_HIST_OP == OP_BILATERAL_MEAN || RANK_HIST_OP == OP_BILATERAL_POP || RANK_HIST_OP == OP_BILATERAL_SUM
     // clang-format on
-    histogramWeightedPrefixScan256(hist, tmp1);
+    histogramWeightedPrefixScan256(hist, weighted_scan);
     if (tx == 0)
     {
         range_start_sum = 0;
@@ -516,8 +517,8 @@
         {
             range_start = (start_bin > 0) ? scan[start_bin - 1] : 0;
             range_end = scan[stop_bin - 1];
-            range_start_sum = (start_bin > 0) ? tmp1[start_bin - 1] : 0;
-            range_end_sum = tmp1[stop_bin - 1];
+            range_start_sum = (start_bin > 0) ? weighted_scan[start_bin - 1] : 0;
+            range_end_sum = weighted_scan[stop_bin - 1];
         }
     }
     __syncthreads();
@@ -526,23 +527,23 @@
     {
         int bin_end = scan[tx];
         int bin_start = bin_end - hist[tx];
-        int weighted_end = tmp1[tx];
-        int weighted_start = weighted_end - hist[tx] * tx;
+        long long weighted_end = weighted_scan[tx];
+        long long weighted_start = weighted_end - static_cast<long long>(hist[tx]) * tx;

         if (range_start > 0 && bin_start < range_start && bin_end >= range_start)
         {
-            range_start_sum = weighted_start + (range_start - bin_start) * tx;
+            range_start_sum = weighted_start + static_cast<long long>(range_start - bin_start) * tx;
         }
         if (range_end > 0 && bin_start < range_end && bin_end >= range_end)
         {
-            range_end_sum = weighted_start + (range_end - bin_start) * tx;
+            range_end_sum = weighted_start + static_cast<long long>(range_end - bin_start) * tx;
         }
     }
     __syncthreads();
 #        endif

     int selected_count_total = range_end - range_start;
-    int selected_sum_total = range_end_sum - range_start_sum;
+    long long selected_sum_total = range_end_sum - range_start_sum;
     if (op == OP_BILATERAL_POP)
     {
         return static_cast<RANK_HIST_OUTPUT_T>(selected_count_total);
```

Merely changing `selected_sum_total` is insufficient: the products, scan, and range endpoints would already have overflowed. Likewise, promoting the product to int64 while storing it back in an int32 scan does not resolve the problem.

#### Regression coverage

- Add a focused CUDA helper test using synthetic histogram counts just below and above `INT32_MAX // 255`; this can test the overflow boundary without constructing a huge image.
- Retain at least one public-API integration case demonstrating that a constant image remains constant under the large-window bilateral mean.
- Cover weighted percentile endpoints, including partial-bin reconstruction and both range-start/range-end sums.
- Exercise float32 and float64 outputs and retain existing uint8/uint16 output behavior, including documented narrow-output overflow.
- Compare forced histogram and automatic dispatch where applicable. Do not depend on a huge elementwise sort as the reference for the large-window case.
- Re-run existing non-weighted histogram tests to verify the shared-buffer change does not affect percentile, modal, entropy, equalize, or extrema operations.

</details>

gigony (commented):
Thank you @grlee77 for the change!

Sharing five AI-assisted inline review comments based on the prepared review report. Each comment includes the finding, proposed fix, and reproduction/regression-test guidance.

The affected rank-filter code is unchanged from the report's reviewed commit (`eb2b908`); these comments are anchored to the current PR head (`db4875c`). The proposed patches were checked with `git apply --check`, and host-side logic checks were performed. CUDA tests have not been run, and the suggested fixes have not been applied.

benchmarks/skimage/run-nv-bench-filters-rank.sh
grlee77: Thanks, not sure when pre-commit got uninstalled but I installed it and have now resolved via the automatic reformatting

python/cucim/src/cucim/skimage/filters/rank/_rank_filter.py
grlee77: Good catch, addressed in [5475b66](https://github.com/rapidsai/cucim/pull/1115/commits/5475b663f37ac00029ce9a9252acc6a818977e4b)

python/cucim/src/cucim/skimage/filters/rank/cuda/histogram_rank.cu
grlee77: Thanks updated in [a7f5624](https://github.com/rapidsai/cucim/pull/1115/commits/a7f5624c12c9727424192286730065fdc76bad1a) to promote to int64 when needed but retain int32 for smaller footprints where the issue cannot occur

python/cucim/src/cucim/skimage/filters/rank/_rank_filter.py
grlee77: addressed in [a76eef9](https://github.com/rapidsai/cucim/pull/1115/commits/a76eef9f7dcc39dbcab7ce47f0b4ac2d46e99fd2)

python/cucim/src/cucim/skimage/filters/rank/_rank_filter.py
grlee77: resolved by [d875949](https://github.com/rapidsai/cucim/pull/1115/commits/d87594904a8da3ef06a827bc1b3b9d6da30bb93d)

python/cucim/src/cucim/skimage/filters/rank/_generic.py
grlee77: Checked against scikit-image and updated for consistency in [7af9a10](https://github.com/rapidsai/cucim/pull/1115/commits/7af9a109e2729b429a6403cab354d362ad6fc219)

grlee77:
Thanks, @gigony and @jakirkham, all comments should now be addressed

python/cucim/src/cucim/skimage/filters/rank/_rank_filter.py
```diff
@@ -0,0 +1,1141 @@
...
+                    y = cast<Y>(x[i]);
+                    return;
+                }}
+                X g = x[i];
+                X clamped = (g < min_val) ? min_val :
+                    ((g > max_val) ? max_val : g);
+                double delta = static_cast<double>(max_val - min_val);
+                if (delta > 0) {{
+                    double scaled = (static_cast<double>(clamped - min_val)
+                                     / delta) * static_cast<double>({dtype_max});
```
gigony: ### [P2] Signed extrema differences overflow before wider output conversion

**AI-assisted review comment (round 2, finding 1/1).** Based on static analysis and host-side UBSan checks of generated arithmetic; the CuPy/CUDA reproduction has not been run.

With native int32/int64 inputs (`cast_to_uint8=False`), these subtractions occur in `X` before the conversion to double or the output dtype. For a neighborhood containing `INT32_MIN`, `0`, and `INT32_MAX`, the gradient is mathematically `4294967295`, but the signed input type cannot represent the intermediate difference. Autolevel and the guarded contrast-distance comparison have the same problem, including their sorted masked/unmasked variants. A wider `out` cannot repair overflow that already occurred.

This is separate from the fixed unsigned/outside-interval finding: the center here is inside the interval. It was newly identified in round two, but already existed before the review-response commits.

<details>
<summary>Affected locations, reasoning, reproduction, and validation limits</summary>

### Related locations

All paths below are under `python/cucim/src/cucim/skimage/filters/rank/`:

- `_rank_filter.py:85-122`: streaming gradient, contrast, and autolevel.
- `_rank_filter.py:536-549`: percentile-range gradient.
- `_rank_filter.py:577-602`: percentile-range contrast.
- `_rank_filter.py:707-738`: percentile-range autolevel.


### Root cause and impact

The subtraction is evaluated in `X` before `cast<Y>` or `static_cast<double>` widens it. For native int32/int64 images, an ordered pair of valid intensities can have a positive difference larger than the signed input maximum.

For a neighborhood containing `INT32_MIN`, `0`, and `INT32_MAX`:

| Operation at the center | Correct result with suitable output | Problematic intermediate |
| --- | --- | --- |
| Gradient with float64 output | `4294967295.0` | `INT32_MAX - INT32_MIN` cannot fit in int32 |
| Autolevel with float64 output | `2147483648 / 4294967295`, approximately `0.5000000001164153` | Both the range and center-minus-minimum are evaluated as signed int32 first |
| Enhance contrast with int32 output | `INT32_MAX`, which is one unit closer to zero | `0 - INT32_MIN` overflows in the distance comparison |

This is different from the resolved R1 unsigned issue: the center is **inside** the interval, so the new endpoint guards correctly allow the comparison to proceed but cannot make a signed difference representable. It also affects the masked and unmasked sorted-neighborhood variants.

Providing a float64 or wider integer output does not prevent the intermediate overflow. The native signed-dtype documentation discusses output scaling but does not make these arithmetic errors intentional.

### Reproduction for a CUDA environment

The following expresses the expected behavior. It was not run with CuPy here:

```python
import cupy as cp
from cucim.skimage.filters import rank

image = cp.asarray([[-2**31, 0, 2**31 - 1]], dtype=cp.int32)
footprint = cp.ones((1, 3), dtype=bool)
kwargs = dict(backend="elementwise", cast_to_uint8=False)

gradient = rank.gradient(
    image, footprint, out=cp.empty(image.shape, dtype=cp.float64), **kwargs
)
autolevel = rank.autolevel(
    image, footprint, out=cp.empty(image.shape, dtype=cp.float64), **kwargs
)
contrast = rank.enhance_contrast(image, footprint, **kwargs)

assert gradient[0, 1].item() == 4294967295.0
assert abs(autolevel[0, 1].item() - 2147483648 / 4294967295) < 1e-15
assert contrast[0, 1].item() == 2**31 - 1
```

### Validation performed

Extracted the actual kernel-generator functions from the reviewed Git object, captured their generated pre/update/post strings, and compiled the arithmetic in a host C++ harness under UBSan. Checked the streaming, unmasked sorted, and masked sorted paths for all three operations.

UBSan reported signed integer overflow in the current generated arithmetic. The observed host results were `-1` for gradient, `0` for autolevel, and `INT32_MIN` for contrast in all three paths. These are host observations of undefined signed arithmetic, not a guarantee of the exact values every CUDA compiler will produce.

The proposed generated arithmetic below passed the same nine checks without UBSan diagnostics. Additional assertions covered an int64 interval whose difference is `UINT64_MAX`, int64 contrast at zero, and small uint64 differences above `2**53`.

Neighborhood gathering, actual GPU sorting, CuPy casting implementation, and CUDA kernel execution were not exercised by this host harness.

</details>

<details>
<summary>Proposed fix and regression coverage</summary>

Use an ordered-difference helper. For integer `high >= low`, subtract after conversion to the corresponding unsigned type; the mathematical difference fits that unsigned width, including the full signed int64 range. For floating-point inputs, widen before subtraction. Preserve the existing outside-interval guards and lower-endpoint tie behavior.

Do not simply cast every input value to double before integer subtraction: that would lose small uint64 differences above `2**53` and regress the precision cases already covered by the PR.

Apply the helper to gradient, autolevel, and the guarded contrast distances in every generated variant:

```diff
diff --git a/python/cucim/src/cucim/skimage/filters/rank/_rank_filter.py b/python/cucim/src/cucim/skimage/filters/rank/_rank_filter.py
--- a/python/cucim/src/cucim/skimage/filters/rank/_rank_filter.py
+++ b/python/cucim/src/cucim/skimage/filters/rank/_rank_filter.py
@@ -28,6 +28,24 @@
     _should_use_rank_histogram,
 )
 
+
+_RANK_DIFFERENCE_PREAMBLE = r"""
+template <typename T, bool = std::is_floating_point<T>::value>
+struct RankOrderedDifference {
+    typedef typename std::make_unsigned<T>::type type;
+    __device__ static type get(T high, T low) {
+        return static_cast<type>(
+            static_cast<type>(high) - static_cast<type>(low));
+    }
+};
+
+template <typename T>
+struct RankOrderedDifference<T, true> {
+    __device__ static double get(T high, T low) {
+        return static_cast<double>(high) - static_cast<double>(low);
+    }
+};
+"""
 
 def _get_streaming_rank_kernel(
     p0,
@@ -88,7 +106,7 @@
                     y = cast<Y>(x[i]);
                     return;
                 }
-                y = cast<Y>(max_val - min_val);
+                y = cast<Y>(RankOrderedDifference<X>::get(max_val, min_val));
             """
         elif operation == "enhance_contrast":
             post = """
@@ -98,7 +116,7 @@
                 }
                 X g = x[i];
                 if (g >= max_val ||
-                    (g > min_val && max_val - g < g - min_val)) {
+                    (g > min_val && RankOrderedDifference<X>::get(max_val, g) < RankOrderedDifference<X>::get(g, min_val))) {
                     y = cast<Y>(max_val);
                 } else {
                     y = cast<Y>(min_val);
@@ -113,9 +131,9 @@
                 X g = x[i];
                 X clamped = (g < min_val) ? min_val :
                     ((g > max_val) ? max_val : g);
-                double delta = static_cast<double>(max_val - min_val);
+                double delta = static_cast<double>(RankOrderedDifference<X>::get(max_val, min_val));
                 if (delta > 0) {{
-                    double scaled = (static_cast<double>(clamped - min_val)
+                    double scaled = (static_cast<double>(RankOrderedDifference<X>::get(clamped, min_val))
                                      / delta) * static_cast<double>({dtype_max});
                     y = cast<Y>(scaled);
                 }} else {{
@@ -285,9 +303,9 @@
 
     op_name = operation.replace("_", "")
     mask_str = "_masked" if has_mask else ""
-    preamble = ""
+    preamble = _RANK_DIFFERENCE_PREAMBLE
     if operation == "noise_filter":
-        preamble = r"""
+        preamble += r"""
 template <typename T, bool = std::is_floating_point<T>::value>
 struct RankNoiseDistance {
     typedef typename std::make_unsigned<T>::type type;
@@ -539,13 +557,13 @@
             post += """
                 X min_val = values[actual_start];
                 X max_val = values[actual_end - 1];
-                y = cast<Y>(max_val - min_val);
+                y = cast<Y>(RankOrderedDifference<X>::get(max_val, min_val));
             """
         else:
             post += f"""
                 X min_val = values[{idx_start}];
                 X max_val = values[{idx_end - 1}];
-                y = cast<Y>(max_val - min_val);
+                y = cast<Y>(RankOrderedDifference<X>::get(max_val, min_val));
             """
     elif operation == "subtract_mean":
         # Subtract mean: scikit-image formula:
@@ -583,7 +601,7 @@
                 X g = x[i];
                 // Replace with whichever extreme is closer
                 if (g >= max_val ||
-                    (g > min_val && max_val - g < g - min_val)) {{
+                    (g > min_val && RankOrderedDifference<X>::get(max_val, g) < RankOrderedDifference<X>::get(g, min_val))) {{
                     y = cast<Y>(max_val);
                 }} else {{
                     y = cast<Y>(min_val);
@@ -595,7 +613,7 @@
                 X max_val = values[{idx_end - 1}];
                 X g = x[i];
                 if (g >= max_val ||
-                    (g > min_val && max_val - g < g - min_val)) {{
+                    (g > min_val && RankOrderedDifference<X>::get(max_val, g) < RankOrderedDifference<X>::get(g, min_val))) {{
                     y = cast<Y>(max_val);
                 }} else {{
                     y = cast<Y>(min_val);
@@ -716,9 +734,9 @@
                 X g = x[i];
                 X clamped = (g < min_val) ? min_val : \
 ((g > max_val) ? max_val : g);
-                double delta = static_cast<double>(max_val - min_val);
+                double delta = static_cast<double>(RankOrderedDifference<X>::get(max_val, min_val));
                 if (delta > 0) {{
-                    double scaled = (static_cast<double>(clamped - min_val) \
+                    double scaled = (static_cast<double>(RankOrderedDifference<X>::get(clamped, min_val)) \
 / delta) * static_cast<double>({dtype_max});
                     y = cast<Y>(scaled);
                 }} else {{
@@ -732,9 +750,9 @@
                 X g = x[i];
                 X clamped = (g < min_val) ? min_val : \
 ((g > max_val) ? max_val : g);
-                double delta = static_cast<double>(max_val - min_val);
+                double delta = static_cast<double>(RankOrderedDifference<X>::get(max_val, min_val));
                 if (delta > 0) {{
-                    double scaled = (static_cast<double>(clamped - min_val) \
+                    double scaled = (static_cast<double>(RankOrderedDifference<X>::get(clamped, min_val)) \
 / delta) * static_cast<double>({dtype_max});
                     y = cast<Y>(scaled);
                 }} else {{
@@ -865,7 +883,7 @@
         cval,
         has_weights=has_weights,
         has_mask=has_mask,
-        preamble=sorter,
+        preamble=sorter + _RANK_DIFFERENCE_PREAMBLE,
     )
 
 
```

The patch is a proposal checked with `git apply --check` and host-generated arithmetic tests; it was not applied. Formatting, NVRTC compilation, and GPU performance/correctness checks remain necessary before integration.

### Regression coverage

- int32/int64 neighborhoods spanning negative and positive extrema.
- Wider output arrays for gradient and autolevel; contrast must select the nearer endpoint without changing the output dtype.
- Streaming, sorted unmasked, and sorted masked paths; use a non-full interval such as `p0=0.01, p1=1` to exercise sorted range code while including all three samples.
- Existing uint32/uint64 outside-interval cases, large uint64 values, exact endpoints, equal endpoints, and ties.
- Float16/float32/float64 specializations and existing noise-filter preamble generation.

</details>

gigony (commented):
Thank you @grlee77 for addressing the earlier comments!

**AI-assisted round-two review of `a7f5624c12c9727424192286730065fdc76bad1a`.** The five round-one findings are addressed in source, including the documented dimension-dependent shift convention and singleton-NaN policy. One additional native signed-integer arithmetic finding is posted inline.

The description, prior report, current diff, and existing discussions were checked before posting. The new finding is not a regression introduced by the response commits.

**Validation:** During round-two report preparation, the generated arithmetic for streaming, unmasked sorted, and masked sorted gradient/autolevel/contrast paths was exercised in a host C++ UBSan harness. The current code reported signed overflow; the proposed arithmetic passed the nine host checks plus signed/unsigned boundary assertions. The patch passed `git apply --check` without being applied. Actual GPU sorting, neighborhood gathering, CuPy casts, NVRTC compilation, and the full CUDA/project test suite were not run.

Submitted as `COMMENT`, not approval or a formal request for changes. No implementation edits or fixes were applied.

<details>
<summary>Round-one finding dispositions</summary>

| R1 finding | Status | Current evidence |
| --- | --- | --- |
| 1. NaNs can prevent population-loop progress | Resolved in source | `_rank_filter.py:644-676` consumes the first group member unconditionally in both variants. `_percentile.py` and the module overview document singleton NaN groups. New masked/unmasked NaN and finite-group regression tests are present. |
| 2. Nonzero Z shift changes X/Y axes | Resolved with a revised documented convention | `_generic.py:146-165` chooses the mapping based on dimensionality, not whether Z is zero. Two-dimensional mapping is Y/X; three-dimensional and higher mapping is X/Y/Z. Documentation and 2-D/3-D/4-D tests reflect that deliberate scikit-image-compatible choice. The R1 proposed mapping is therefore superseded, not still required. |
| 3. Empty masked populations copy intensities | Resolved in source | Streaming counts return their zero counters at `_rank_filter.py:169-174,257-260`; masked percentile populations explicitly return zero at `468-474`. The brute-force reference and empty/sparse-mask tests were corrected. |
| 4. Native unsigned contrast wraps outside the interval | Resolved for the reported unsigned case | All three comparisons now guard outside-interval centers before subtraction (`100-105,585-602`). Added tests cover uint32/uint64, masks, endpoint/tie behavior, and values above `2**53`. |
| 5. Weighted histogram sums overflow int32 | Resolved in source | `_histogram.py:169-176` selects int64 when `area * 255` exceeds int32. The CUDA product, scan storage, scan temporary, range endpoints, and final selected sum all use `RANK_HIST_WEIGHTED_SUM_T`. New tests exercise the production helper with a large logical window. |

These dispositions are based on code review and the host checks described below. The new GPU tests were not executed in this environment.

</details>

python/cucim/src/cucim/skimage/filters/rank/_rank_filter.py
grlee77: Thanks, should now be resolved as suggested

python/cucim/src/cucim/skimage/filters/rank/_rank_filter.py
```diff
@@ -0,0 +1,1185 @@
...
+            double bilat_sum = 0.0;
+            X g = x[i];
+            double gd = static_cast<double>(g);
+        """
+        update = f"""
+            X v = {{value}};
+            double vd = static_cast<double>(v);
+            if (gd > (vd - {s0}) && gd < (vd + {s1})) {{
+                bilat_pop++;
+                bilat_sum += vd;
```
gigony, line 274: ### [P2] Bilateral range checks exclude equal high-magnitude neighbors

**AI-assisted review comment (round 3, finding 1/1).** Based on static analysis and host execution of generated arithmetic. The CuPy/CUDA reproduction below has not been run.

With native int64/uint64 inputs or sufficiently large floating-point values (`cast_to_uint8=False`), a small bilateral tolerance can disappear when added to or subtracted from the absolute grayvalue. At `g = v = 2**60` and the default `s0=s1=10`, both `vd-10` and `vd+10` round to `vd`, so this test excludes even a neighbor equal to the center. The generated host code consequently returns zero population, mean, and sum for three equal samples, although all expected results fit uint64. Both masked and unmasked paths are affected.

Please compare native integer differences, or widened floating-point differences, with the tolerance bounds while preserving strict endpoints. This issue is separate from the now-fixed signed-extrema overflow and was already present before `96608bb`.

<details>
<summary>Root cause, observed host results, and CUDA regression example</summary>

### Root cause and impact

The documented interval is `g - s1 < value < g + s0`. With default `s0=s1=10`, a neighbor equal to the center must be included.

For `g = value = 2**60`, both values themselves are exactly representable as doubles, but adding or subtracting 10 rounds back to the same double. The predicate therefore becomes `g > g && g < g`, which is false. A constant image is mapped to zero by the mean filter, and its bilateral population and sum are also zero.

A fresh check of the current generated code produced the following at the center of a three-sample uint64 neighborhood, both without a mask and with an all-true mask:

| Output | Expected | Observed host result |
| --- | --- | --- |
| Population | `3` | `0` |
| Mean | `1152921504606846976` | `0` |
| Sum | `3458764513820540928` | `0` |

All three expected results fit the default uint64 output. This is neighbor-selection failure, not output overflow and not undefined signed arithmetic.

Simply changing the expression to `double(value) - double(center)` would still lose small differences between large integers. For example, samples at `2**60 - 11`, `2**60`, and `2**60 + 11` must yield population 1 with tolerance 10, rather than counting all three after their values round together.

### CUDA reproduction / regression example

This example was not run with CuPy here:

```python
import cupy as cp
from cucim.skimage.filters import rank

base = 2**60
image = cp.full((1, 3), base, dtype=cp.uint64)
footprint = cp.ones((1, 3), dtype=bool)

for mask in (None, cp.ones_like(image, dtype=bool)):
    kwargs = dict(
        mask=mask,
        s0=10,
        s1=10,
        backend="elementwise",
        cast_to_uint8=False,
    )
    assert rank.pop_bilateral(image, footprint, **kwargs)[0, 1].item() == 3
    assert rank.mean_bilateral(image, footprint, **kwargs)[0, 1].item() == base
    assert rank.sum_bilateral(image, footprint, **kwargs)[0, 1].item() == 3 * base
```

The public wrappers permit native dtypes when conversion is disabled. The histogram backend only accepts uint8, so these native examples reach the shared streaming bilateral predicate.

</details>

<details>
<summary>Proposed bilateral-only patch, validation, and regression coverage</summary>

Reuse the newly committed ordered-difference helper for integer distances and compare those distances with safely translated tolerance bounds. For floating-point inputs, form the signed difference before comparing it with `s0` and `-s1`.

The patch below addresses only the remaining bilateral issue. It does **not** repeat the now-implemented signed-extrema fix. Strict endpoints, asymmetric intervals, and zero-width behavior are preserved. The bound guards avoid out-of-range floating-to-unsigned casts, including when rounding upward exceeds a narrow integer type's maximum.

```diff
diff --git a/python/cucim/src/cucim/skimage/filters/rank/_rank_filter.py b/python/cucim/src/cucim/skimage/filters/rank/_rank_filter.py
--- a/python/cucim/src/cucim/skimage/filters/rank/_rank_filter.py
+++ b/python/cucim/src/cucim/skimage/filters/rank/_rank_filter.py
@@ -51,6 +51,41 @@
     "enhance_contrast",
     "gradient",
 }
+
+_RANK_BILATERAL_RANGE_PREAMBLE = r"""
+template <typename T, bool = std::is_floating_point<T>::value>
+struct RankBilateralRange {
+    typedef typename RankOrderedDifference<T>::type type;
+    __device__ __forceinline__ static bool below(type difference, double upper) {
+        if (!(upper > 0)) return false;
+        double bound = ceil(upper);
+        return bound >= ldexp(1.0, sizeof(type) * 8) ||
+            difference < static_cast<type>(bound);
+    }
+    __device__ __forceinline__ static bool above(type difference, double lower) {
+        if (lower < 0) return true;
+        if (!(lower < ldexp(1.0, sizeof(type) * 8))) return false;
+        return difference > static_cast<type>(floor(lower));
+    }
+    __device__ __forceinline__ static bool contains(T value, T center, double s0, double s1) {
+        if (value >= center) {
+            type difference = RankOrderedDifference<T>::get(value, center);
+            return below(difference, s0) && above(difference, -s1);
+        }
+        type difference = RankOrderedDifference<T>::get(center, value);
+        return above(difference, -s0) && below(difference, s1);
+    }
+};
+
+template <typename T>
+struct RankBilateralRange<T, true> {
+    __device__ __forceinline__ static bool contains(T value, T center, double s0, double s1) {
+        double difference = static_cast<double>(value) -
+                            static_cast<double>(center);
+        return difference < s0 && difference > -s1;
+    }
+};
+"""
 
 
 def _get_streaming_rank_kernel(
@@ -264,12 +299,11 @@
             int bilat_pop = 0;
             double bilat_sum = 0.0;
             X g = x[i];
-            double gd = static_cast<double>(g);
         """
         update = f"""
             X v = {{value}};
             double vd = static_cast<double>(v);
-            if (gd > (vd - {s0}) && gd < (vd + {s1})) {{
+            if (RankBilateralRange<X>::contains(v, g, {s0}, {s1})) {{
                 bilat_pop++;
                 bilat_sum += vd;
             }}
@@ -319,6 +353,11 @@
         if operation in _RANK_ORDERED_DIFFERENCE_OPS
         else ""
     )
+    if operation in ("bilateral_mean", "bilateral_pop", "bilateral_sum"):
+        preamble = (
+            _RANK_ORDERED_DIFFERENCE_PREAMBLE
+            + _RANK_BILATERAL_RANGE_PREAMBLE
+        )
     if operation == "noise_filter":
         preamble += r"""
 template <typename T, bool = std::is_floating_point<T>::value>
```

This proposal passed `git apply --check` against `96608bb…`, without application. Its modified Python source parsed successfully. The proposed generated code returned the correct population, mean, and sum in all six constant-neighborhood cases, while retaining the 49 passing arithmetic/noise assertions.

Additional proposed-code checks covered center-only masks, just-outside neighbors, negative int64 centers, float32/float64 centers, and **102,528 exact integer range comparisons** across signed/unsigned 8-, 16-, 32-, and 64-bit types. No UBSan diagnostics were reported.

This remains a proposal requiring formatting, NVRTC compilation, GPU regression testing, and performance evaluation. It does not alter the existing double accumulation for means/sums or claim bit-exact reductions for arbitrary 64-bit inputs.

### Regression coverage to add

- Constant high-magnitude int64/uint64, float32, and float64 inputs with default positive tolerances.
- Negative signed centers, all-true masks, center-only masks, and empty population masks.
- Neighbors just inside, exactly on, and just outside each endpoint at a large common offset.
- Asymmetric and zero-width intervals, while preserving the existing accepted parameter behavior.
- Low-valued uint8/uint16 controls and backend agreement where histogram dispatch is eligible.

</details>

**Validation limits:** The host harness supplied neighborhood values and mask/index scaffolding. It did not execute actual CuPy array traversal, CUDA kernel launch, NVRTC, GPU sorting, or the full rank/median suites. The proposed patch has not been applied; GPU validation and performance evaluation remain necessary.

gigony (commented):
Thank you @grlee77 for the signed-extrema fix!

**AI-assisted round-three review of `96608bb032c7659cf958b5dc86c0249579085d38`.** The round-two signed-extrema finding is addressed in source, and the five round-one fixes remain in place. One separate P2 bilateral-neighbor selection finding is posted inline; it is not a regression introduced by the latest commit.

The PR description, current diff, updated round-three report, and existing discussions were checked before publication. The current head and inline anchor were revalidated.

**Validation during report preparation:** The current generated arithmetic, using its actual preambles, shell-sort implementation, and casting helper bodies in a host C++ UBSan harness, passed 49 signed/unsigned/floating/noise assertions. Separate checks reproduced zero population/mean/sum for equal high-magnitude bilateral samples in masked and unmasked paths. The proposed bilateral-only patch passed `git apply --check` without application; its host-generated arithmetic passed the targeted examples and 102,528 exact integer range comparisons. These are host-side checks, not CUDA runtime results.

CuPy/CUDA tests, NVRTC compilation, real GPU traversal/sorting, and the full project suites were not run. The checked-in float16 tests were reviewed but not executed against CuPy's CUDA half type.

Submitted as a non-blocking `COMMENT` review, not an approval or formal request for changes. No implementation changes were made and no proposed fixes were applied.

gigony (approved):
Thank you @grlee77 for the change!

Aside from one remaining AI comment (P2), it looks good to me.
Approving in advance.
Thank you!

grlee77:
@msarahan, the `pre-commit` config was updated to list the new files under the appropriate copyright sections. Please check and approve that change

msarahan (approved):

python/cucim/src/cucim/skimage/filters/rank/cuda/histogram_rank.cu
```diff
@@ -0,0 +1,759 @@
...
+#    define RANK_HIST_WEIGHTED_SUM_T int
+#endif
+
+#if RANK_HIST_OP == OP_GEOMETRIC_MEAN
+__device__ __constant__ double geometricMeanLogLut[256] = { 0,
+                                                            0.69314718055994529,
+                                                            1.0986122886681098,
+                                                            1.3862943611198906,
+                                                            1.6094379124341003,
+                                                            1.791759469228055,
```
jakirkham, line 41: Would add a code comment explaining what this is and how it was generated. This may be important if it needs to be updated in the future

python/cucim/src/cucim/skimage/filters/rank/cuda/histogram_rank.cu
```diff
@@ -0,0 +1,759 @@
...
+                                                 double dtype_max,
+                                                 unsigned char center)
+{
+    int tx = threadIdx.x;
+    __shared__ int result;
+    __shared__ int range_start;
+    __shared__ int range_end;
+    // clang-format off
+#if RANK_HIST_OP == OP_MEAN || RANK_HIST_OP == OP_SUM || RANK_HIST_OP == OP_SUBTRACT_MEAN || RANK_HIST_OP == OP_BILATERAL_MEAN || RANK_HIST_OP == OP_BILATERAL_POP || RANK_HIST_OP == OP_BILATERAL_SUM
+    // clang-format on
```
jakirkham, line 377: Could you please share a bit more about why `clang-format` doesn't work here?

python/cucim/src/cucim/skimage/filters/rank/cuda/histogram_rank.cu
```diff
@@ -0,0 +1,759 @@
...
+    __shared__ RANK_HIST_WEIGHTED_SUM_T range_start_sum;
+    __shared__ RANK_HIST_WEIGHTED_SUM_T range_end_sum;
+#endif
+
+#if RANK_HIST_OP == OP_ENTROPY
+    double ent = 0.0;
+    if (tx < 256 && hist[tx] > 0)
+    {
+        double p = ((double)hist[tx]) / window_size;
+        ent = -p * log(p) / 0.6931471805599453;
```
jakirkham, line 387: Guessing the constant here is `ln(2)`. As `log` is `ln`, then this is to convert to a base 2 logarithm

However CUDA already has [`log2`]( https://docs.nvidia.com/cuda/cuda-math-api/cuda_math_api/group__CUDA__MATH__DOUBLE.html#_CPPv44log2d ), which is a base 2 logarithm. Could we just use that directly (and avoid some floating point imprecision)?

PR #1117, opened 2026-07-09
fix(deps): preload `cusolver` if `cupy` is installed

gforsyth:
I'm testing out smaller base images for wheel testing in #1116 and those tests flagged up that we're missing a `cusolver` dependency on wheel-only installs (the previous testing images have a full cuda toolkit installed beforehand).

The missing dependency turns out to be due to a `cupy` bug upstream with wheel-only installs (reported in https://github.com/cupy/cupy/issues/10095).

This is a temporary workaround (if we want one) to enable switching to smaller test images (and also allowing end-users to use `cucim` on wheel-only installs)

xref https://github.com/rapidsai/build-planning/issues/143
xref https://github.com/rapidsai/ci-imgs/pull/400
xref https://github.com/rapidsai/shared-workflows/pull/521

### DIFF
--- a/dependencies.yaml
+++ b/dependencies.yaml
@@ -316,10 +316,6 @@ dependencies:
         packages:
           - libnvimgcodec>=0.8.0,<0.9.0
           - nvimgcodec>=0.8.0,<0.9.0
-      - output_types: [requirements, pyproject]
-        packages:
-          - cuda-toolkit[cusolver]
-          - cuda-pathfinder
     specific:
       - output_types: [requirements, pyproject]
         matrices:
--- a/dependencies.yaml
+++ b/dependencies.yaml
@@ -316,6 +316,10 @@ dependencies:
         packages:
           - libnvimgcodec>=0.8.0,<0.9.0
           - nvimgcodec>=0.8.0,<0.9.0
+      - output_types: [requirements, pyproject]
+        packages:
+          - cuda-toolkit[cusolver]
+          - cuda-pathfinder
     specific:
       - output_types: [requirements, pyproject]
         matrices:
--- a/python/cucim/src/cucim/__init__.py
+++ b/python/cucim/src/cucim/__init__.py
@@ -1,5 +1,5 @@
 #
-# SPDX-FileCopyrightText: Copyright (c) 2020-2021, NVIDIA CORPORATION.
+# SPDX-FileCopyrightText: Copyright (c) 2020-2026, NVIDIA CORPORATION.
 # SPDX-License-Identifier: Apache-2.0
 #
 
@@ -39,6 +39,20 @@ try:
     submodules += ["core", "skimage"]
 except ImportError:
     pass
+
+if _is_cupy_available:
+    # If CuPy is available AND we aren't on a system CTK install, then
+    # `cuda.pathfinder` will also be available
+    # Use it to pre-load `cusolver` to get around an upstream issue in CuPy
+    try:
+        from cuda.pathfinder import (
+            DynamicLibNotFoundError,
+            load_nvidia_dynamic_lib,
+        )
+
+        load_nvidia_dynamic_lib("cusolver")
+    except (ImportError, DynamicLibNotFoundError):
+        pass
 
 try:
     from .clara import CuImage, cli
--- a/python/cucim/src/cucim/__init__.py
+++ b/python/cucim/src/cucim/__init__.py
@@ -45,7 +45,8 @@ if _is_cupy_available:
     # If CuPy is available AND we aren't on a system CTK install, then
     # the end-user will have to have used the `ctk` extra to install `cupy-cuda13x` (or similar)
     # `cuda.pathfinder` will also be available
-    # Use it to pre-load `cusolver` to get around an upstream issue in CuPy
+    # Use it to pre-load `cusolver` to get around an upstream issue in CuPy:
+    # https://github.com/cupy/cupy/issues/10095
     #
     # If CuPy is available and we ARE on a system CTK install, then a user might have not installed the `ctk` extra
     # and so `cuda.pathfinder` might not be available. In this scenario, we don't NEED `cuda.pathfinder` since on CTK installs
--- a/python/cucim/src/cucim/__init__.py
+++ b/python/cucim/src/cucim/__init__.py
@@ -56,13 +56,8 @@ if _is_cupy_available:
     # `ctk` extra, all we do is pre-load the DSO that would've been loaded
     # anyway
     try:
-        from cuda.pathfinder import (
-            DynamicLibNotFoundError,
-            load_nvidia_dynamic_lib,
-        )
-
-        load_nvidia_dynamic_lib("cusolver")
-    except (ImportError, DynamicLibNotFoundError):
+        from cupy_backends.cuda.libs import cusolver
+    except ImportError:
         pass
 
 try:
--- a/python/cucim/src/cucim/__init__.py
+++ b/python/cucim/src/cucim/__init__.py
@@ -1,5 +1,5 @@
 #
-# SPDX-FileCopyrightText: Copyright (c) 2020-2026, NVIDIA CORPORATION.
+# SPDX-FileCopyrightText: Copyright (c) 2020-2021, NVIDIA CORPORATION.
 # SPDX-License-Identifier: Apache-2.0
 #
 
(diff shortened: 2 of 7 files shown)


dependencies.yaml
```diff
@@ -316,6 +316,10 @@ dependencies:
         packages:
           - libnvimgcodec>=0.8.0,<0.9.0
           - nvimgcodec>=0.8.0,<0.9.0
+      - output_types: [requirements, pyproject]
+        packages:
+          - cuda-toolkit[cusolver]
+          - cuda-pathfinder
```
jameslamb: As written, this would allow `cucim-cu13` to be installed with `cuda-toolkit==12.*`. I don't think we want that.. We won't catch that in CI because we `==` pin `cuda-toolkit` as an additional  constraint:

https://github.com/rapidsai/cucim/blob/5c68f29c15d74c1fff234aa6bbeb024d292bc66c/dependencies.yaml#L185-L204

... but that's a CI only thing.

We should ensure the wheel metadata provides a correct solve for users (CUDA 12 for `-cu12`, CUDA 13 for `-cu13`).

For this I think we want:

1. matrices by CUDA major version, with the dependency spelled `cuda-toolkit[cusolver]==12.*` and `cuda-toolkit[cusolver]==13.*`
2. `usa_cuda_wheels` guards, similar to https://github.com/rapidsai/cudf/blob/dda07069af24236c612dec623f471e946a6cdcfb/dependencies.yaml#L943-L962

The `usa_cuda_wheels` guards aren't *strictly* necessary since this project doesn't have devcontainers and isn't part of DLFW builds, but having the spelling be consistent with the rest of RAPIDS would make all-of-RAPIDS updates easier.

jameslamb (changes requested):
Left a blocking suggestion.

If we end up using Python code to pre-load `cusolver`  after you consider this feedback then I think that's ok and I think where you chose is the right place.

But some other thoughts...

1. if some DSO in `cucim` needs `libcusolver.so.*` at runtime, why isn't `auditwheel` trying to vendor it? Or IS it making a copy? Worth checking, because I don't see it excluded here and would expect to:

https://github.com/rapidsai/cucim/blob/5c68f29c15d74c1fff234aa6bbeb024d292bc66c/ci/build_wheel.sh#L67

2. I don't think we need a `cuda-pathfinder` dependency. Could you look into just adding RPATH entries into `site-packages/`, like we do for RAFT's `cusolver` dependency?

https://github.com/NVIDIA/raft/blob/99ca82537c2263249018f6bfb53de3d31442f8ca/python/libraft/CMakeLists.txt#L40-L51

Normally when I jump into one of these `packaging-codeowners` reviewers where I wasn't assigned I remove the auto-assigned person, but leaving @KyleFromNVIDIA here because I think his thoughts on this topic would also be really helpful and possibly different from mine.

gforsyth:
> 1. if some DSO in `cucim` needs `libcusolver.so.*` at runtime, why isn't `auditwheel` trying to vendor it? Or IS it making a copy? Worth checking, because I don't see it excluded here and would expect to:

This is a good question -- I took another look at the failures and the tracebacks a little more closely and this is looking less like a `cucim` problem and like maybe more of a `cupy` problem.

From this run on #1116: https://github.com/rapidsai/cucim/actions/runs/29044082057/job/86214270820?pr=1116

```python

________ test_custom_eigvals_kernels_vs_linalg_eigvalsh[float32-shape1] ________
[gw4] linux -- Python 3.12.13 /pyenv/versions/3.12.13/bin/python
python/cucim/src/cucim/skimage/feature/tests/test_corner.py:351: in test_custom_eigvals_kernels_vs_linalg_eigvalsh
    evs1 = _reference_eigvals_computation(H)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
python/cucim/src/cucim/skimage/feature/tests/test_corner.py:338: in _reference_eigvals_computation
    eigs = cp.linalg.eigvalsh(matrices)[..., ::-1]
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
/pyenv/versions/3.12.13/lib/python3.12/site-packages/cupy/linalg/_eigenvalue.py:343: in eigvalsh
    import cupyx.cusolver
E   ImportError: libcusolver.so.12: cannot open shared object file: No such file or directory

```

gforsyth:
> like maybe more of a `cupy` problem.

or alternatively, not a problem, just a mismatch.  We have all of this wheel loading logic that prefers wheel installations over system installations, but because there CAN be a system installation, we don't fail loudly when we're missing DSOs in wheels.

I wonder if all that's needed for `cucim` is adding `cusolver` to the test environment?

gforsyth:
> I wonder if all that's needed for `cucim` is adding `cusolver` to the test environment?

or potentially even simpler, should we be installing the `cupy-cuda??x[ctk]` extra?

gforsyth:
> or potentially even simpler, should we be installing the `cupy-cuda??x[ctk]` extra?

nope, we are already installing the `ctk` extra in the test runs

gforsyth:
> nope, we are already installing the `ctk` extra in the test runs

Ok, this is a cupy bug, reporting upstream, then we can decide what to do here

gforsyth:
Reported upstream in https://github.com/cupy/cupy/issues/10095, trying out temporary fixes involving `cuda.pathfinder` in #1116 (TLDR: I think we can assume `cuda.pathfinder` is installed without adding a new dependency here, because `cupy` ships it, so IF `cupy` is installed, we can attempt to preload `libcusolver` using `pathfinder` to avoid the load error)

gforsyth:
Thanks for the careful review @jameslamb !  I wouldn't have thought I'd end up at `cupy` bug when I first saw the tracebacks in #1116

gforsyth:
Tested these changes here (obviously) and also in https://github.com/rapidsai/cucim/pull/1116 to demonstrate that `libcusolver` gets correctly loaded from the wheel (since that's the only option in that image)

python/cucim/src/cucim/__init__.py
```diff
@@ -40,6 +41,29 @@
 except ImportError:
     pass
 
+if _is_cupy_available:
+    # If CuPy is available AND we aren't on a system CTK install, then
+    # the end-user will have to have used the `ctk` extra to install `cupy-cuda13x` (or similar)
+    # `cuda.pathfinder` will also be available
+    # Use it to pre-load `cusolver` to get around an upstream issue in CuPy
```
bdice: Please link the upstream issue:
```suggestion
    # Use it to pre-load `cusolver` to get around an upstream issue in CuPy:
    # https://github.com/cupy/cupy/issues/10095
```

bdice (approved):

jameslamb (approved):
Thanks so much for looking into those things I linked, I support the implementation you ended up with.

I read (and subscribed to) the `cupy` issue, great report! Thanks for doing that.

It's a nice side benefit of using a slimmer test image, that we can catch issues like this in CI 🙂

python/cucim/src/cucim/__init__.py
jakirkham: Thanks Bradley! 🙏

Agreed. Am going to take the liberty of committing this change to help move this along.

python/cucim/src/cucim/__init__.py
```diff
@@ -40,6 +41,30 @@
...
+    # `ctk` extra, all we do is pre-load the DSO that would've been loaded
+    # anyway
+    try:
+        from cuda.pathfinder import (
+            DynamicLibNotFoundError,
+            load_nvidia_dynamic_lib,
+        )
+
+        load_nvidia_dynamic_lib("cusolver")
+    except (ImportError, DynamicLibNotFoundError):
```
jakirkham: @seberg made the following simplified workaround upstream: https://github.com/cupy/cupy/issues/10095#issuecomment-4969116038

Adding it as a suggestion below:

```suggestion
        from cupy_backends.cuda.libs import cusolver
    except ImportError:
```

jakirkham: Gil already signed off on this in the same upstream issue: https://github.com/cupy/cupy/issues/10095#issuecomment-4970176744

So going ahead and applying it.

jakirkham (approved):
Thanks Gil! 🙏

Have gone ahead and applied reasonable suggestions from here and the upstream issue. Hope that is ok.

LGTM

PR #1122, opened 2026-07-22
cuslide2: add pyramidal OME-TIFF semantics and nvImageCodec 0.9 readiness

cdinea:
## What this PR changes

- Add decode correctness updates for multiplex OME-TIFF (dynamic GRAY/RGB and UINT8/UINT16 selection)
- Add dedicated OME-XML parser module (`ome_xml.h/.cpp`)
- Add multi-file OME fileset handling using `TiffData` companion-file references
- Use OME channel names and physical spacing metadata where available
- Relax RGB-only optimization guards in TIFF/IFD read paths
- Add OME parser unit tests
- Update nvImageCodec integration for 0.9.0 compatibility (version header + SONAME discovery order)
- Report TIFF resolution from tags 282/283/296 instead of hardcoded defaults, including RATIONAL decoding
- Preserve pyramid sub-resolution levels on OME files whose `TiffData` only references full-resolution planes
- Return an all-background raster for regions that do not intersect the image, matching `cuslide`
- Stop destroying the nvImageCodec decoder at process exit, which intermittently aborted in `free()`

## Scope notes

- Explicit multiplex selector API expansion is intentionally out of scope for this PR.

## Validation

- Built `cucim.kit.cuslide2` successfully in the prepared build environment
- Added OME parser tests and integrated into `cuslide2_tests`
- `pre-commit run --hook-stage manual --all-files` passes
- Unit tests and the pyramidal OME-TIFF script pass as shown below

### Running the pyramidal OME-TIFF script

```bash
# Build the plugin
export CUDACXX=$CONDA_PREFIX/pkgs/cuda-toolkit/bin/nvcc
./run build_local all release $CONDA_PREFIX
python -m pip install --editable python/cucim

# Run the script against an OME-TIFF
export ENABLE_CUSLIDE2=1
python scripts/test_pyramidal_ome_tiff.py <path-to>/retina_large.ome.tiff --no-cache
```

The script needs no GPU-specific flags: it decodes each region on both CPU and
GPU and compares the results, so a GPU is required for the comparison checks to
run rather than be skipped. Pass `--no-cache` to exercise the direct decode path
instead of the tile cache.

If you rebuild the C++ but see unchanged behavior under `pytest`, note that the
Python package loads the plugin from `python/cucim/src/cucim/clara/`, not from
`cpp/plugins/cucim.kit.cuslide2/build-release/lib/`. Re-running
`./run build_local` refreshes that copy.

### Expected output, multi-plane OME-TIFF

`retina_large.ome.tiff`, 2 resolution levels, `SizeC=2`, `SizeZ=64`, uint16:

```
🧪 Synthetic uint16 decode checks
  ✅ Metadata dtype is uint16 (bits=16, typestr=<u2)
  ✅ Level-0 uint16 patch matches source ((10, 10), dtype=uint16)
  ✅ High-bit values preserved (max=12345)
  ✅ Level-1 uint16 plane matches source ((64, 64), dtype=uint16)
  ✅ Synthetic uint16 decode path validated
======================================================================
🔬 Testing pyramidal OME-TIFF with cuslide2
======================================================================
📊 Image summary
  Shape: [1567, 2048, 2]
  Dtype: <cucim.clara.DLDataType code:DLUInt(1) bits:16 lanes:1>
  Levels: 2
  Level 0: 2048x1567 (downsample: 1.000x)
  Level 1: 256x195 (downsample: 8.018x)
  ✅ Input reported as uint16 — decode checks will enforce dtype
🧬 OME metadata checks
  ✅ Found OME metadata block
  size_c=2, size_z=64, size_t=1
  channel_names_count=2
📐 Pyramid geometry checks
  ✅ Pyramid dims are non-increasing and match reported downsamples (2 distinct resolution(s) over 2 level(s))
🧪 Multi-level decode checks
  ✅ GPU and CPU level 0 decode are identical
  ✅ GPU and CPU level 1 decode are identical
🎯 Plane selection checks (C/Z/T kwargs)
    ✅ GPU and CPU plane decode are identical      # C=0,Z=0,T=0
    ✅ GPU and CPU plane decode are identical      # C=1,Z=0,T=0
🔄 Batch decode validation
  ✅ Batch decode GPU and CPU are identical
✅ Pyramidal OME-TIFF test completed
```

Two details worth checking in that output. `Levels: 2` confirms the pyramid
sub-resolution survives: OME `TiffData` entries reference only the
full-resolution planes, so mapping planes alone collapses the list to level 0 and
makes level 1 unreachable through `read_region()`. And the run must end at
`✅ Pyramidal OME-TIFF test completed` with exit status 0; a trailing
`free(): invalid pointer` / `Aborted` was the exit-time teardown crash fixed here,
which reproduced on roughly one run in five.

### Expected output, non-OME pyramidal TIFF

The same script accepts ordinary WSIs. Against `CMU-1.tif` the OME and plane
checks report as skipped, which is correct for a file with no OME-XML:

```
  Levels: 7
  Level 0: 46592x33792 (downsample: 1.000x)
  Level 6: 624x432 (downsample: 76.444x)
🧬 OME metadata checks
  ⚠️  OME metadata block not found in img.metadata
  ✅ Pyramid dims are non-increasing and match reported downsamples (7 distinct resolution(s) over 7 level(s))
  ✅ GPU and CPU level 0 decode are identical
  ✅ GPU and CPU level 6 decode are identical
⚠️  Skipping C/Z/T plane checks (size_c not available)
  ✅ Batch decode GPU and CPU are identical
✅ Pyramidal OME-TIFF test completed
```

Note that per-axis downsample ratios differ on real WSIs, so the geometry check
compares against the mean of the axis ratios rather than assuming a single factor.

### Unit tests

```bash
cd python/cucim
ENABLE_CUSLIDE2=1 python -m pytest tests/unit/clara/test_tiff_read_region.py -q
```

```
..........................                                               [100%]
26 passed
```

All 26 pass with `cuslide2` enabled and with the default plugin. The three
`test_tiff_stripe_outside` cases previously failed under `cuslide2` with
`Failed to decode IFD[0] with nvImageCodec. ROI: (-48,-40) 16x16`: nvImageCodec
fills out-of-bounds pixels for a partially overlapping region, which is why
boundary reads passed, but rejects a region that does not intersect the image at
all.


VERSION
```diff
@@ -1 +1 @@
-26.08.00
+26.10.00
```
KyleFromNVIDIA: Please rebase your PR to avoid changes to `VERSION`, `dependencies.yaml`, and `.github/workflows/*.yaml` that update the version. This branch is targeting 26.08 and should not be merged with anything from `main`.

KyleFromNVIDIA (changes requested):

KyleFromNVIDIA (approved):

cpp/plugins/cucim.kit.cuslide2/src/cuslide/nvimgcodec/nvimgcodec_tiff_parser.cpp
```diff
@@ -351,7 +275,7 @@ NvImageCodecTiffParserManager::NvImageCodecTiffParserManager()
         exec_params.struct_next = nullptr;
         exec_params.device_allocator = nullptr;
         exec_params.pinned_allocator = nullptr;
-        exec_params.max_num_cpu_threads = compute_max_decoder_threads();
+        exec_params.max_num_cpu_threads = 0;
```
gigony: I'm wondering why we're setting this to 0 now.

cpp/plugins/cucim.kit.cuslide2/src/cuslide/nvimgcodec/nvimgcodec_tiff_parser.cpp
```diff
@@ -631,9 +589,11 @@ bool TiffFileParser::parse_tiff_structure()
...
         extract_tiff_tags(ifd_info);
 
-        // codec_name returns "tiff" (container format), not compression type.
-        // Individual TIFF tags are queried via NVIMGCODEC_METADATA_KIND_TIFF_TAG,
-        // and TIFF_TAG_LIST enumeration is available in v0.8.0+.
+        // Current limitation (nvImageCodec v0.6.0):
+        // - codec_name returns "tiff" (container format) not compression type
+        // - Individual TIFF tags not exposed through metadata API
+        // - Only vendor-specific metadata blobs available (MED_APERIO, MED_PHILIPS, etc.)
+        //
```
gigony: I am wondering if nvImageCodec v0.9 still has these limitation.

cpp/plugins/cucim.kit.cuslide2/src/cuslide/nvimgcodec/nvimgcodec_tiff_parser.cpp
```diff
@@ -958,6 +931,13 @@ void TiffFileParser::extract_ifd_metadata(IfdInfo& ifd_info)
         }
     }
 
+    // WORKAROUND for nvImageCodec 0.6.0: Philips TIFF metadata limitation
+    // ========================================================================
+    // nvImageCodec 0.6.0 does NOT expose:
+    // 1. Individual TIFF tags (SOFTWARE, ImageDescription, etc.)
+    // 2. Philips format detection for some files
+    //
+
```
gigony: Is this still specific to nvImageCodec v0.9.0, or does the code also work with nvImageCodec v0.6.0?

cpp/plugins/cucim.kit.cuslide2/src/cuslide/tiff/types.h
```diff
@@ -1,5 +1,5 @@
 /*
- * SPDX-FileCopyrightText: Copyright (c) 2020-2026, NVIDIA CORPORATION.
+ * SPDX-FileCopyrightText: Copyright (c) 2020-2021, NVIDIA CORPORATION.
```
gigony:
```suggestion
 * SPDX-FileCopyrightText: Copyright (c) 2020-2026, NVIDIA CORPORATION.
```

gigony (approved):
Thanks @cdinea for adding ome-tiff support!
Overall, it looks good to me.

cpp/cmake/deps/nvimgcodec.cmake
```diff
@@ -6,6 +6,22 @@
 #
 
 if (NOT TARGET deps::nvimgcodec)
+    set(NVIMGCODEC_SONAME_CANDIDATES
+        "libnvimgcodec.so.0.9.0"
+        "libnvimgcodec.so.0.8.0"
```
mkepa-nv: In the conda yaml file you set requirements >= 0.9 for nvimagecodec. Why do we need 0.8 here then?

cpp/plugins/cucim.kit.cuslide2/cmake/deps/nvimgcodec.cmake
```diff
@@ -9,6 +9,22 @@ if (NOT TARGET deps::nvimgcodec)
     set(NVIMGCODEC_LIB_PATH "")
     set(NVIMGCODEC_INCLUDE_PATH "")
 
+    set(NVIMGCODEC_SONAME_CANDIDATES
+        "libnvimgcodec.so.0.9.0"
+        "libnvimgcodec.so.0.8.0"
```
mkepa-nv: Similar question as above. Also maybe the cmake structure can be simplified so nvimagecodec search is done in a single place? But this is a question for a future pull request

cpp/plugins/cucim.kit.cuslide2/src/cuslide/nvimgcodec/nvimgcodec_decoder.cpp
```diff
@@ -285,7 +340,7 @@ bool decode_ifd_region_nvimgcodec(const IfdInfo& ifd_info,
 
         nvimgcodecCodeStreamView_t view{};
         view.struct_type = NVIMGCODEC_STRUCTURE_TYPE_CODE_STREAM_VIEW;
-        view.struct_size = sizeof(nvimgcodecCodeStreamView_t);
+        view.struct_size = offsetof(nvimgcodecCodeStreamView_t, limit_images);
```
mkepa-nv: This doesn't look right. Why would we ask for offset of struct member instead of the whole struct size? And also `limit_images` was removed in 0.9 so I have no idea how is this working right now

cpp/plugins/cucim.kit.cuslide2/src/cuslide/nvimgcodec/nvimgcodec_decoder.cpp
```diff
@@ -331,24 +386,24 @@ bool decode_ifd_region_nvimgcodec(const IfdInfo& ifd_info,
...
-        output_image_info.color_spec = NVIMGCODEC_COLORSPEC_SRGB;
+        output_image_info.color_spec = decode_spec.color_spec;
         output_image_info.chroma_subsampling = NVIMGCODEC_SAMPLING_NONE;
         output_image_info.num_planes = 1;
         output_image_info.buffer_kind = buffer_kind;
 
-        uint32_t num_channels = ifd_info.num_channels > 0 ? ifd_info.num_channels : 3;
-        size_t row_stride = width * num_channels;
+        uint32_t num_channels = decode_spec.num_channels;
+        size_t row_stride = width * num_channels * decode_spec.bytes_per_sample;
```
mkepa-nv: This can overflow for large `width`, I would static cast `widt` to size_t if you store it in that type anyway

cpp/plugins/cucim.kit.cuslide2/src/cuslide/nvimgcodec/nvimgcodec_decoder.cpp
```diff
@@ -623,9 +678,9 @@ std::vector<BatchDecodeResult> decode_batch_regions_nvimgcodec(
             }
 
             const auto& region = regions[i];
-            // Use num_channels from IFD info, fallback to 3 (RGB)
-            uint32_t num_channels = (ifd_info.num_channels > 0) ? ifd_info.num_channels : 3;
-            size_t row_stride = region.width * num_channels;
+            const DecodePixelSpec decode_spec = resolve_decode_pixel_spec(ifd_info);
+            uint32_t num_channels = decode_spec.num_channels;
+            size_t row_stride = region.width * num_channels * decode_spec.bytes_per_sample;
```
mkepa-nv: The same overflow can happen here

cpp/plugins/cucim.kit.cuslide2/src/cuslide/nvimgcodec/nvimgcodec_tiff_parser.cpp
```diff
@@ -54,7 +52,7 @@ namespace cuslide2::nvimgcodec
 // nvImageCodec extension path auto-detection
 // ============================================================================
 //
-// Workaround for a known nvImageCodec pre-0.8.0 bug: when the library is installed
+// Workaround for older nvImageCodec extension-path behavior: when the library is installed
```
mkepa-nv, line 56: With this pull request you support only nvimagecodec 0.9, right? Is this workaround still needed? If yes please update the comment

cpp/plugins/cucim.kit.cuslide2/src/cuslide/nvimgcodec/nvimgcodec_tiff_parser.cpp
```diff
@@ -66,8 +64,8 @@ namespace cuslide2::nvimgcodec
 //      NvimgcodecLoadSymbol() + dladdr() (not the stub address)
 //   3. Probes <lib_dir>/extensions/ and <lib_dir>/../extensions/
 //
-// NOTE: nvImageCodec 0.8.0 fixes the extension search path for conda installs.
-// This workaround is kept as a safety net for non-standard install layouts.
+// Keep this path probe in place for compatibility across mixed environments,
+// including older deployments where extension auto-discovery is incomplete.
```
mkepa-nv, line 69: But the old have different API (e.g. limit_images in old view). And I didn't see any compile time checks for nvimagecodec macros, so I assume that the old nvimagecodec will no longer compile. I would either verify that it works with the old nvimagecodec, or drop the old code

cpp/plugins/cucim.kit.cuslide2/src/cuslide/nvimgcodec/nvimgcodec_tiff_parser.cpp
```diff
@@ -360,21 +284,32 @@ NvImageCodecTiffParserManager::NvImageCodecTiffParserManager()
...
-            instance_ = nullptr;
-            status_message_ = fmt::format("Failed to create decoder for metadata extraction (status: {})",
-                                         static_cast<int>(status));
-            #ifdef DEBUG
-            fmt::print("⚠️  {}\n", status_message_);
-            #endif // DEBUG
-            return;
+            // If CUDA initialization is unavailable (for example, no GPU device
+            // in the current runtime), retry with CPU-only decoder so metadata
+            // extraction and CPU decode paths can still run.
```
mkepa-nv, line 393: What kind of metadata extraction do you want? In the cpu only env getting metadata connected to an image (size, datatype, color spec, etc) should work. But I don't think getting tiff tags would work, as this is done with nvtiff, which I assume need some cuda libs. But I will check with rest of the team and come back.

cpp/plugins/cucim.kit.cuslide2/src/nvimgcodec_dynlink/nvimgcodec_version.h
mkepa-nv, line 1: You should update to the full new headers, using old headers can easily cause a nontrivial bug.

cpp/plugins/cucim.kit.cuslide2/src/cuslide/nvimgcodec/nvimgcodec_tiff_parser.cpp
mkepa-nv: We do support generic tags in the API, here is python example: https://docs.nvidia.com/cuda/nvimagecodec/samples/metadata.html#Generic-tiff-tag-reading
iirc this was already supported in 0.8? Lev can know more details on this and how to querry the tags

cpp/plugins/cucim.kit.cuslide2/src/cuslide/nvimgcodec/nvimgcodec_tiff_parser.cpp
```diff
@@ -750,6 +710,19 @@ bool TiffFileParser::parse_tiff_structure()
         #endif // DEBUG
     }
 
+    if (ifd_infos_.empty())
```
mkepa-nv, line 812: There is already error message in the `else` above - I would move the details from this error message there. I don't see a reason for having two similar error messages when the ifd_infos is empty. And having more details when only some ifd infos are missing would be great too

cpp/plugins/cucim.kit.cuslide2/src/cuslide/nvimgcodec/nvimgcodec_tiff_parser.cpp
```diff
@@ -1005,7 +985,7 @@ void TiffFileParser::extract_tiff_tags(IfdInfo& ifd_info)
     }
 
     // ========================================================================
-    // Direct TIFF Tag Retrieval by ID (nvImageCodec 0.7.0+)
+    // nvImageCodec 0.7.0+: Direct TIFF Tag Retrieval by ID
```
mkepa-nv: I would remove mentions for the old nvimagecodec (as the API is not compatible) and just keep the code for the newest version

cpp/plugins/cucim.kit.cuslide2/src/cuslide/tiff/ifd.cpp
```diff
@@ -191,11 +200,27 @@ IFD::IFD(TIFF* tiff, uint16_t index, const ::cuslide2::nvimgcodec::IfdInfo& ifd_
...
+            }
+        }
     }
 
     // Set format defaults
     planar_config_ = cuslide::tiff::PLANARCONFIG_CONTIG;  // nvImageCodec outputs interleaved
-    photometric_ = cuslide::tiff::PHOTOMETRIC_RGB;
+    if (photometric_ == 0)
+    {
+        photometric_ = (samples_per_pixel_ == 1) ? cuslide::tiff::PHOTOMETRIC_MINISBLACK
```
mkepa-nv, line 271: I would change the check to `samples_per_pixel_ <= 2` - in case there is grayscale with alpha image (this is a rather rare case, but it doesn't cost us anything to implement

cpp/plugins/cucim.kit.cuslide2/src/cuslide/tiff/ifd.cpp
```diff
@@ -272,10 +298,11 @@ bool IFD::read([[maybe_unused]] const TIFF* tiff,
     const uint64_t location_len = request->location_len;
 
     uint32_t batch_size = request->batch_size;
-    int32_t n_ch = samples_per_pixel_;
+    int32_t n_ch = static_cast<int32_t>(std::max<uint32_t>(1, samples_per_pixel_));
```
mkepa-nv, line 346: can samples_per_pixel ever be zero?

cpp/plugins/cucim.kit.cuslide2/src/cuslide/tiff/ome_xml.cpp
```diff
@@ -0,0 +1,171 @@
...
+    }
+
+    Model model;
+    model.image_id = image_node.attribute("ID").as_string("");
+    model.image_name = image_node.attribute("Name").as_string("");
+
+    Pixels pixels;
+    pixels.size_x = pixels_node.attribute("SizeX").as_uint(0);
+    pixels.size_y = pixels_node.attribute("SizeY").as_uint(0);
+    pixels.size_c = std::max<uint32_t>(1, pixels_node.attribute("SizeC").as_uint(1));
```
mkepa-nv, line 103: you already set the default value with `as_uint(1)`. is the max still needed? If the xml has zero as size, maybe it should be left at that value?

cpp/plugins/cucim.kit.cuslide2/src/cuslide/tiff/ome_xml.cpp
```diff
@@ -0,0 +1,171 @@
...
+            {
+                entry.uuid = uuid_node.child_value();
+                entry.file_name = uuid_node.attribute("FileName").as_string("");
+            }
+
+            pixels.tiff_data.emplace_back(std::move(entry));
+        }
+    }
+
+    if (pixels.size_x == 0 || pixels.size_y == 0)
```
mkepa-nv, line 155: I would move this check up, where you read the size_x and y

cpp/plugins/cucim.kit.cuslide2/src/cuslide/tiff/ome_xml.h
```diff
@@ -0,0 +1,64 @@
...
+
+struct Model
+{
+    bool valid = false;
+    Pixels pixels;
+    std::string image_id;
+    std::string image_name;
+};
+
+bool parse(const std::string& xml_text, Model* out_model, std::string* out_error = nullptr);
```
mkepa-nv, line 61: You always populate `out_model` - I would change its type from pointer to reference

cpp/plugins/cucim.kit.cuslide2/src/cuslide/tiff/tiff.cpp
```diff
@@ -82,44 +70,30 @@ static void parse_string_array(const char* values, json& arr, PhilipsMetadataTyp
...
-                        arr.emplace_back(std::stod(val));
-                        break;
-                    case PhilipsMetadataType::IUInt16:
-                    case PhilipsMetadataType::IUInt32:
-                    case PhilipsMetadataType::IUInt64:
-                        arr.emplace_back(std::stoul(val));
-                        break;
-                    }
-                }
-                catch (const std::exception&)
```
mkepa-nv, line 103: the exception will be propagated up when parsing fails. Is this expected? It will also mean that this function will return on the first error, rather than continuing and ignoring errors as it did before

cpp/plugins/cucim.kit.cuslide2/src/cuslide/tiff/tiff.cpp
```diff
@@ -249,20 +221,13 @@ static std::string strip_string(const std::string& str)
...
 }
 
-static constexpr size_t kMaxImageDescriptionBytes = 1 * 1024 * 1024; // 1 MB
-
 static void parse_aperio_svs_metadata(std::shared_ptr<IFD>& first_ifd, json& metadata)
 {
     (void)metadata;
     std::string& desc = first_ifd->image_description();
 
-    if (desc.size() > kMaxImageDescriptionBytes)
```
mkepa-nv, line 259: Why are you removing the limit now? And why was it needed before? Maybe you should just increase the limit?

cpp/plugins/cucim.kit.cuslide2/src/cuslide/tiff/tiff.cpp
```diff
@@ -379,6 +344,8 @@ TIFF::TIFF(const cucim::filesystem::Path& file_path) : file_path_(file_path)
 
     // Initialize metadata container
     metadata_ = new json{};
+    parser_file_hashes_.clear();
+    parser_file_hashes_.push_back(file_handle_shared_ ? file_handle_shared_->hash_value : 0);
```
mkepa-nv, line 357: can file_handle_shared_ ever be null? It looks like it is reset only if there is an error (and in that case, new exception is thrown after cleanup so this code won't be reached)

cpp/plugins/cucim.kit.cuslide2/src/cuslide/tiff/tiff.cpp
```diff
@@ -548,19 +633,20 @@ void TIFF::resolve_vendor_format()
     {
         bool is_aperio = false;
 
+        // Method 1: Check ImageDescription starts with "Aperio "
         auto& image_desc = first_ifd->image_description();
         std::string_view prefix("Aperio ");
-        if (image_desc.size() >= prefix.size() &&
-            std::mismatch(prefix.begin(), prefix.end(), image_desc.begin()).first == prefix.end())
+        auto res = std::mismatch(prefix.begin(), prefix.end(), image_desc.begin());
+        if (res.first == prefix.end())
```
mkepa-nv, line 649: NIT: it looks like the `res` is only used in this `if`. So maybe you can just call missmatch().first as you did before?

cpp/plugins/cucim.kit.cuslide2/src/cuslide/tiff/tiff.cpp
```diff
@@ -548,19 +633,20 @@ void TIFF::resolve_vendor_format()
...
             is_aperio = true;
         }
 
-        // Method 2: Check metadata_blobs for Aperio (MED_APERIO)
+        // Method 2: Check metadata_blobs for Aperio (kind=5 in v0.8.0)
         if (!is_aperio && nvimgcodec_parser_)
         {
             const auto& metadata_blobs = nvimgcodec_parser_->get_metadata_blobs(0);
-            if (metadata_blobs.find(kMetadataKindMedAperio) != metadata_blobs.end())
+            if (metadata_blobs.find(5) != metadata_blobs.end())  // MED_APERIO = 5
```
mkepa-nv, line 658: IMO the old named constant were more readable, what was the reason for removing them?

cpp/plugins/cucim.kit.cuslide2/src/cuslide/tiff/tiff.cpp
```diff
@@ -715,6 +805,208 @@ void TIFF::resolve_vendor_format()
...
+    json* json_metadata = reinterpret_cast<json*>(metadata);
+
+    ome::Model model;
+    std::string parse_error;
+    if (!ome::parse(first_ifd->image_description(), &model, &parse_error) || !model.valid)
+    {
+        return;
+    }
+
+    ome_size_c_ = std::max<int64_t>(1, model.pixels.size_c);
```
mkepa-nv, line 833: you already set the size to at least 1 in the ome::parse, there is no need to repeat it here

cpp/plugins/cucim.kit.cuslide2/src/cuslide/tiff/ome_xml.h
mkepa-nv, line 61: Also in the main lib you always pass `out_error`. So maybe it can be changed to reference as well? Then you won't need the checks for nullptr. The only place where nullptr is passed is in a single test - but imo you can pass there a valid string

cpp/plugins/cucim.kit.cuslide2/src/cuslide/tiff/tiff.cpp
```diff
@@ -715,6 +805,208 @@ void TIFF::resolve_vendor_format()
...
+        {
+            std::string channel_name = channel.name.empty() ? fmt::format("C{}", channel.index) : channel.name;
+            if (!ome_channel_name_to_index_.count(channel_name))
+            {
+                ome_channel_name_to_index_.emplace(channel_name, static_cast<int64_t>(ome_channel_names_.size()));
+            }
+            ome_channel_names_.emplace_back(std::move(channel_name));
+        }
+    }
+    else
```
mkepa-nv, line 852: Is this is a valid file if the channels are missing?

cpp/plugins/cucim.kit.cuslide2/src/cuslide/tiff/tiff.cpp
```diff
@@ -715,6 +805,208 @@ void TIFF::resolve_vendor_format()
...
+                return a.first > b.first;
+            }
+            return a.second > b.second;
+        });
+
+        auto level_for_ifd = [&level_dims, this](size_t ifd_idx) -> uint16_t {
+            const auto& ifd = ifds_[ifd_idx];
+            std::pair<uint32_t, uint32_t> dim = { ifd->width(), ifd->height() };
+            auto it = std::find(level_dims.begin(), level_dims.end(), dim);
+            return (it == level_dims.end()) ? 0 : static_cast<uint16_t>(std::distance(level_dims.begin(), it));
```
mkepa-nv, line 956: should you rather return some error if the dimension was not found or throw an exception? If it should always be present I would add an assert

cpp/plugins/cucim.kit.cuslide2/src/cuslide/tiff/tiff.cpp
```diff
@@ -715,6 +805,208 @@ void TIFF::resolve_vendor_format()
...
+            for (const auto& [key, ifd_idx] : ome_plane_to_ifd_)
+            {
+                uint16_t level = std::get<3>(key);
+                if (!level_representative_ifd.count(level))
+                {
+                    level_representative_ifd[level] = ifd_idx;
+                }
+            }
+        }
+        if (!level_representative_ifd.empty())
```
mkepa-nv, line 1009: this if will never execute, in the if above we will always add at least one element, because `has_ome_plane_index_ ` is true, so `ome_plane_to_ifd_` is not empty

cpp/plugins/cucim.kit.cuslide2/src/cuslide/tiff/tiff.cpp
```diff
@@ -715,6 +805,208 @@ void TIFF::resolve_vendor_format()
...
+    {
+        std::map<uint16_t, size_t> level_representative_ifd;
+        for (const auto& [key, ifd_idx] : ome_plane_to_ifd_)
+        {
+            if (std::get<0>(key) == 0 && std::get<1>(key) == 0 && std::get<2>(key) == 0)
+            {
+                level_representative_ifd[std::get<3>(key)] = ifd_idx;
+            }
+        }
+        if (level_representative_ifd.empty())
```
mkepa-nv, line 976: what if there are mixed ifds? some with c = z = t = 0 and some with other values? Then you will set representative only for the c = z = t = 0. Is this on purpose?

cpp/plugins/cucim.kit.cuslide2/src/cuslide/tiff/tiff.cpp
```diff
@@ -842,11 +1124,9 @@ void TIFF::_populate_philips_tiff_metadata(uint16_t ifd_count, void* metadata, s
                 double downsample = std::round((pixel_spacings[spacing_index].first / spacing_x_l0 +
                                                 pixel_spacings[spacing_index].second / spacing_y_l0) /
                                                2);
-                if (downsample > 0)
-                {
-                    ifd->width_ = width_l0 / downsample;
-                    ifd->height_ = height_l0 / downsample;
-                }
+                // Fix width and height of IFD
+                ifd->width_ = width_l0 / downsample;
```
mkepa-nv: what if downsample is 0?

cpp/plugins/cucim.kit.cuslide2/src/cuslide/tiff/tiff.cpp
```diff
@@ -885,23 +1165,27 @@ void TIFF::_populate_philips_tiff_metadata(uint16_t ifd_count, void* metadata, s
...
-                        // Find '>' that closes the opening <Attribute ...> tag
-                        while (data_ptr < desc_end && *data_ptr != '>')
+                        // `image_desc_cstr[node_offset]` would point to the following text:
+                        //   Attribute Element="0x1004" Group="0x301D" Name="PIM_DP_IMAGE_DATA" PMSVR="IString">
+                        //     (base64-encoded JPEG image)
+                        //   </Attribute>
+                        //
+
+                        // 34 is from `Attribute Name="PIM_DP_IMAGE_DATA"`
+                        char* data_ptr = const_cast<char*>(image_desc_cstr) + node_offset + 34;
```
mkepa-nv: What if this is outside of the `image_desc_cstr`?

cpp/plugins/cucim.kit.cuslide2/src/cuslide/cuslide.cpp
```diff
@@ -181,32 +185,109 @@ static bool CUCIM_ABI parser_parse(CuCIMFileHandle_ptr handle_ptr, cucim::io::fo
...
+        shape[2] = n_ch;
+    }
+
     std::pmr::vector<std::string_view> channel_names(&resource);
     channel_names.reserve(n_ch);
-    if (n_ch == 3)
+    auto add_owned_channel_name = [&resource, &channel_names](const std::string& name) {
+        void* raw = resource.allocate(name.size() + 1, alignof(char));
+        auto* buf = static_cast<char*>(raw);
+        memcpy(buf, name.c_str(), name.size() + 1);
```
mkepa-nv: I would just make the channel_names to store string instead of the string_view. That vector is the owner of the names, so this even sounds like a more correct approach.

scripts/test_pyramidal_ome_tiff.py
```diff
@@ -0,0 +1,471 @@
...
+        cpu_time = time.time() - start
+        cpu_np = _to_numpy(cpu_region)
+        print(f"  CPU level {level}: {cpu_np.shape}, {cpu_np.dtype}, {cpu_time:.4f}s")
+        try:
+            start = time.time()
+            gpu_region = img.read_region((0, 0), read_size, level=level, device="cuda")
+            gpu_time = time.time() - start
+            gpu_np = _to_numpy(gpu_region)
+            print(f"  GPU level {level}: {gpu_np.shape}, {gpu_np.dtype}, {gpu_time:.4f}s")
+            if gpu_np.shape != cpu_np.shape:
```
mkepa-nv, line 614: shape check is not enough. I would check the pixels too. And maybe you should check that each level have a correct dimensions (that they decrease with correct ratio for each level done) - otherwise you will not be sure that parsing is working as expected.

scripts/test_pyramidal_ome_tiff.py
```diff
@@ -0,0 +1,471 @@
...
+            max_diff = int(np.max(np.abs(gpu_np.astype(np.int64) - cpu_np.astype(np.int64))))
+            raise RuntimeError(
+                f"Pixel mismatch for plane C={c_index},Z={z_index},T={t_index}: max_diff={max_diff}"
+            )
+        print("    ✅ GPU and CPU plane decode are identical")
+    except Exception as e:
+        print(f"    ⚠️  GPU plane validation skipped/failed: {e}")
+
+
+def _validate_batch_decode(img, level_dims):
```
mkepa-nv, line 437: You added  a bunch of code for higher resolution decode (with dtype uint 16) - I would add test for such case as well

mkepa-nv (changes requested):

cpp/plugins/cucim.kit.cuslide2/src/cuslide/nvimgcodec/nvimgcodec_tiff_parser.cpp
LevKruglyak: Yeah this would recover the overcontention issues which were fixed in the last release.

cpp/plugins/cucim.kit.cuslide2/src/cuslide/nvimgcodec/nvimgcodec_tiff_parser.cpp
LevKruglyak: This has been supported since 0.8, but only with a GPU decoder.

cpp/plugins/cucim.kit.cuslide2/src/cuslide/nvimgcodec/nvimgcodec_tiff_parser.cpp
cdinea: Good catch — unintentional, and Lev is right that it reintroduces the over-contention regression.

`nvimgcodec.h` documents `max_num_cpu_threads = 0` as "default value equal to number of cpu cores", so this decoder was silently taking a full-core CPU threadpool. That was harmless when it was CPU-only and metadata-only, but it now backs the primary decode path (`NVIMGCODEC_DEVICE_CURRENT`, all backends), so with N dataloader workers you get N x cores threads.

The `0` is a leftover from the pre-promotion shape of this code; the rebase dropped `compute_max_decoder_threads()` along with it, which is the ~76-line shift in this hunk.

Fixed in 9bd44b8 — heuristic is back to `min(max(1, hardware_concurrency / 4), 8)`, still overridable via `CUCIM_MAX_DECODER_THREADS` (>0 = exact count, 0 = explicit opt-in to the nvImageCodec default).

cpp/plugins/cucim.kit.cuslide2/src/cuslide/nvimgcodec/nvimgcodec_tiff_parser.cpp
cdinea: You're right, and thanks @mkepa-nv for the pointer — the comment was stale. Individual TIFF tags are queried unconditionally via `NVIMGCODEC_METADATA_KIND_TIFF_TAG` in `extract_tiff_tags()`, which runs on the line directly above, so the claim contradicted the adjacent code.

Removed in 9bd44b8. I kept only the part that's still true: `codec_name` reports the container format rather than the compression type, so compression is inferred from the tags below.

cpp/plugins/cucim.kit.cuslide2/src/cuslide/nvimgcodec/nvimgcodec_tiff_parser.cpp
cdinea: That block was a stale v0.6.0 workaround note describing limitations that no longer apply — removed entirely in 9bd44b8. The surrounding code path is not v0.6.0-compatible anyway, since it depends on the 0.7.0+ TIFF tag metadata API.

cpp/cmake/deps/nvimgcodec.cmake
cdinea: You're right, that 0.8.0 line shouldn't be there — it's left over from when this branch still supported 0.8. Everything else already asks for 0.9: `dependencies.yaml`, the conda environment files, `conda_build_config.yaml` and `pyproject.toml` are all `>=0.9.0,<0.10.0`, and the vendored headers say `NVIMGCODEC_VER_MINOR 9`. Dropped it from this file and the cuslide2 copy.

Digging into it though, removing that line isn't quite enough. The list still ends with `libnvimgcodec.so.0` and `libnvimgcodec.so`, and those symlinks point at whatever major-0 build happens to be installed — so on a machine with only 0.8 we'd still pick it up without noticing. Combined with the two things you spotted elsewhere (`limit_images` gone in 0.9, no compile-time version checks), that would blow up at runtime instead of failing the build, which is a much worse way to find out.

So I added a check at configure time: it reads `NVIMGCODEC_VER_MAJOR`/`MINOR` out of the `nvimgcodec_version.h` we discover, errors below 0.9, warns if it's outside the pinned range, and skips quietly when the header isn't there so wheel and system installs keep working. The `find_package` branch now looks at `nvimgcodec_VERSION` too — it wasn't checking anything at all before.

On consolidating the two `nvimgcodec.cmake` files: fair point, and the check I just added is now duplicated as well. I'd rather do that separately than grow this PR — the two files get picked up through different `superbuild_depend()` search paths (cuslide2 via `SUPERBUILD_ADDITIONAL_DEPS_DIRS`, the other via `CMAKE_SUPERBUILD_DEPS_ROOT_DIR`, which points at the cuslide plugin's cmake dir and has no nvimgcodec entry), so merging them takes a little care not to bbreak standalone plugin builds.

cpp/plugins/cucim.kit.cuslide2/src/cuslide/tiff/types.h
cdinea: Fixed. This one had actually gone backwards — the base was already 2020-2026 and the branch reverted it to 2021

cpp/plugins/cucim.kit.cuslide2/src/cuslide/nvimgcodec/nvimgcodec_decoder.cpp
cdinea: You're right on both counts. `struct_size` should be `sizeof(nvimgcodecCodeStreamView_t)` — the `offsetof(..., limit_images)` was an old ABI shrink so pre-`limit_images` libraries wouldn't read past their known layout. That doesn't make sense in a 0.9-only PR, and `limit_images` is gone in the real 0.9 headers anyway.

Switched all four sites (three in `nvimgcodec_decoder.cpp`, one in `nvimgcodec_tiff_parser.cpp`) back to `sizeof`, and dropped the `limit_images = 0` assignment plus the ABI comment.

Why it still compiled: the vendored `nvimgcodec.h` in this tree still had `limit_images` even though it claimed 0.9. That lines up with your other comment about updating to the full new headers — done in ce00b56 (headers swapped from a real 0.9.0 install, and the plugin builds cleanly against them).

cpp/plugins/cucim.kit.cuslide2/src/cuslide/nvimgcodec/nvimgcodec_decoder.cpp
cdinea: Fixed in d7c0023 — `width`, `num_channels`, and `bytes_per_sample` are all `uint32_t`, so the product was wrapping in 32-bit before the assignment to `size_t`. Cast `width` to `size_t` so the multiply happens in 64-bit.

cpp/plugins/cucim.kit.cuslide2/src/cuslide/nvimgcodec/nvimgcodec_decoder.cpp
cdinea: Same fix applied here in d7c0023, plus the matching ssite in `schedule_batch_decode`.

cpp/plugins/cucim.kit.cuslide2/src/cuslide/tiff/tiff.cpp
cdinea: Fixed in 1328a76 — that `downsample > 0` guard was already present and got dropped when this block moved during the OME refactor. Restored it

cpp/plugins/cucim.kit.cuslide2/src/cuslide/tiff/tiff.cpp
cdinea: Fixed in bafbdb5. You're right that it could be — `node_offset` was only checked for `>= 0`, never against the length, so the `+ 34` could put us past the terminator, and both loop stop conditions (`>` and `\0`) are byte values inside the buffer, so the scan would just walk off the end of the string allocation.

This turned out to be a revert rather than a new bug: the bounds checks existed before this branch and got lost in a rebase. Restored the length check

cpp/plugins/cucim.kit.cuslide2/src/cuslide/cuslide.cpp
cdinea: Addressed in eac5aee — channel names are now built in an owning `std::vector<std::string>`.

We still convert to `std::pmr::vector<std::string_view>` at the end because `ImageMetadata::channel_names` takes that type and may retain pointers into the metadata buffer. With `_GLIBCXX_USE_CXX11_ABI=0` we also can't use `std::pmr::string`, so the one-shot copy into `resource` is what keeps the views alive for the metada

scripts/test_pyramidal_ome_tiff.py
cdinea: Addressed in 84fe4ed. Multi-level GPU/CPU reads now also do `np.array_equal` (with a max-diff error on mismatch), and `_validate_pyramid_geometry` asserts that each level shrinks, downsamples increase, and dims match `base / ds` within ±1 for rounding.let me know if further testign is needed

scripts/test_pyramidal_ome_tiff.py
cdinea: Addressed in 15752d3. Added a synthetic uint16 pyramidal OME-TIFF decode check that always runs: it asserts metadata/region dtype stays uint16, preserves values >255 (vso we didn't silently narrow to uint8), and matches source pixels on level 0 and level 1. When the caller-provided file is already uint16, the multi-level CPU/GPU checks now enforce that dtype too.

cpp/plugins/cucim.kit.cuslide2/cmake/deps/nvimgcodec.cmake
cdinea: Same as on the other cmake file — `0.8.0` is gone in 0912636, with a configure-time `>=0.9.0` check added.

On consolidating the two `nvimgcodec.cmake` files: agreed it's worth doing, but I'd rather keep that as a follow-up. They're reached through different `superbuild_depend()` search paths, so merging them cleanly without breaking the standalone plugin build is a bit more than this PR should take on.

cpp/plugins/cucim.kit.cuslide2/src/cuslide/nvimgcodec/nvimgcodec_tiff_parser.cpp
```diff
@@ -1153,7 +1256,8 @@ void TiffFileParser::extract_tiff_tags(IfdInfo& ifd_info)
     }
 #endif // CUSLIDE2_NVIMGCODEC_HAS_TIFF_TAG_METADATA
 
-    // Fallback: file extension heuristics when COMPRESSION tag is not present in file
+    // Fallback: file extension heuristics when COMPRESSION tag is not available
+    // (either nvImageCodec < 0.7.0 or tag not present in file)
```
LevKruglyak: Is comment still relevant? As far as I understand it is not possible to have nvImageCodec < 0.7.0 now

cpp/cmake/deps/nvimgcodec.cmake
mkepa-nv: sure, we can consolidate in another PR. In the next PR you can also consider adding runtime check that the correct version of nvimagecodec was loaded - I think it is good to load `libnvimgcodec.so.0` or just `libnvimgcodec.so` in case there are non others in the system, but I would verify in runtime via getProperties that the version is supported

cpp/plugins/cucim.kit.cuslide2/src/cuslide/cuslide.cpp
mkepa-nv: What is the reason for using _GLIBCXX_USE_CXX11_ABI=0? 

Before, I missed the fact that you are using pmr. If possible using string with pmr would be less tricky to read. But this can be done in a future PR

cpp/plugins/cucim.kit.cuslide2/src/cuslide/tiff/tiff.cpp
cdinea: I updated `parse_string_array` to resttore tolerant parsing behavior: per-item conversion failures are now handled locally, and parsing continues for remaining elements instead of aborting on the first malformed value.

This is now fixed in commit `78b8b15`.

cdinea:
/ok to test

78b8b15

cpp/plugins/cucim.kit.cuslide2/src/cuslide/nvimgcodec/nvimgcodec_tiff_parser.cpp
cdinea: Thanks for flagging this, Lev — addressed in `e011f44`; I removed the `<0.7.0` wording and updated the comment to a version-agnostic fallbak

mkepa-nv (approved):
There are a few of my comments left not addressed. But they are more nitpicks / code style, so I approve

PR #1131, opened 2026-08-14
selectively skip CI jobs based on changed files, update pre-commit hooks

jameslamb:
Closes #1128

Introduces the `changed-files` GitHub workflow used in many other RAPIDS repos. This configuration allows skipping CI in PRs based on which files changed.

For example, PRs that only change the README or `.gitignore` will now be mergeable without any test jobs that run on GPU runners 😁 

This also updates all pre-commit hooks via `pre-commit autoupdate`, to take advantage of the CI run + reviews.

PR #1133, opened 2026-08-19
ci: ensure pr-builder always runs

jameslamb:
#1131 introduced `changed-files` to skip CI based on which files a PR changes.

It missed also updating the `pr-builder` configuration to ensure that job always runs. That job is required by branch protections here, so when it is skipped merging will be blocked.

This fixes that.

## Notes for Reviewers

Changes like this one were made in most of the PRs for https://github.com/rapidsai/build-planning/issues/94, but `cucim` was excluded from that migration. I'm not sure why.


bdice (approved):

jakirkham (approved):
Thanks James! 🙏

Issue #1137, opened 2026-08-25
[BUG] `TestStainExtractorMacenko::test_result_value` fails with CuPy 14.2.0


jakirkham:
Seeing the following [test failure on CI]( https://github.com/rapidsai/cucim/actions/runs/32813452388/job/97793714999#step:12:27631 ):

```python
FAILED tests/unit/core/test_stain_normalizer.py::TestStainExtractorMacenko::test_result_value[image1-expected1] - AssertionError: 
Not equal to tolerance rtol=1e-06, atol=0

Mismatched elements: 6 / 6 (100%)
First 5 mismatches are at indices:
 [0, 0]: 0.5773503184318542 (ACTUAL), 0.0 (DESIRED)
 [0, 1]: 0.5773503184318542 (ACTUAL), 0.0 (DESIRED)
 [1, 0]: 0.5773503184318542 (ACTUAL), 0.70710678 (DESIRED)
 [1, 1]: 0.5773503184318542 (ACTUAL), 0.70710678 (DESIRED)
 [2, 0]: 0.5773503184318542 (ACTUAL), 0.70710678 (DESIRED)
Max absolute difference among violations: 0.57735032
Max relative difference among violations: 0.18350335
 ACTUAL: array([[0.57735, 0.57735],
       [0.57735, 0.57735],
       [0.57735, 0.57735]], dtype=float32)
 DESIRED: array([[0.      , 0.      ],
       [0.707107, 0.707107],
       [0.707107, 0.707107]])
```

Edit: For context `0.5773503184318542 ≈ 1 / √3` and `0.70710678 ≈ 1 / √2`. Also this is the case where it shows up

https://github.com/rapidsai/cucim/blob/209cd93ab3fa1b826a4739eb858daa2516fae025/python/cucim/tests/unit/core/test_stain_normalizer.py#L74-L84

jakirkham:
Have xfailed the test with PR: https://github.com/rapidsai/cucim/pull/1138

Though the underlying issue remains

grlee77:
## Root Cause

This is a numerical accuracy issue exposed by CuPy 14.2 when CUB is enabled. The test uses a degenerate, zero-variance input whose covariance matrix should be exactly zero.

[...]

PR #1138, opened 2026-08-25
XFail `TestStainExtractorMacenko::test_result_value` on CuPy 14.2.0

jakirkham:
This test fails with CuPy 14.2.0 (recently released). For now xfail it and raise a tracking issue.

Refs:

* https://github.com/rapidsai/cucim/issues/1137

### DIFF
--- a/python/cucim/tests/unit/core/test_stain_normalizer.py
+++ b/python/cucim/tests/unit/core/test_stain_normalizer.py
@@ -1,14 +1,18 @@
-# SPDX-FileCopyrightText: Copyright (c) 2022, NVIDIA CORPORATION.
+# SPDX-FileCopyrightText: Copyright (c) 2022-2026, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
 # SPDX-License-Identifier: Apache-2.0
 
 
 import cupy as cp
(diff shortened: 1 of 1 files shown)


python/cucim/tests/unit/core/test_stain_normalizer.py
jakirkham, line 84: When there is a test failure, it appears this is the case that fails

```
[gw6] [ 88%] XFAIL tests/unit/core/test_stain_normalizer.py::TestStainExtractorMacenko::test_result_value[image1-expected1]
```

Here is [an example on CI]( https://github.com/rapidsai/cucim/actions/runs/32885557880/job/97934070157?pr=1138#step:12:23441 )

jakirkham (commented):
Looks like this passed. IOW the xfail worked

Can also see the specific case that is xfailed CI. Have mentioned it below

grlee77 (approved):
Thanks @jakirkham  for adding the test skip to unblock CI for other issues. I will try to investigate the root cause soon.

jakirkham:
Thanks Greg! 🙏

Sounds good. Hope the insight above helps identify the cause

PR #1142, opened 2026-09-01
Fix stain covariance centering with CuPy CUB

### LINKED ISSUE #1137: [BUG] `TestStainExtractorMacenko::test_result_value` fails with CuPy 14.2.0

grlee77:
resolves #1137 by making stain normalization robust to a small change in numerical accuracy of some axis-specific `cupy.mean` operations when CUB is enabled in CuPy 14.2.

The root cause and two possible fixes are explained in this comment in that issue:
https://github.com/rapidsai/cucim/issues/1137#issuecomment-5495378665

The second option of shifting each row to make the computation numerically robust while preserving the performance of float32 reduction was kept.

### DIFF
--- a/python/cucim/src/cucim/core/operations/color/stain_normalizer.py
+++ b/python/cucim/src/cucim/core/operations/color/stain_normalizer.py
@@ -1,4 +1,4 @@
-# SPDX-FileCopyrightText: Copyright (c) 2022-2025, NVIDIA CORPORATION.
+# SPDX-FileCopyrightText: Copyright (c) 2022-2026, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
 # SPDX-License-Identifier: Apache-2.0
 
 import math
@@ -196,6 +196,11 @@ def _covariance(a):
     if fact <= 0:
         raise RuntimeError("Degrees of freedom <= 0")
 
+    # Shift before computing the mean to improve the accuracy of centering.
+    # In particular, this ensures that constant rows center to exact zeros
+    # regardless of the floating-point reduction order used by the backend.
+    offset = X[:, :1].copy()
+    X -= offset
     X -= X.mean(axis=1, keepdims=True)
     if not X.flags.f_contiguous:
         # TODO: Empirically, it is faster to use F order for .dot() call below.
--- a/python/cucim/tests/unit/core/test_stain_normalizer.py
+++ b/python/cucim/tests/unit/core/test_stain_normalizer.py
@@ -4,15 +4,21 @@
 
 import cupy as cp
 import pytest
-from packaging.version import parse
 
 from cucim.core.operations.color import (
     normalize_colors_pca,
     stain_extraction_pca,
 )
+from cucim.core.operations.color.stain_normalizer import _covariance
 
-# check CuPy instead of SciPy
-CUPY_GT_14_2 = parse(cp.__version__) >= parse("14.2.0a0")
+
+def test_covariance_constant_rows_are_exactly_zero():
+    value = cp.float32(5.4806389808654785)
+    image = cp.full((3, 8), value, dtype=cp.float32)
+
+    cp.testing.assert_array_equal(
+        _covariance(image), cp.zeros((3, 3), dtype=cp.float32)
+    )
 
 
 class TestStainExtractorMacenko:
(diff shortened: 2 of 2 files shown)


python/cucim/src/cucim/core/operations/color/stain_normalizer.py
```diff
@@ -196,6 +196,11 @@ def _covariance(a):
     if fact <= 0:
         raise RuntimeError("Degrees of freedom <= 0")
 
+    # Shift before computing the mean to improve the accuracy of centering.
+    # In particular, this ensures that constant rows center to exact zeros
+    # regardless of the floating-point reduction order used by the backend.
+    offset = X[:, :1].copy()
+    X -= offset
```
jakirkham, line 203: How do we decide which part of X to use for this offset?

python/cucim/tests/unit/core/test_stain_normalizer.py
```diff
@@ -4,15 +4,21 @@
...
-# check CuPy instead of SciPy
-CUPY_GT_14_2 = parse(cp.__version__) >= parse("14.2.0a0")
+
+def test_covariance_constant_rows_are_exactly_zero():
+    value = cp.float32(5.4806389808654785)
+    image = cp.full((3, 8), value, dtype=cp.float32)
+
+    cp.testing.assert_array_equal(
+        _covariance(image), cp.zeros((3, 3), dtype=cp.float32)
+    )
```
jakirkham, line 21: Good idea to test precision issues are avoided.

jakirkham (commented):
Thanks Greg! 🙏

Had one question below

python/cucim/src/cucim/core/operations/color/stain_normalizer.py
grlee77: For a constant-valued matrix as in the failure case it shouldn't matter which element is subtracted for each row. Maybe "improve the accuracy" is overstating things a bit as this row could be an outlier for more general data so may not always improve for general cases.

The general property of covariance matrices used here is that they are translation invariant. For random variables `X` and `Y` and some constants `a` and `b`:  `cov(X + a, Y + b) == cov(X, Y)`

jakirkham: Yep agree that mathematically this should be a no-op as far as the result is concerned

Was just curious about the particular choice of translation to apply. Is there anything else we should be considering when making this choice?

grlee77: Not sure if it may be more robust in general to subtract the mean of the values instead of one particular one. Let's just try it like this for now

jakirkham: Greg and I discussed this. Maybe we could subtract the mean. For now we will give this a try since it doesn't mathematically change the result, but it should fix the constant case that was causing us issues before. Can revisit if different issues show up later

jakirkham (approved):

jakirkham:
Whoops, this should have gone in 26.10

Backporting the fix with PR: https://github.com/rapidsai/cucim/pull/1154

Issue #1069, opened 2026-04-01
[FEA] add  GPU-resident tile caching for cuslide2


cdinea:
<img width="1012" height="1084" alt="Image" src="https://github.com/user-attachments/assets/954b0a82-fd21-4f09-8475-1f21da65e43e" />

PR #1143, opened 2026-09-07
Add per-codec tile cache benchmark for cuslide2, fix raw TIFF recipe

cdinea:
Addresses the review feedback on #1034 / #1069 from @mkepa-nv, which asked for two things: numbers on whether a warm cache read actually beats decoding again when the codec is light, and whether the tiles would be better cached in GPU memory.

This PR delivers the benchmark. It does **not** implement a GPU-resident cache — that is an architecture decision (make the global cache device-aware, vs. keep separate CPU and GPU caches as suggested in the thread), and the point of the benchmark is to make that call with evidence rather than intuition.

## What the benchmark measures

`scripts/benchmark_tile_cache.py` sweeps compression × cache setting × output device, timing two access patterns per cell:

- **warm** — the same region read repeatedly, which is the case tile caching is meant to win
- **cold** — distinct non-overlapping regions, which pays decode plus insert overhead

Cache hit/miss counters are reported for every cell. This is load-bearing rather than decorative: the cache path requires `tile_width_ > 0 && tile_height_ > 0`, so whenever nvImageCodec does not report `TileWidth`/`TileLength` the tile dims stay `(0,0)` and the cache is bypassed entirely. Without the counters, a run where the cache silently never engaged looks exactly like a run where caching genuinely did not help. The script labels that case instead of printing a meaningless ratio.

## The CUDA rows need reading carefully

For `device="cuda"` this is not a plain cache-vs-no-cache comparison, and the script says so in its output. With the cache **off**, nvImageCodec decodes straight into device memory:

https://github.com/rapidsai/cucim/blob/main/cpp/plugins/cucim.kit.cuslide2/src/cuslide/nvimgcodec/nvimgcodec_decoder.cpp#L317-L374

With the cache **on**, tiles are decoded to host regardless of the requested output device, assembled there with a single-threaded per-row `memcpy`, and the finished region is copied up with one `cudaMemcpyHostToDevice`.

So enabling the cache *gives up* direct-to-device decode for GPU output. A warm ratio below `1.0x` on the CUDA rows is the cost of that host round-trip, and is an argument **for** the GPU-resident cache in #1069 rather than against caching in general. Worth noting the infrastructure is already most of the way there: `PerProcessImageCache::allocate()` calls `cudaMalloc` when constructed with `DeviceType::kCUDA`, the global cache simply never is, and legacy cuslide's `NvJpegProcessor` already keeps a private device-resident tile cache as a working in-tree reference.

## Generator fix (prerequisite)

The uncompressed case central to the review feedback could not be generated at all. In `gen_tiff.py`, `raw` mapped to `None` and was then overwritten:

```python
compression = COMPRESSION_MAP.get(compression)
compressionargs = None
if not compression:
    compression = "jpeg"
```

So every `testimg_tiff_stripe_*_raw` fixture has in fact been producing JPEG images, and unknown compression names took the same silent fallback. Both are fixed; unknown names now raise.

Verified directly:

```
    jpeg: compression=COMPRESSION.JPEG      size=  46666 bytes
 deflate: compression=COMPRESSION.DEFLATE   size=   3380 bytes
     raw: compression=COMPRESSION.NONE      size= 246832 bytes

unknown compression rejected: Unknown compression 'lzw'. Expected one of ['deflate', 'jpeg', 'raw']
```

## Reviewer notes

- **Please let CI confirm the fixture change.** `gen_tiff.py` feeds the compression-parametrized fixtures used across `test_tiff_read_region.py`, `test_batch_decoding.py` and `test_load_image_metadata.py`. I checked their assertions are structural (e.g. `assert np.all(channel_value_count > 0)`) rather than exact-pixel comparisons, so lossy→lossless should be safe, but I had no cuCIM install or GPU available to run the suite locally.
- The `4096x4096` `raw` fixture gets materially larger now that it is genuinely uncompressed instead of silently JPEG.
- Unrelated rough edge noticed while working here, left alone: `--discard_cache` in the C++ benchmarks only drops the OS page cache via `posix_fadvise`, not cuCIM's `ImageCache`, so no existing C++ benchmark exercises the tile cache.

### DIFF
--- a/scripts/benchmark_tile_cache.py
+++ b/scripts/benchmark_tile_cache.py
@@ -1,0 +1,355 @@
+#!/usr/bin/env python3
+# SPDX-FileCopyrightText: Copyright (c) 2026, NVIDIA CORPORATION. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+
+"""Measure whether cuslide2's tile cache pays for itself, per codec and device.
+
+Review feedback on PR #1034 / issue #1069 raised two questions that no
+existing benchmark answers:
+
+1. For light codecs, is a warm cache read actually *slower* than decoding
+   again?  The cache stores decoded tiles in host memory and assembles the
+   requested region with a single-threaded per-row ``memcpy``, so when decode
+   is cheap that copy can dominate.
+
+2. Would a GPU-resident cache be better?  This matters more than it first
+   appears: with the cache disabled, nvImageCodec decodes straight into device
+   memory (``STRIDED_DEVICE``).  With it enabled, cuslide2 decodes tiles to
+   host, assembles on host, then copies the whole region up with a single
+   ``cudaMemcpyHostToDevice``.  Enabling the cache therefore *gives up*
+   direct-to-device decode for CUDA output, so a "cache on" number that looks
+   bad for ``device="cuda"`` is evidence for the GPU-resident design rather
+   than against caching as such.
+
+The measured grid is compression x cache setting x output device.  Two access
+patterns are timed for each cell: repeated reads of one region (the warm-hit
+case the cache is meant to win) and reads of distinct regions (the miss path,
+which pays decode plus insert overhead).
+
+Cache hit/miss counters are reported alongside the timings.  Without them a
+run where the cache silently never engaged -- which happens whenever
+nvImageCodec does not expose TileWidth/TileLength, leaving tile dims at (0,0)
+-- is indistinguishable from one where caching genuinely did not help.
+"""
+
+import argparse
+import shutil
+import statistics
+import sys
+import tempfile
+import time
+from pathlib import Path
+
+import numpy as np
+from test_common import setup_environment
+
+REPO_ROOT = Path(__file__).resolve().parent.parent
+COMPRESSIONS = ("jpeg", "deflate", "raw")
+
+
+def generate_images(dest, compressions, image_size, tile_size):
+    """Generate one tiled TIFF per compression scheme.
+
+    Each recipe gets its own subdirectory because the generator derives the
+    output filename from pattern/size/tile only -- compression is not part of
+    it, so same-size recipes would otherwise overwrite each other.
+    """
+    sys.path.insert(0, str(REPO_ROOT / "python" / "cucim" / "tests" / "util"))
+    from gen_image import ImageGenerator
+
+    paths = {}
+    for compression in compressions:
+        subdir = dest / compression
+        subdir.mkdir(parents=True, exist_ok=True)
+        recipe = f"tiff::stripe:{image_size}:{tile_size}:{compression}"
+        (path,) = ImageGenerator(str(subdir), [recipe]).gen()
+        paths[compression] = Path(path)
+        print(
+            f"  {compression:<8} {Path(path).name}  "
+            f"{Path(path).stat().st_size / 2**20:.1f} MB"
+        )
+    return paths
+
+
+def _synchronize():
+    """Best-effort device sync so async work is not attributed to the next read."""
+    try:
+        import cupy
+
+        cupy.cuda.runtime.deviceSynchronize()
+    except Exception:
+        pass
+
+
+def time_reads(img, regions, size, device, repeats):
+    """Return per-read wall times in milliseconds for *device* output.
+
+    The CPU path is materialized via ``np.asarray``, which is a zero-copy view
+    over the returned buffer.  The CUDA path deliberately does *not* copy back
+    to host: a device-to-host transfer is not part of what read_region costs,
+    and adding one would swamp the very difference being measured.
+    """
+    on_cuda = device == "cuda"
+    timings = []
+    for _ in range(repeats):
+        for origin in regions:
+            if on_cuda:
+                _synchronize()
+            start = time.perf_counter()
+            region = img.read_region(origin, size, level=0, device=device)
+            if on_cuda:
+                _synchronize()
+            else:
+                # Zero-copy view; ensures the buffer is real, not lazy.
+                np.asarray(region)
+            timings.append((time.perf_counter() - start) * 1000.0)
+    return timings
+
+
+def measure(path, CuImage, cache_type, device, size, repeats, num_regions):
+    """Time warm and cold access patterns for one cache/device combination."""
+    # The IFD binds to whatever cache manager is active when the image is
+    # opened, so configure the cache first and open afterwards.
+    if cache_type == "nocache":
+        CuImage.cache("nocache")
+    else:
+        CuImage.cache(cache_type, memory_capacity=1024)
+    CuImage.cache().record(True)
+
+    img = CuImage(str(path))
+    tile_sizes = img.resolutions.get("level_tile_sizes", ())
+    tile = list(tile_sizes[0]) if tile_sizes else [0, 0]
+
+    width, height = img.shape[1], img.shape[0]
+    # Distinct, non-overlapping origins for the miss path.
+    step = max(size[0], size[1])
+    regions = [
+        (
+            min((i * step) % max(width - size[0], 1), width - size[0]),
+            min((i * step) % max(height - size[1], 1), height - size[1]),
+        )
+        for i in range(num_regions)
+    ]
+
+    result = {
+        "cache": cache_type,
+        "device": device,
+        "tile": tile,
+        "cache_active": bool(tile[0] > 0 and tile[1] > 0) and cache_type != "nocache",
+    }
+
+    # Warm: same region repeatedly.  First read populates, the rest should hit.
+    time_reads(img, [regions[0]], size, device, 1)
+    before_hits = CuImage.cache().hit_count
+    before_misses = CuImage.cache().miss_count
+    warm = time_reads(img, [regions[0]], size, device, repeats)
+    result["warm_ms"] = statistics.median(warm)
+    result["warm_hits"] = CuImage.cache().hit_count - before_hits
+    result["warm_misses"] = CuImage.cache().miss_count - before_misses
+
+    # Cold: distinct regions, so every tile is a first touch.  Reopening also
+    # drops any per-image state carried over from the warm pass.
+    img_cold = CuImage(str(path))
+    before_hits = CuImage.cache().hit_count
+    before_misses = CuImage.cache().miss_count
+    cold = time_reads(img_cold, regions, size, device, 1)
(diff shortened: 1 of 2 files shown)


scripts/benchmark_tile_cache.py
```diff
@@ -0,0 +1,355 @@
...
+    tile_sizes = img.resolutions.get("level_tile_sizes", ())
+    tile = list(tile_sizes[0]) if tile_sizes else [0, 0]
+
+    width, height = img.shape[1], img.shape[0]
+    # Distinct, non-overlapping origins for the miss path.
+    step = max(size[0], size[1])
+    regions = [
+        (
+            min((i * step) % max(width - size[0], 1), width - size[0]),
+            min((i * step) % max(height - size[1], 1), height - size[1]),
```
gigony, line 129: ### [P2] Generate a tile-disjoint grid instead of cycling diagonal coordinates

**AI-assisted review comment (finding 1/2).**

With the CLI defaults, these modulo expressions produce only three distinct origins across eight reads: `(0,0)`, `(1024,1024)`, and `(2048,2048)`, then repeat them. Five reads therefore revisit already-cached tiles even with a fresh cache, so the reported cold median can describe warm hits rather than decode-plus-insert cost. Please enumerate a bounded 2-D, tile-disjoint grid and reject requests for more regions than fit.

**Validation:** Validation during report preparation used the actual measure() control flow with an in-memory tile-cache/timing test double. It reproduced the default 96-hit/32-miss cold pass. No real TIFF decoding, benchmark timing, or CUDA execution was performed. Proposed patches passed git apply --check without being applied.

<details>

### Root cause and impact

Both axes advance with the same index, then wrap independently. With the CLI defaults (`4096x4096` image, `1024x1024` region, eight regions), the generated sequence is:

```text
(0, 0), (1024, 1024), (2048, 2048),
(0, 0), (1024, 1024), (2048, 2048),
(0, 0), (1024, 1024)
```

There are only three unique origins. The final valid origin at coordinate 3072 is never reached because the modulo denominator is 3072. Five of the eight reads repeat a region from the same pass, even if the cache is initially empty. A median can therefore describe warm hits rather than decode-plus-insert cost.

For other CLI dimensions, merely ensuring distinct pixel rectangles is also insufficient: neighboring rectangles can share decoded TIFF tiles when their boundaries are not tile-aligned. The workload must be disjoint at the cache's tile granularity.

### Reproduction and validation

Executed the actual `measure()` function extracted from the checked-out Git object, replacing only image I/O/timing with an in-memory tile-cache test double. At the default 256-pixel tile size, each region covers 16 tiles. With the existing warm pass and global cache lifetime, the cold pass produced:

```text
unique cold origins: 3
cold_hits: 96
cold_misses: 32
```

The test double assigns hits and misses synthetic costs; those costs are not real benchmark timings. The coordinates and hit/miss accounting validate the workload defect without requiring CUDA.

### Suggested fix

Enumerate a bounded 2-D grid, round each stride up to a whole tile, and reject requests for more disjoint regions than fit. Validate positive sizes/counts rather than wrapping back over previously read tiles. When tile dimensions are unknown, use a one-pixel alignment; the existing inactive-cache reporting still applies.

```diff
diff --git a/scripts/benchmark_tile_cache.py b/scripts/benchmark_tile_cache.py
--- a/scripts/benchmark_tile_cache.py
+++ b/scripts/benchmark_tile_cache.py
@@ -123,10 +123,19 @@
     width, height = img.shape[1], img.shape[0]
     # Distinct, non-overlapping origins for the miss path.
-    step = max(size[0], size[1])
+    if len(size) != 2 or any(s <= 0 for s in size):
+        raise ValueError("region must contain two positive dimensions")
+    if size[0] > width or size[1] > height:
+        raise ValueError("region must fit inside the image")
+    if repeats < 1 or num_regions < 1:
+        raise ValueError("repeats and regions must be positive")
+    tile_w, tile_h = (max(t, 1) for t in tile)
+    step_x = ((size[0] + tile_w - 1) // tile_w) * tile_w
+    step_y = ((size[1] + tile_h - 1) // tile_h) * tile_h
+    xs = range(0, width - size[0] + 1, step_x)
+    ys = range(0, height - size[1] + 1, step_y)
+    if num_regions > len(xs) * len(ys):
+        raise ValueError("not enough tile-disjoint regions fit in the image")
     regions = [
-        (
-            min((i * step) % max(width - size[0], 1), width - size[0]),
-            min((i * step) % max(height - size[1], 1), height - size[1]),
-        )
+        (xs[i % len(xs)], ys[i // len(xs)])
         for i in range(num_regions)
     ]
```

### Regression coverage

- Assert eight distinct regions for the CLI defaults, including origins in multiple rows.
- Assert that every region fits inside the image and that the sets of covered tile indices are pairwise disjoint.
- Cover rectangular images/regions and region dimensions not divisible by the tile dimensions.
- Reject zero counts, oversized regions, and requests exceeding the grid capacity.
- With a fresh active cache, require zero hits and the expected number of tile misses during the cold pass.

</details>

scripts/benchmark_tile_cache.py
```diff
@@ -0,0 +1,355 @@
...
+    result["warm_ms"] = statistics.median(warm)
+    result["warm_hits"] = CuImage.cache().hit_count - before_hits
+    result["warm_misses"] = CuImage.cache().miss_count - before_misses
+
+    # Cold: distinct regions, so every tile is a first touch.  Reopening also
+    # drops any per-image state carried over from the warm pass.
+    img_cold = CuImage(str(path))
+    before_hits = CuImage.cache().hit_count
+    before_misses = CuImage.cache().miss_count
+    cold = time_reads(img_cold, regions, size, device, 1)
```
gigony, line 155: ### [P2] Reset the global cache before the cold phase

**AI-assisted review comment (finding 2/2).**

Reopening the same file does not reset the process-wide tile cache. The preceding warm phase populated `regions[0]`, which is also the first cold region; with `--regions 1`, the entire purported cold measurement is warm. Please recreate the configured cache before this phase, outside the timing interval. This is independent of fixing the repeated coordinates.

**Validation:** Validation during report preparation used the actual measure() control flow with an in-memory tile-cache/timing test double. It reproduced the default 96-hit/32-miss cold pass. No real TIFF decoding, benchmark timing, or CUDA execution was performed. Proposed patches passed git apply --check without being applied.

<details>

### Root cause and impact

The warm phase fills `regions[0]`, and the cold phase immediately reads that same region. A new `CuImage` object is not a new tile cache:

- `cpp/plugins/cucim.kit.cuslide2/src/cuslide/tiff/ifd.cpp:612-623` obtains the global cache from `CuImage::cache_manager()` at read time.
- `ifd.cpp:646-727` keys lookups by the file/IFD hash and tile index, not the Python image object's identity.
- `cpp/src/filesystem/file_handle.cpp:29-36` derives the file hash from device, inode, and modification time. Reopening an unchanged file preserves that identity.
- `cpp/src/cache/image_cache_manager.cpp:60-64` replaces the cache when a new cache configuration is supplied. Constructing another image does not do so.

Consequently, even after finding 1 is fixed, the first cold read is still a complete cache hit. With `--regions 1`, the entire supposed cold result is a warm read, directly defeating the PR's stated decode-plus-insert measurement.

### Suggested fix

Recreate the configured cache and enable recording before opening the cold image. Cache initialization must remain outside the timed reads. Keep this as a tile-cache cold measurement; dropping the OS page cache is a separate experiment and is not required here.

```diff
diff --git a/scripts/benchmark_tile_cache.py b/scripts/benchmark_tile_cache.py
--- a/scripts/benchmark_tile_cache.py
+++ b/scripts/benchmark_tile_cache.py
@@ -150,6 +150,11 @@
     # Cold: distinct regions, so every tile is a first touch.  Reopening also
     # drops any per-image state carried over from the warm pass.
+    if cache_type == "nocache":
+        CuImage.cache("nocache")
+    else:
+        CuImage.cache(cache_type, memory_capacity=1024)
+    CuImage.cache().record(True)
     img_cold = CuImage(str(path))
     before_hits = CuImage.cache().hit_count
     before_misses = CuImage.cache().miss_count
     cold = time_reads(img_cold, regions, size, device, 1)
```

### Regression coverage

- Warm a region, reopen the same file without changing the cache, and verify that this alone still hits the cache.
- Recreate the cache and verify that the same region then misses.
- Test `num_regions=1`: cold hits must be zero despite the preceding warm phase.
- Apply both fixes together: the default cold pass should have 0 hits and 128 misses for 256x256 tiles and 1024x1024 regions, assuming the cache path is active.

</details>

gigony (commented):

**AI-assisted review.** Sharing 2 AI-assisted inline findings from the prepared review report.

Reviewed commit: `301ced1b217429fc8cda1c97e4dd7f806e5acd37`. The PR description was read before the diff; the current head and inline locations were rechecked before publication.

**Validation and limitations:** Validation during report preparation used the actual measure() control flow with an in-memory tile-cache/timing test double. It reproduced the default 96-hit/32-miss cold pass. No real TIFF decoding, benchmark timing, or CUDA execution was performed. Proposed patches passed git apply --check without being applied.

<details>

### Scope and method

Checked out with `gh pr checkout 1143`, read `gh pr view 1143`, then reviewed `gh pr diff 1143 --color never` and the relevant generator, cache manager, file-hash, and cuslide2 read paths. The intended CPU/GPU decode-path difference is explicitly described in the PR and is not itself a finding.

All locations below refer to the reviewed head. Relevant PR additions are excerpts, not standalone patches. Suggested fixes are proposals against that checkout; no implementation files were changed.

### Other reviewed change

The generator change correctly preserves `COMPRESSION_MAP["raw"] == None` instead of substituting JPEG. `ImageGenerator.gen()` supplies `"jpeg"` when a recipe omits compression, so the explicit validation does not remove that documented default. No additional correctness finding was identified in this change. Actual TIFF generation and the compression-parametrized reader tests still need validation in the supported environment.

</details>

cdinea:
@gigony thank you for the review - I addressed all the feedback .Please take another look

gigony (approved):
Thanks @cdinea for the change!
It looks good to me.

jakirkham:
One CI job failed, but it looks unrelated. Have raised an issue: https://github.com/rapidsai/cucim/issues/1166

As it looks sporadic, will try restarting to see if it clears for now

Though we should follow up in the issue

jakirkham (approved):

cdinea:
> One CI job failed, but it looks unrelated. Have raised an issue: #1166
> 
> As it looks sporadic, will try restarting to see if it clears for now
> 
> Though we should follow up in the issue

thank you for the help @jakirkham

Issue #859, opened 2025-04-02
[FEA] Refactor the build system to adopt rapids-cmake


gigony:
cuCIM's current build setup does not use rapids-cmake.  
To align with other RAPIDS projects, I plan to refactor the build system to adopt rapids-cmake.

Reference: https://github.com/rapidsai/cucim/pull/858#discussion_r2025289416

bdice:
See also #627 and https://github.com/rapidsai/cucim/issues/305#issuecomment-1670273320.

PR #1147, opened 2026-09-07
Route device allocations through a CCCL memory resource

cdinea:
## Summary

Adopts CCCL memory resources for cuCIM's device allocations, following the RAPIDS guidance from @bdice in `#swrapids-cucim`: APIs should accept a `cuda::mr::resource_ref` so resources from either CCCL or RMM can be passed in, allocations returned to the user should come from the caller's resource, and there should be no process-wide "current" resource.

This is the foundation plus the internal allocation sites. It deliberately stops short of output allocations, for a reason detailed below that I think needs a decision before going further.

## What this adds

`cucim::memory::DeviceResources` — a non-global handle bundling a device-accessible memory resource with the stream that orders its allocations. It is passed explicitly and carried by objects that own device memory, so the resource that allocated a buffer is the one that frees it. There is no singleton and no default global resource, per the guidance that RMM's per-device resource map is being removed.

`cucim::memory::CudaMemoryResource` — the previous `cudaMalloc`/`cudaFree` behaviour expressed as a CCCL resource, so a default-constructed handle behaves exactly as cuCIM did before and only callers who supply a resource see anything different.

The public type in the API is `cuda::mr::resource_ref<cuda::mr::device_accessible>`, as recommended, which is what allows an `rmm::device_async_resource_ref` to be passed straight through.

## What this converts

The sites where core both allocates *and* frees, so the pairing is provably correct:

- **Device tile cache** (`PerProcessImageCache`). Allocates through the resource; each cached value keeps the resource it was allocated from and frees through that, so an entry evicted while a reader still holds it is still returned correctly, and a pool gets its blocks back rather than losing them to `cudaFree`.
- **`move_raster_from_host` / `move_raster_from_device`**, which gain resource-aware overloads. The existing signatures delegate to a default handle, so plugin callers are unaffected.

## What this does not convert, and why

Output allocations are the one thing the guidance calls a must ("if cuCIM returns device memory, it should definitely use the provided memory resource"). They are not converted here because cuCIM cannot currently do it correctly:

- Device memory reaching the user is often allocated **inside a plugin** — `cuslide2`'s `ifd.cpp:556,836` and `nvimgcodec_decoder.cpp:179` all call `cudaMalloc` directly.
- It is freed by **core**, in `CuImage::~CuImage` (`cuimage.cpp:285`), which calls `cudaFree` regardless of where the buffer came from.
- `ImageDataDesc` is `{container, shm_name, loader}` — there is **no field through which a resource could cross the plugin boundary**, and the boundary is a C ABI.

So converting `CuImage`'s free today would pair a plugin's raw `cudaMalloc` with a caller's pool resource, corrupting the pool. Getting output allocations onto a memory resource needs the plugin ABI to carry one first, which is a separate design change. Related: cuCIM has no `DLManagedTensor` deleter at all, so ownership of returned memory is implicit in `CuImage` rather than expressed in the DLPack container.

## Notes for reviewers

**CCCL is a new dependency.** cuCIM has no rapids-cmake or CPM, so it is added through the existing hand-rolled `FetchContent` pattern in `cpp/cmake/deps/`, pinned to `v3.4.2`. Worth noting that RMM was deliberately *removed* from cuCIM in #175/#946, so re-introducing a RAPIDS memory stack is a reversal that should be a conscious choice.

**Include order is load-bearing.** The CUDA toolkit bundles its own libcudacxx, older than `cuda::mr` for every CUDA version cuCIM supports. If the toolkit's include directory is searched first, CCCL's `<cuda/memory_resource>` still resolves but its `cuda/std/*` dependencies come from the toolkit, the version guards do not line up, and `cuda::mr` silently disappears with a confusing "`mr` in namespace `cuda` does not name a type". Both affected targets add CCCL with `target_include_directories(... BEFORE ...)`. I hit this while verifying, so it seemed worth commenting in-tree rather than leaving as a trap.

**The API in the original advice has since split.** CCCL now has both `resource_ref` (stream-ordered) and `synchronous_resource_ref`, and the async methods are spelled `allocate(stream, bytes, alignment)`. This uses the stream-ordered `resource_ref` as named in the guidance, since that is what matches RMM.

**`cuda::buffer` is not adopted here.** It is recommended for holding device memory, but is still published under `/unstable/` in the CCCL docs, and using it would change `ImageDataDesc`/DLPack layout. Left as a follow-up rather than baking an unstable API into the public container — happy to do it if you'd rather.

**Packaging is unfinished.** `device_resources.h` exposes `cuda::mr` types, so it is linked `PUBLIC` but `BUILD_INTERFACE` only. Installed consumers of that header will need CCCL exported or a `cuda-cccl` dependency added; not done here.

## Open question from the guidance

The advice explicitly left this to us: whether the memory resource should control output allocations, temporary allocations, or both, and whether to accept a second resource for temporaries. This PR uses a single resource for the internal allocations it converts. If you want separate output/temporary resources, that shape is easier to change now than later.

## Testing

`cpp/tests/test_device_resources.cpp` (new) covers:

- `CudaMemoryResource` and a user-written resource satisfy `cuda::mr::synchronous_resource`, `cuda::mr::resource` and `device_accessible` — this is the property that makes RMM/CCCL/user resources interchangeable, so it is asserted statically.
- Allocations reach the supplied resource, with the requested alignment, and every allocation is matched by a deallocation carrying the original size.
- Zero-byte allocations and `nullptr` deallocations never reach the resource.
- Allocation failure surfaces as `std::bad_alloc`, which the tile cache relies on to fall back to an uncached decode.
- A copied handle can still free memory the original allocated, which is what cached tile values depend on.

Verification I was able to do in this environment, and its limits: the new test file and both modified translation units compile clean with the build's own flags. All assertions were additionally run standalone and pass (16/16) — the checked-in Catch2 test could not be linked locally because the prebuilt `libCatch2.a` is ABI-incompatible with the local g++. **CMake configure was verified to fetch CCCL without configuring CCCL's own project, but a full cuCIM build and a GPU run were not possible here (no GPU access), so CI is the real check.**

One thing worth flagging: writing the tests caught that `any_resource` stores the resource *by value*, so wrapping a stateful allocator by value duplicates it and the handle's copy is what serves requests. That is fine for the intended usage (wrapping a reference such as `rmm::device_async_resource_ref`) but is a footgun, so it is documented on the class.

## Drive-by fixes

Two leaks on error paths in code this PR already touches: a failed `cudaMemcpy` in each of `move_raster_from_host`/`move_raster_from_device` threw without releasing the buffer it had just allocated. Also drops a dead `<memory_resource>` include in `memory_manager.cpp` that referred to `std::pmr`, which is actively confusing next to real memory-resource work.

### DIFF
--- a/cpp/CMakeLists.txt
+++ b/cpp/CMakeLists.txt
@@ -99,6 +99,7 @@ add_library(${CUCIM_PACKAGE_NAME}
         src/loader/thread_batch_data_loader.cpp
         src/logger/logger.cpp
         src/logger/timer.cpp
+        src/memory/device_resources.cpp
         src/memory/memory_manager.cpp
         src/plugin/image_format.cpp
         src/plugin/plugin_config.cpp
@@ -155,6 +156,24 @@ target_link_libraries(${CUCIM_PACKAGE_NAME}
             deps::json
             deps::nvtx3
             deps::taskflow
+        )
+
+# CCCL's headers have to be searched before the CUDA toolkit's.
+#
+# The toolkit ships its own copy of libcudacxx, and for every CUDA version
+# cuCIM supports that copy predates cuda::mr. If the toolkit include directory
+# (pulled in by CUDA::cudart_static) is searched first, CCCL's
+# <cuda/memory_resource> still resolves but the cuda/std/* headers it depends
+# on come from the toolkit, the version guards do not line up, and cuda::mr
+# silently disappears. BEFORE is what keeps the two copies from being mixed.
+#
+# PUBLIC because <cucim/memory/device_resources.h> exposes cuda::mr types.
+# BUILD_INTERFACE only: installed consumers of that header still need CCCL
+# exported or a cuda-cccl dependency added, which is not done yet.
+target_include_directories(${CUCIM_PACKAGE_NAME}
+        BEFORE
+        PUBLIC
+            $<BUILD_INTERFACE:${deps-cccl_SOURCE_DIR}/libcudacxx/include>
         )
 
 if (CUCIM_STATIC_GDS)
--- a/cpp/cmake/deps/cccl.cmake
+++ b/cpp/cmake/deps/cccl.cmake
@@ -1,0 +1,43 @@
+#
+# cmake-format: off
+# SPDX-FileCopyrightText: Copyright (c) 2026, NVIDIA CORPORATION.
+# SPDX-License-Identifier: Apache-2.0
+# cmake-format: on
+#
+
+if (NOT TARGET deps::cccl)
+    FetchContent_Declare(
+            deps-cccl
+            GIT_REPOSITORY https://github.com/NVIDIA/cccl.git
+            GIT_TAG v3.4.2
+            GIT_SHALLOW TRUE
+            # `libcudacxx/include` holds no CMakeLists.txt, so this downloads the
+            # sources without configuring CCCL's own project, which would build
+            # Thrust, CUB and their test suites. Only the header-only
+            # `cuda::mr` types are needed here.
+            SOURCE_SUBDIR libcudacxx/include
+            EXCLUDE_FROM_ALL
+    )
+    message(STATUS "Fetching CCCL sources")
+
+    FetchContent_MakeAvailable(deps-cccl)
+    message(STATUS "Fetching CCCL sources - done")
+
+    # Note for consumers: linking this target is not by itself enough for any
+    # target that also pulls in the CUDA toolkit include directory. The toolkit
+    # bundles an older libcudacxx, and whichever include directory is searched
+    # first wins for cuda/std/*, which decides whether cuda::mr exists at all.
+    # Targets in that situation add this directory with
+    # target_include_directories(... BEFORE ...); see cpp/CMakeLists.txt.
+    add_library(deps::cccl INTERFACE IMPORTED GLOBAL)
+    # SYSTEM so that CCCL's own headers do not trip cuCIM's -Werror settings.
+    set_target_properties(deps::cccl PROPERTIES
+        INTERFACE_SYSTEM_INCLUDE_DIRECTORIES
+            "${deps-cccl_SOURCE_DIR}/libcudacxx/include"
+        INTERFACE_INCLUDE_DIRECTORIES
+            "${deps-cccl_SOURCE_DIR}/libcudacxx/include"
+    )
+
+    set(deps-cccl_SOURCE_DIR ${deps-cccl_SOURCE_DIR} CACHE INTERNAL "" FORCE)
+    mark_as_advanced(deps-cccl_SOURCE_DIR)
+endif ()
--- a/cpp/include/cucim/memory/device_resources.h
+++ b/cpp/include/cucim/memory/device_resources.h
@@ -1,0 +1,211 @@
+/*
+ * SPDX-FileCopyrightText: Copyright (c) 2026, NVIDIA CORPORATION.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+#ifndef CUCIM_MEMORY_DEVICE_RESOURCES_H
+#define CUCIM_MEMORY_DEVICE_RESOURCES_H
+
+#include "cucim/core/framework.h"
+#include "cucim/io/device.h"
+
+#include <cuda/memory_resource>
+#include <cuda/stream_ref>
+
+#include <cstddef>
+
+/**
+ * Device memory allocation through a caller-supplied CCCL memory resource.
+ *
+ * Note on packaging: this header exposes `cuda::mr` types, so anything
+ * including it needs CCCL headers on its include path. The build links CCCL
+ * through `$<BUILD_INTERFACE:deps::cccl>`; exporting it for installed consumers
+ * (either by installing the headers or by depending on `cuda-cccl`) is still
+ * outstanding.
+ */
+namespace cucim::memory
+{
+
+/**
+ * @brief The set of CCCL properties cuCIM requires of a device memory resource.
+ *
+ * Device-accessible only. cuCIM hands the resulting pointers to CUDA kernels,
+ * nvImageCodec and `__cuda_array_interface__` consumers, none of which can use
+ * host-only memory.
+ */
+using device_accessible = ::cuda::mr::device_accessible;
+
+/**
+ * @brief Alignment cuCIM requests for device allocations.
+ *
+ * `cudaMalloc` already guarantees at least this much, so the default resource
+ * ignores the value; it is stated explicitly because a pool resource supplied
+ * by the caller will honour it, and image rows are addressed on 256-byte
+ * boundaries downstream.
+ */
+inline constexpr std::size_t kDefaultAlignment = 256;
+
+/**
+ * @brief Stream-ordered, device-accessible resource reference.
+ *
+ * This is the type that should appear in cuCIM APIs that allocate device
+ * memory. It binds to resources from either CCCL or RMM, so callers already
+ * holding an `rmm::device_async_resource_ref` can pass it straight through.
+ */
+using device_resource_ref = ::cuda::mr::resource_ref<device_accessible>;
+
+/**
+ * @brief Owning, type-erased, device-accessible resource.
+ *
+ * Used where cuCIM has to keep a resource alive rather than borrow one, most
+ * importantly for memory handed back to the user: whatever allocated a buffer
+ * has to still exist when that buffer is freed, which for an image can be long
+ * after the call that produced it returned.
+ */
+using any_device_resource = ::cuda::mr::any_resource<device_accessible>;
+
+/**
+ * @brief A resource that allocates with `cudaMalloc` and frees with `cudaFree`.
+ *
+ * This is what cuCIM did before memory resources existed, expressed as a
+ * resource so that the default path stays byte-for-byte what it was and only
+ * callers who supply their own resource see different behaviour.
+ *
+ * Stateless, so it can be stored by value and compares equal to every other
+ * instance. Being stateless is also why it needs no global: there is nothing
+ * to share.
+ *
+ * The stream-ordered overloads ignore the stream, since `cudaMalloc` and
+ * `cudaFree` synchronize the whole device rather than a single stream. That
+ * makes them correct but not stream-ordered, which is the reason a caller
+ * wanting real asynchrony should pass a pool resource instead.
+ */
+class EXPORT_VISIBLE CudaMemoryResource
+{
+public:
+    void* allocate_sync(std::size_t bytes, std::size_t alignment);
+    void deallocate_sync(void* ptr, std::size_t bytes, std::size_t alignment);
+
+    void* allocate(::cuda::stream_ref stream, std::size_t bytes, std::size_t alignment);
+    void deallocate(::cuda::stream_ref stream, void* ptr, std::size_t bytes, std::size_t alignment);
+
+    bool operator==(const CudaMemoryResource&) const noexcept
+    {
+        return true;
+    }
+    bool operator!=(const CudaMemoryResource&) const noexcept
+    {
+        return false;
+    }
+
+    friend void get_property(const CudaMemoryResource&, device_accessible) noexcept
+    {
+    }
+};
+
+/**
+ * @brief Where device memory comes from, and the stream that orders it.
+ *
+ * Deliberately not a singleton and never consulted from global scope. RAPIDS
+ * guidance is to avoid a process-wide "current" resource, because it makes the
+ * owner of a buffer impossible to determine locally and is being removed from
+ * RMM; instead an instance of this is passed to whatever needs to allocate, and
+ * carried by objects that own device memory so they can free it through the
+ * same resource later.
+ *
+ * The resource is held by value in an owning `any_device_resource` rather than
+ * as a bare `device_resource_ref`. A reference would be the cheaper choice but
+ * would let a handle outlive what it points at, which is easy to do here since
+ * handles get copied into images that may outlive the call that created them.
+ *
+ * Because it owns, constructing a handle *copies* the resource, and so does
+ * copying a handle. For the usual case that is what you want: callers wrap a
+ * reference to their allocator, such as an `rmm::device_async_resource_ref`,
+ * and copying a reference still points at the one underlying pool. Wrapping a
+ * stateful allocator *by value* instead would duplicate it, and the copy the
+ * handle owns is then the one that serves every request.
+ *
+ * Default construction yields the pre-existing behaviour: `cudaMalloc` and
+ * `cudaFree` on the default stream.
+ */
+class EXPORT_VISIBLE DeviceResources
+{
+public:
+    /// Raw `cudaMalloc`/`cudaFree` on the default stream.
+    DeviceResources();
+
+    /// Allocate from *resource*, ordering allocations on *stream*.
+    explicit DeviceResources(any_device_resource resource,
+                             ::cuda::stream_ref stream = ::cuda::stream_ref{ cudaStream_t{} });
+
+    /**
+     * @brief Allocate *bytes* of device memory.
+     *
+     * @return pointer to device memory, or nullptr when *bytes* is 0.
+     * @throws std::bad_alloc if the resource cannot satisfy the request.
+     */
+    void* allocate(std::size_t bytes);
+
+    /**
+     * @brief Return memory obtained from allocate().
+     *
+     * *bytes* must be the size originally requested; pool resources need it to
+     * return the block to the right free list, unlike `cudaFree`.
+     */
+    void deallocate(void* ptr, std::size_t bytes) noexcept;
+
+    /**
+     * @brief Borrowed reference, for handing to APIs that take a resource.
+     *
+     * Non-const because a `resource_ref` allocates through the resource it
+     * refers to, so CCCL will only build one from a non-const resource.
+     */
+    device_resource_ref resource() noexcept;
+
+    ::cuda::stream_ref stream() const noexcept
+    {
+        return stream_;
+    }
+
+private:
+    any_device_resource resource_;
+    ::cuda::stream_ref stream_;
+};
+
+/**
+ * @brief move_raster_from_host() allocating from *resources*.
+ *
+ * Same contract as the overload in memory_manager.h, except that the device
+ * memory written to *target* comes from *resources*. The caller therefore has
+ * to free it through the same resource, so it must keep *resources* alive at
+ * least as long as the buffer.
+ *
+ * Declared here rather than beside the original in memory_manager.h because
+ * that header is included throughout the plugins, which would then all need
+ * CCCL on their include path.
+ *
+ * EXPORT_VISIBLE rather than CUCIM_API: the latter adds `extern "C"`, which
+ * would both rule out overloading the existing name and be wrong for a
+ * function taking a C++ reference.
+ */
+EXPORT_VISIBLE bool move_raster_from_host(void** target,
+                                          size_t size,
+                                          const cucim::io::Device& dst_device,
+                                          DeviceResources& resources);
+
+/**
+ * @brief move_raster_from_device() freeing through *resources*.
+ *
+ * *resources* must be the resource that allocated `*target`; this releases it
+ * after copying to the host. Passing a different resource is undefined, which
+ * is why the plain overload in memory_manager.h still exists for buffers that
+ * came from raw `cudaMalloc` (notably anything allocated inside a plugin).
+ */
+EXPORT_VISIBLE bool move_raster_from_device(void** target,
+                                            size_t size,
+                                            const cucim::io::Device& dst_device,
+                                            DeviceResources& resources);
+
+} // namespace cucim::memory
+
+#endif // CUCIM_MEMORY_DEVICE_RESOURCES_H
--- a/cpp/src/cache/image_cache_per_process.cpp
+++ b/cpp/src/cache/image_cache_per_process.cpp
@@ -46,8 +46,9 @@ struct PerProcessImageCacheItem
 PerProcessImageCacheValue::PerProcessImageCacheValue(void* data,
                                                      uint64_t size,
                                                      void* user_obj,
-                                                     const cucim::io::DeviceType device_type)
-    : ImageCacheValue(data, size, user_obj, device_type){};
+                                                     const cucim::io::DeviceType device_type,
+                                                     cucim::memory::DeviceResources device_resources)
+    : ImageCacheValue(data, size, user_obj, device_type), device_resources_(std::move(device_resources)){};
 
 PerProcessImageCacheValue::~PerProcessImageCacheValue()
 {
@@ -59,8 +60,10 @@ PerProcessImageCacheValue::~PerProcessImageCacheValue()
             cucim_free(data);
             break;
         case io::DeviceType::kCUDA: {
-            cudaError_t cuda_status;
-            CUDA_TRY(cudaFree(data));
+            // Returned to the resource that allocated it, not to cudaFree: a
+            // pool resource has to get its block back to stay a pool, and
+            // `size` is what tells it which block this was.
+            device_resources_.deallocate(data, size);
             break;
         }
         case io::DeviceType::kCUDAHost:
@@ -103,7 +106,15 @@ std::shared_ptr<ImageCacheValue> PerProcessImageCache::create_value(void* data,
                                                                     uint64_t size,
                                                                     const cucim::io::DeviceType device_type)
 {
-    return std::make_shared<PerProcessImageCacheValue>(data, size, nullptr, device_type);
+    // The value keeps the resource so it can free through it later, which is
+    // the only way the pairing stays correct if set_device_resources() swaps
+    // the resource while this entry is still cached.
+    return std::make_shared<PerProcessImageCacheValue>(data, size, nullptr, device_type, device_resources_);
+}
+
+void PerProcessImageCache::set_device_resources(cucim::memory::DeviceResources device_resources)
+{
+    device_resources_ = std::move(device_resources);
 }
 
 void* PerProcessImageCache::allocate(std::size_t n)
@@ -113,10 +124,17 @@ void* PerProcessImageCache::allocate(std::size_t n)
     case io::DeviceType::kCPU:
         return cucim_malloc(n);
     case io::DeviceType::kCUDA: {
-        cudaError_t cuda_status;
-        void* image_data_ptr = nullptr;
-        CUDA_TRY(cudaMalloc(&image_data_ptr, n));
-        return image_data_ptr;
+        // The resource throws on failure, whereas this returns nullptr to
+        // signal "no room in the cache" to callers that then decode without
+        // caching. Keep that contract rather than propagating.
+        try
+        {
+            return device_resources_.allocate(n);
+        }
+        catch (const std::bad_alloc&)
+        {
+            return nullptr;
+        }
     }
     case io::DeviceType::kCUDAHost:
     case io::DeviceType::kCUDAManaged:
--- a/cpp/src/memory/device_resources.cpp
+++ b/cpp/src/memory/device_resources.cpp
@@ -1,0 +1,105 @@
+/*
+ * SPDX-FileCopyrightText: Copyright (c) 2026, NVIDIA CORPORATION.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+#include "cucim/memory/device_resources.h"
+
+#include "cucim/profiler/nvtx3.h"
+
+#include <cuda_runtime.h>
+
+#include <new>
+
+namespace cucim::memory
+{
+
+void* CudaMemoryResource::allocate_sync(std::size_t bytes, std::size_t /*alignment*/)
+{
+    PROF_SCOPED_RANGE(PROF_EVENT_P(cucim_malloc, bytes));
+
+    if (bytes == 0)
+    {
+        return nullptr;
+    }
+
+    // cudaMalloc already satisfies any alignment cuCIM asks for (256 bytes),
+    // so the requested alignment needs no extra handling.
+    void* ptr = nullptr;
+    if (cudaMalloc(&ptr, bytes) != cudaSuccess)
+    {
+        // Clear the sticky error so the next CUDA call is not blamed for this
+        // failure, then report it the way a resource is expected to.
+        static_cast<void>(cudaGetLastError());
+        throw std::bad_alloc();
+    }
+    return ptr;
+}
+
+void CudaMemoryResource::deallocate_sync(void* ptr, std::size_t /*bytes*/, std::size_t /*alignment*/)
+{
+    PROF_SCOPED_RANGE(PROF_EVENT(cucim_free));
+
+    if (ptr != nullptr)
+    {
+        static_cast<void>(cudaFree(ptr));
+    }
+}
+
+void* CudaMemoryResource::allocate(::cuda::stream_ref /*stream*/, std::size_t bytes, std::size_t alignment)
+{
+    // cudaMalloc synchronizes the device, so it is trivially ordered with
+    // respect to any stream and the argument can be ignored.
+    return allocate_sync(bytes, alignment);
+}
+
+void CudaMemoryResource::deallocate(::cuda::stream_ref /*stream*/,
+                                    void* ptr,
+                                    std::size_t bytes,
+                                    std::size_t alignment)
+{
+    deallocate_sync(ptr, bytes, alignment);
+}
+
+DeviceResources::DeviceResources()
+    : resource_(CudaMemoryResource{}), stream_(::cuda::stream_ref{ cudaStream_t{} })
+{
+}
+
+DeviceResources::DeviceResources(any_device_resource resource, ::cuda::stream_ref stream)
+    : resource_(std::move(resource)), stream_(stream)
+{
+}
+
+void* DeviceResources::allocate(std::size_t bytes)
+{
+    if (bytes == 0)
+    {
+        return nullptr;
+    }
+    return resource_.allocate(stream_, bytes, kDefaultAlignment);
+}
+
+void DeviceResources::deallocate(void* ptr, std::size_t bytes) noexcept
+{
+    if (ptr == nullptr)
+    {
+        return;
+    }
+    // Deallocation runs from destructors, where letting an exception escape
+    // would terminate the process; a leak is the better failure here.
+    try
+    {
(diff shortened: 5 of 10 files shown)


gigony (commented):

**AI-assisted review.** Sharing 4 AI-assisted inline findings from the prepared review report.

Reviewed commit: `3a3179716af01d804198454d83702754fa3e8342`. The PR description was read before the diff; the current head and inline locations were rechecked before publication.

**Validation and limitations:** Validation during report preparation checked pinned CCCL headers, a host ownership/provenance model using the PR's TrackingResource, and the actual movement helper bodies with CUDA call-tracing stubs. These are not GPU or full CCCL/cuCIM runtime reproductions. Proposed patches passed git apply --check; none was applied. Full builds, Catch2, installed-consumer, async-pool, and GPU integration tests remain unrun.

<details>

### Scope and method

Ran `gh pr checkout 1147`, read `gh pr view 1147`, then reviewed `gh pr diff 1147 --color never`. Reviewed all ten changed files, related cache callers, existing raster ownership, and installation/export rules. Also read the pinned CCCL v3.4.2 resource and stream headers through the GitHub API rather than assuming older CCCL semantics.

The description deliberately defers plugin output-allocation ownership and the C ABI redesign. Those deferred features are not reported as implementation defects. The packaging gap is acknowledged in the description, but remains an actionable problem for the newly installed public header.

All locations refer to the reviewed head. Relevant PR changes are excerpts, not standalone patches. Suggested fixes were checked for textual applicability without applying them; they are not CUDA-validated implementations.

### Additional integration checks

- The legacy NvJpegProcessor allocates `tile_raster_nbytes_` but records `raw_cuda_inputs_len_[i]` as the cache value size (`cpp/plugins/cucim.kit.cuslide/src/cuslide/loader/nvjpeg_processor.cpp:317-324`). That is a pre-existing size mismatch, not a separate new-diff finding here; audit it before enabling a size-sensitive custom pool in that path. The new sized-deallocation assumption makes exact allocation provenance important.
- Validate async allocation/use/deallocation ordering through any cache/decoder consumer that receives custom resources, not just the movement helpers.
- Verify actual allocator lifetime requirements for borrowed RMM references and CUDA streams.
- Confirm requested alignment with a real allocator, not only that the alignment argument was forwarded to the host-only tracking resource.

</details>

cpp/include/cucim/memory/device_resources.h
```diff
@@ -0,0 +1,211 @@
...
+    device_resource_ref resource() noexcept;
+
+    ::cuda::stream_ref stream() const noexcept
+    {
+        return stream_;
+    }
+
+private:
+    any_device_resource resource_;
+    ::cuda::stream_ref stream_;
```
gigony, line 172: ### [P1] Preserve allocator identity when copying a resource handle

**AI-assisted review comment (finding 1/4).**

An implicit copy of this handle copies the concrete allocator held in `any_resource`, rather than retaining one allocating instance. Cache values copy the handle after allocation, so resources with independent per-copy bookkeeping can deallocate through an inequivalent instance. The current copy test returns a local by NRVO/move and does not exercise a real copy. Please share the owned resource across copies and provide an explicit borrowing path for `device_resource_ref`; promotion of a type-erased reference to the pinned CCCL owning wrapper deep-copies the referenced object.

**Validation:** Validation during report preparation checked pinned CCCL headers, a host ownership/provenance model using the PR's TrackingResource, and the actual movement helper bodies with CUDA call-tracing stubs. These are not GPU or full CCCL/cuCIM runtime reproductions. Proposed patches passed git apply --check; none was applied. Full builds, Catch2, installed-consumer, async-pool, and GPU integration tests remain unrun.

<details>

### Locations

- `cpp/include/cucim/memory/device_resources.h:131-172`: implicitly copyable handle contains `any_device_resource` by value.
- `cpp/src/cache/image_cache_per_process.cpp:109-112`: copies that handle into every cache value.
- `cpp/tests/test_device_resources.cpp:241-245`: the supposed copy test returns a local by NRVO or move instead of exercising an actual copy.

### Root cause and impact

`cuda::mr::any_resource` is a value-owning type-erased wrapper, not shared ownership of one allocator instance. Its copy constructor copies the concrete resource. The cache therefore allocates using one concrete resource, then copies the resource when constructing the value that will later deallocate the pointer.

For an allocator whose copies have independent allocation bookkeeping, the deallocation goes to an inequivalent allocator. A per-instance pool can retain stale allocation records, reclaim a block twice, reject a foreign pointer, or release storage while another handle still uses it. Stateless `CudaMemoryResource` hides the defect because all its instances delegate to the same raw CUDA allocation/free operations.

The documentation acknowledges that stateful allocators can be duplicated, but this does not satisfy the stronger promise that a copied handle or cached value frees through the resource that allocated its buffer.

There is also an important reference-conversion detail: the pinned CCCL documentation explicitly describes construction of an owning resource from a type-erased reference as a **deep copy of the referenced object**. Merely passing a `resource_ref` through the existing `any_device_resource` parameter is not a reliable way to preserve borrowing semantics.

Verified dependency reference: [CCCL v3.4.2 any_resource.h](https://github.com/NVIDIA/cccl/blob/v3.4.2/libcudacxx/include/cuda/__memory_resource/any_resource.h), particularly the copy/deep-copy constructor documentation at lines 498-534 and the value-owning implementation.

### Reproduction / validation

A minimal regression using the existing test resource is:

```cpp
DeviceResources original{ any_device_resource{ TrackingResource{} } };
DeviceResources copy = original;
void* ptr = original.allocate(2048);
copy.deallocate(ptr, 2048);
```

With value-copy semantics, `copy` was created before the allocation and does not contain its bookkeeping. The existing `TrackingResource` reports a foreign pointer, while the original tracker still has an outstanding allocation record.

During report preparation, the host harness compiled the actual `TrackingResource` class extracted from the PR into a host-side ownership model. It reproduced that bookkeeping mismatch; a shared-owner variant updated the original tracker correctly. This was not a full CCCL/cuCIM runtime test.

The checked-in test does not expose this: `return original;` either elides the transfer or implicitly moves the local. It need not invoke the copy constructor at all.

### Suggested fix

Share ownership of the concrete owning wrapper across handle copies, and keep a reference to that same wrapper for dispatch. Add a distinct `device_resource_ref` constructor that genuinely borrows rather than first converting to an owning wrapper.

**Lifetime contract:** The borrowed constructor cannot extend the allocator's or stream's lifetime. Its caller must keep both alive until all uses and deallocations finish. The owning constructor shares its owned allocator across handle copies. Update the class documentation to make these two contracts explicit when integrating this proposal.

```diff
diff --git a/cpp/include/cucim/memory/device_resources.h b/cpp/include/cucim/memory/device_resources.h
--- a/cpp/include/cucim/memory/device_resources.h
+++ b/cpp/include/cucim/memory/device_resources.h
@@ -13,6 +13,7 @@
 #include <cuda/stream_ref>
 
 #include <cstddef>
+#include <memory>
 
 /**
  * Device memory allocation through a caller-supplied CCCL memory resource.
@@ -136,6 +137,8 @@
 
     /// Allocate from *resource*, ordering allocations on *stream*.
     explicit DeviceResources(any_device_resource resource,
+                             ::cuda::stream_ref stream = ::cuda::stream_ref{ cudaStream_t{} });
+    explicit DeviceResources(device_resource_ref resource,
                              ::cuda::stream_ref stream = ::cuda::stream_ref{ cudaStream_t{} });
 
     /**
@@ -168,7 +171,8 @@
     }
 
 private:
-    any_device_resource resource_;
+    std::shared_ptr<any_device_resource> resource_owner_;
+    device_resource_ref resource_;
     ::cuda::stream_ref stream_;
 };
 
diff --git a/cpp/src/memory/device_resources.cpp b/cpp/src/memory/device_resources.cpp
--- a/cpp/src/memory/device_resources.cpp
+++ b/cpp/src/memory/device_resources.cpp
@@ -62,12 +62,18 @@
 }
 
 DeviceResources::DeviceResources()
-    : resource_(CudaMemoryResource{}), stream_(::cuda::stream_ref{ cudaStream_t{} })
+    : DeviceResources(any_device_resource{ CudaMemoryResource{} })
 {
 }
 
 DeviceResources::DeviceResources(any_device_resource resource, ::cuda::stream_ref stream)
-    : resource_(std::move(resource)), stream_(stream)
+    : resource_owner_(std::make_shared<any_device_resource>(std::move(resource))),
+      resource_(*resource_owner_), stream_(stream)
+{
+}
+
+DeviceResources::DeviceResources(device_resource_ref resource, ::cuda::stream_ref stream)
+    : resource_(resource), stream_(stream)
 {
 }
 
@@ -99,7 +105,7 @@
 
 device_resource_ref DeviceResources::resource() noexcept
 {
-    return device_resource_ref{ resource_ };
+    return resource_;
 }
 
 } // namespace cucim::memory
diff --git a/cpp/tests/test_device_resources.cpp b/cpp/tests/test_device_resources.cpp
--- a/cpp/tests/test_device_resources.cpp
+++ b/cpp/tests/test_device_resources.cpp
@@ -240,8 +240,9 @@
 
     DeviceResources copy = [&] {
         DeviceResources original{ any_device_resource{ TrackingResource{} } };
+        DeviceResources retained = original;
         ptr = original.allocate(nbytes);
-        return original;
+        return retained;
     }();
 
     REQUIRE(ptr != nullptr);
@@ -271,3 +272,15 @@
     REQUIRE(ptr != nullptr);
     ref.deallocate(resources.stream(), ptr, 256, cucim::memory::kDefaultAlignment);
 }
+
+TEST_CASE("DeviceResources borrows resource_ref without copying the allocator", "[test_device_resources.cpp]")
+{
+    TrackingResource tracker;
+    DeviceResources resources{ device_resource_ref{ tracker } };
+    void* ptr = resources.allocate(256);
+    REQUIRE(tracker.allocate_calls == 1);
+    resources.deallocate(ptr, 256);
+    REQUIRE(tracker.deallocate_calls == 1);
+    REQUIRE(tracker.outstanding.empty());
+    REQUIRE_FALSE(tracker.saw_foreign_pointer);
+}
```

### Regression coverage

Test explicit copies both before and after allocation, copy assignment, destruction of the original handle, and values surviving cache eviction/destruction. Check resource identity and outstanding allocation state, not merely successful pointer deletion. Add an actual supported RMM resource-reference integration case rather than relying only on concept assertions.

</details>

cpp/src/memory/memory_manager.cpp
```diff
@@ -73,15 +75,15 @@ CUCIM_API bool move_raster_from_host(void** target, size_t size, const cucim::io
...
-        void* cuda_mem;
-        CUDA_TRY(cudaMalloc(&cuda_mem, size));
-        if (cuda_status)
-        {
-            throw std::bad_alloc();
-        }
+        // Throws std::bad_alloc on failure, which is what the hand-rolled
+        // check below used to raise.
+        void* cuda_mem = resources.allocate(size);
         CUDA_TRY(cudaMemcpy(cuda_mem, host_mem, size, cudaMemcpyHostToDevice));
```
gigony, line 81: ### [P1] Order raster copies on the resource's stream

**AI-assisted review comment (finding 2/4).**

Allocation now runs on `resources.stream()`, but this copy still uses the default stream. With an async allocator on a nonblocking stream, allocation readiness is not ordered before the copy. The inverse helper likewise ignores a producer on the resource stream. Please enqueue both movement copies on the supplied stream and complete them before freeing host input or returning a completed host result.

**Validation:** Validation during report preparation checked pinned CCCL headers, a host ownership/provenance model using the PR's TrackingResource, and the actual movement helper bodies with CUDA call-tracing stubs. These are not GPU or full CCCL/cuCIM runtime reproductions. Proposed patches passed git apply --check; none was applied. Full builds, Catch2, installed-consumer, async-pool, and GPU integration tests remain unrun.

<details>

### Root cause and impact

`resources.allocate()` now permits a stream-ordered allocator and dispatches allocation on `resources.stream()`. The subsequent `cudaMemcpy` still uses the default stream. A nonblocking allocation stream does not acquire an ordering dependency merely because the following API call is spelled without `Async`.

For a `cudaMallocAsync`-backed resource on a nonblocking stream, the default-stream transfer can use the pointer before its allocation is ready. In the other direction, `move_raster_from_device` can copy data to the host before a producer on the resource stream has completed. The default raw `cudaMalloc` resource and the synchronous host-only tracking resource used in the tests conceal the problem.

### Validation

Compiled the actual two helper bodies with CUDA API stubs that record the chosen stream. A supplied non-default stream numbered 7 produced:

```text
original: allocate:7 copy:0 copy:0 deallocate:7
proposed: allocate:7 copy:7 sync:7 copy:7 sync:7 deallocate:7
```

This establishes which stream the implementation calls, not a hardware reproduction of the race.

### Suggested fix

Enqueue copies on the resource stream and synchronize that stream before freeing host source memory, exposing a completed host result, or returning from these synchronous movement helpers. Preserve the existing failure cleanup and do not overwrite an earlier copy error with a synchronization result.

```diff
diff --git a/cpp/src/memory/memory_manager.cpp b/cpp/src/memory/memory_manager.cpp
--- a/cpp/src/memory/memory_manager.cpp
+++ b/cpp/src/memory/memory_manager.cpp
@@ -78,7 +78,11 @@
         // Throws std::bad_alloc on failure, which is what the hand-rolled
         // check below used to raise.
         void* cuda_mem = resources.allocate(size);
-        CUDA_TRY(cudaMemcpy(cuda_mem, host_mem, size, cudaMemcpyHostToDevice));
+        CUDA_TRY(cudaMemcpyAsync(cuda_mem, host_mem, size, cudaMemcpyHostToDevice, resources.stream().get()));
+        if (cuda_status == cudaSuccess)
+        {
+            CUDA_TRY(cudaStreamSynchronize(resources.stream().get()));
+        }
         if (cuda_status)
         {
             // Give the block back before unwinding, or the failed transfer
@@ -107,7 +111,11 @@
         cudaError_t cuda_status;
         void* cuda_mem = *target;
         void* host_mem = cucim_malloc(size);
-        CUDA_TRY(cudaMemcpy(host_mem, cuda_mem, size, cudaMemcpyDeviceToHost));
+        CUDA_TRY(cudaMemcpyAsync(host_mem, cuda_mem, size, cudaMemcpyDeviceToHost, resources.stream().get()));
+        if (cuda_status == cudaSuccess)
+        {
+            CUDA_TRY(cudaStreamSynchronize(resources.stream().get()));
+        }
         if (cuda_status)
         {
             cucim_free(host_mem);
```

### Regression coverage

Use a real async pool resource and a nonblocking CUDA stream. Queue a delayed producer before the device-to-host move and verify the returned host bytes are final. Exercise host-to-device allocation readiness, failed copies, and failed synchronization. Cross-stream producers must establish dependencies on the supplied stream before calling the helper.

This patch fixes the movement helpers, not every future cache consumer. Before wiring a custom asynchronous resource into decoder/cache paths, pass or bridge its stream through those paths and order deallocation after the final GPU use; a host-side shared pointer alone does not establish GPU execution ordering.

</details>

cpp/src/cache/image_cache_per_process.cpp
```diff
@@ -103,7 +106,15 @@ std::shared_ptr<ImageCacheValue> PerProcessImageCache::create_value(void* data,
...
+    // The value keeps the resource so it can free through it later, which is
+    // the only way the pairing stays correct if set_device_resources() swaps
+    // the resource while this entry is still cached.
+    return std::make_shared<PerProcessImageCacheValue>(data, size, nullptr, device_type, device_resources_);
+}
+
+void PerProcessImageCache::set_device_resources(cucim::memory::DeviceResources device_resources)
+{
+    device_resources_ = std::move(device_resources);
 }
```
gigony, line 118: ### [P1] Do not change allocators while an allocation is awaiting its cache value

**AI-assisted review comment (finding 3/4).**

The allocator snapshot is taken in `create_value()`, not `allocate()`. Even sequentially, `set(A); p = allocate(n); set(B); create_value(p, n, kCUDA)` records B as the deallocator for A's pointer. The setter also races unsynchronized handle reads. Please retain allocation provenance from allocation through value construction. The patch below is a containment alternative that restricts configuration to before the first successful allocation; it deliberately changes the proposed hot-swapping contract.

**Validation:** Validation during report preparation checked pinned CCCL headers, a host ownership/provenance model using the PR's TrackingResource, and the actual movement helper bodies with CUDA call-tracing stubs. These are not GPU or full CCCL/cuCIM runtime reproductions. Proposed patches passed git apply --check; none was applied. Full builds, Catch2, installed-consumer, async-pool, and GPU integration tests remain unrun.

<details>

### Root cause and impact

The resource snapshot is taken when `create_value()` runs, not when `allocate()` runs. Those are separate API calls. The following sequence selects the wrong deallocator even without concurrent execution:

```text
set_device_resources(A)
p = allocate(n)                  # allocation belongs to A
set_device_resources(B)
v = create_value(p, n, kCUDA)     # records B, not A
v is destroyed                   # deallocates A's pointer through B
```

Concurrent setters also race the unsynchronized handle reads/assignment. Protecting only the assignment, or checking whether the cache currently contains entries, is insufficient: an allocated but not-yet-wrapped pointer is not an entry yet.

Sharing resource ownership as proposed in finding 1 does not repair this separate provenance error. The A/B routing failure was checked with a host model using the PR's tracking resource.

### Suggested containment fix

For this foundational PR, a small safe alternative is to permit resource configuration **only before the first successful device allocation** and serialize resource access. This deliberately narrows the proposed hot-swapping contract; update the setter documentation accordingly if chosen.

If runtime switching is a hard requirement, do not adopt this restriction. Instead redesign allocation to return/store an allocation record containing the pointer, original size, resource handle, and stream, and pass that exact record into the cache value. The raw pointer plus whatever resource is current at wrapping time is not sufficient.

```diff
diff --git a/cpp/src/cache/image_cache_per_process.h b/cpp/src/cache/image_cache_per_process.h
--- a/cpp/src/cache/image_cache_per_process.h
+++ b/cpp/src/cache/image_cache_per_process.h
@@ -11,6 +11,7 @@
 
 #include <libcuckoo/cuckoohash_map.hh>
 #include <memory>
+#include <mutex>
 #include <array>
 
 namespace std
@@ -129,6 +130,8 @@
 
     /// Where device tiles come from. Defaults to cudaMalloc/cudaFree.
     cucim::memory::DeviceResources device_resources_;
+    std::mutex resource_mutex_;
+    bool has_device_allocations_ = false;
 
     std::atomic<uint64_t> size_nbytes_ = 0; /// size of cache memory used
     uint64_t capacity_nbytes_ = 0; /// size of cache memory allocated
diff --git a/cpp/src/cache/image_cache_per_process.cpp b/cpp/src/cache/image_cache_per_process.cpp
--- a/cpp/src/cache/image_cache_per_process.cpp
+++ b/cpp/src/cache/image_cache_per_process.cpp
@@ -10,6 +10,7 @@
 #include "cucim/util/cuda.h"
 
 #include <fmt/format.h>
+#include <stdexcept>
 
 namespace std
 {
@@ -109,11 +110,17 @@
     // The value keeps the resource so it can free through it later, which is
     // the only way the pairing stays correct if set_device_resources() swaps
     // the resource while this entry is still cached.
+    std::lock_guard<std::mutex> guard(resource_mutex_);
     return std::make_shared<PerProcessImageCacheValue>(data, size, nullptr, device_type, device_resources_);
 }
 
 void PerProcessImageCache::set_device_resources(cucim::memory::DeviceResources device_resources)
 {
+    std::lock_guard<std::mutex> guard(resource_mutex_);
+    if (has_device_allocations_)
+    {
+        throw std::logic_error("device resources must be set before the first device allocation");
+    }
     device_resources_ = std::move(device_resources);
 }
 
@@ -129,7 +136,10 @@
         // caching. Keep that contract rather than propagating.
         try
         {
-            return device_resources_.allocate(n);
+            std::lock_guard<std::mutex> guard(resource_mutex_);
+            void* ptr = device_resources_.allocate(n);
+            has_device_allocations_ = has_device_allocations_ || ptr != nullptr;
+            return ptr;
         }
         catch (const std::bad_alloc&)
         {
```

### Regression coverage

Test resource changes before allocation, immediately after allocation but before `create_value`, after cache eviction, and concurrently with allocation/value creation. With the containment policy, the latter changes must reject deterministically without changing the resource. With an allocation-record design, every free must reach the original allocator with the original size and ordered stream.

</details>

cpp/CMakeLists.txt
```diff
@@ -157,6 +158,24 @@ target_link_libraries(${CUCIM_PACKAGE_NAME}
...
+# silently disappears. BEFORE is what keeps the two copies from being mixed.
+#
+# PUBLIC because <cucim/memory/device_resources.h> exposes cuda::mr types.
+# BUILD_INTERFACE only: installed consumers of that header still need CCCL
+# exported or a cuda-cccl dependency added, which is not done yet.
+target_include_directories(${CUCIM_PACKAGE_NAME}
+        BEFORE
+        PUBLIC
+            $<BUILD_INTERFACE:${deps-cccl_SOURCE_DIR}/libcudacxx/include>
+        )
```
gigony, line 177: ### [P2] Export the CCCL dependency required by the installed public header

**AI-assisted review comment (finding 4/4).**

This is the packaging gap already acknowledged in the description: the public `device_resources.h` is installed, but this CCCL include path disappears from the installed target. On a supported toolkit whose bundled libcudacxx lacks the required API, an installed-SDK consumer cannot compile the new header. Please export a compatible external CCCL dependency or install the pinned headers and their usage requirements before publishing this API.

**Validation:** Validation during report preparation checked pinned CCCL headers, a host ownership/provenance model using the PR's TrackingResource, and the actual movement helper bodies with CUDA call-tracing stubs. These are not GPU or full CCCL/cuCIM runtime reproductions. Proposed patches passed git apply --check; none was applied. Full builds, Catch2, installed-consumer, async-pool, and GPU integration tests remain unrun.

<details>

### Root cause and impact

The top-level install rule installs the entire cuCIM public include directory, including `device_resources.h`. That header unconditionally includes CCCL headers. However:

- The new CCCL include path is build-interface-only and disappears from the installed target.
- The fetched CCCL headers are not installed with the existing third-party headers.
- `cucim-config.cmake.in` finds the CUDA toolkit but does not find/export a compatible standalone CCCL package.
- The recipe does not add the independent CCCL version needed by the new API.

An external consumer using the installed SDK and including `cucim/memory/device_resources.h` therefore cannot compile on a supported toolkit whose bundled libcudacxx lacks the new `cuda::mr` API. This is not limited to optional output-allocation work: the public header is already installed by this PR.

The author explicitly acknowledges this gap. It should still be resolved before publishing the new API.

### Suggested fix

One approach consistent with the existing vendored-header installation is to install the pinned libcudacxx headers with the other third-party headers, expose that directory early in installed include resolution, and propagate the required CUDA headers/runtime target. Ensure the header-only target can also resolve the new public header.

```diff
diff --git a/CMakeLists.txt b/CMakeLists.txt
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -196,6 +196,7 @@
 install(DIRECTORY
             ${deps-fmt_SOURCE_DIR}/include/
             ${deps-nvtx3_SOURCE_DIR}/c/include/
+            ${deps-cccl_SOURCE_DIR}/libcudacxx/include/
         DESTINATION
             ${CMAKE_INSTALL_INCLUDEDIR}/${CUCIM_PACKAGE_NAME}/3rdparty)
 
diff --git a/cpp/CMakeLists.txt b/cpp/CMakeLists.txt
--- a/cpp/CMakeLists.txt
+++ b/cpp/CMakeLists.txt
@@ -147,6 +147,7 @@
             Threads::Threads # -lpthread
             $<BUILD_INTERFACE:deps::fmt>
             $<INSTALL_INTERFACE:cucim::fmt-header-only>
+            $<INSTALL_INTERFACE:CUDA::cudart_static>
         PRIVATE
             CUDA::cudart_static
             deps::abseil
@@ -174,6 +175,7 @@
         BEFORE
         PUBLIC
             $<BUILD_INTERFACE:${deps-cccl_SOURCE_DIR}/libcudacxx/include>
+            $<INSTALL_INTERFACE:${CMAKE_INSTALL_INCLUDEDIR}/${CUCIM_PACKAGE_NAME}/3rdparty>
         )
 
 if (CUCIM_STATIC_GDS)
@@ -214,8 +216,10 @@
         CUCIM_HEADER_ONLY=1
         )
 target_compile_features(${CUCIM_PACKAGE_NAME}-header-only INTERFACE ${CUCIM_REQUIRED_FEATURES})
+target_link_libraries(${CUCIM_PACKAGE_NAME}-header-only INTERFACE CUDA::cudart_static)
 target_include_directories(${CUCIM_PACKAGE_NAME}-header-only
         INTERFACE
+            $<BUILD_INTERFACE:${deps-cccl_SOURCE_DIR}/libcudacxx/include>
             $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/include>
             $<INSTALL_INTERFACE:${CMAKE_INSTALL_INCLUDEDIR}>
             $<INSTALL_INTERFACE:${CMAKE_INSTALL_INCLUDEDIR}/${CUCIM_PACKAGE_NAME}/3rdparty>
```

Alternatively, declare a compatible external CCCL package dependency and export its imported target instead of vendoring. Do not export the build machine's fetched source path. Whichever packaging policy is selected, include the required CCCL license notices and update the now-stale “packaging outstanding” documentation.

### Regression coverage

Install to a clean prefix, make the source/build tree unavailable, and compile/link a separate consumer that only uses `find_package(cucim CONFIG REQUIRED)` and the exported target. Include `device_resources.h`, construct a default handle, and exercise a custom resource. Check supported CUDA versions and verify no fetched build-tree paths leak into the installed target.

</details>

cpp/cmake/deps/cccl.cmake
```diff
@@ -0,0 +1,43 @@
+#
+# cmake-format: off
+# SPDX-FileCopyrightText: Copyright (c) 2026, NVIDIA CORPORATION.
+# SPDX-License-Identifier: Apache-2.0
+# cmake-format: on
+#
+
+if (NOT TARGET deps::cccl)
```
miscco, line 8: Shouldnt you reuse the cccl script from rapids-cmake?

cpp/include/cucim/memory/device_resources.h
```diff
@@ -0,0 +1,211 @@
...
+ */
+
+#ifndef CUCIM_MEMORY_DEVICE_RESOURCES_H
+#define CUCIM_MEMORY_DEVICE_RESOURCES_H
+
+#include "cucim/core/framework.h"
+#include "cucim/io/device.h"
+
+#include <cuda/memory_resource>
+#include <cuda/stream_ref>
```
miscco, line 13: The header is deprecated please use 
```suggestion
#include <cuda/stream>
```

cpp/src/cache/image_cache_per_process.cpp
```diff
@@ -46,8 +46,9 @@ struct PerProcessImageCacheItem
 PerProcessImageCacheValue::PerProcessImageCacheValue(void* data,
                                                      uint64_t size,
                                                      void* user_obj,
-                                                     const cucim::io::DeviceType device_type)
-    : ImageCacheValue(data, size, user_obj, device_type){};
+                                                     const cucim::io::DeviceType device_type,
+                                                     cucim::memory::DeviceResources device_resources)
```
miscco, line 50: Critical: You are passing a `DeviceResources` by value

This will copy the stored device resource every time you do. I am not sure this is what you want.

In general you should create one global pool / resource and then pass a reference to that around

cpp/cmake/deps/cccl.cmake
bdice: Yes. We have #859 tracking this.

cpp/include/cucim/memory/device_resources.h
```diff
@@ -0,0 +1,211 @@
...
+{
+
+/**
+ * @brief The set of CCCL properties cuCIM requires of a device memory resource.
+ *
+ * Device-accessible only. cuCIM hands the resulting pointers to CUDA kernels,
+ * nvImageCodec and `__cuda_array_interface__` consumers, none of which can use
+ * host-only memory.
+ */
+using device_accessible = ::cuda::mr::device_accessible;
```
bdice, line 36: I think this is equivalent.
```suggestion
using ::cuda::mr::device_accessible;
```

cpp/include/cucim/memory/device_resources.h
```diff
@@ -0,0 +1,211 @@
...
+ * Stateless, so it can be stored by value and compares equal to every other
+ * instance. Being stateless is also why it needs no global: there is nothing
+ * to share.
+ *
+ * The stream-ordered overloads ignore the stream, since `cudaMalloc` and
+ * `cudaFree` synchronize the whole device rather than a single stream. That
+ * makes them correct but not stream-ordered, which is the reason a caller
+ * wanting real asynchrony should pass a pool resource instead.
+ */
+class EXPORT_VISIBLE CudaMemoryResource
```
bdice, line 83: Can you use CCCL’s async resource as the default? I understand it doesn’t use cudaMalloc but is that really a hard requirement for the library (or did AI just make that up in an attempt to match the current behavior)? We are moving away from cudaMalloc in RMM. https://github.com/rapidsai/rmm/issues/2497

Something more like: https://developer.nvidia.com/blog/cccl-runtime-a-modern-c-runtime-for-cuda/

```cpp
  // 1. Devices and streams                                                       
  cuda::device_ref device = cuda::devices[0];
  cuda::stream stream{device};                                     
 
  // 2. Memory allocation                                                                 
  auto pool = cuda::device_default_memory_pool(device);
```

cpp/include/cucim/memory/device_resources.h
```diff
@@ -0,0 +1,211 @@
...
+
+    friend void get_property(const CudaMemoryResource&, device_accessible) noexcept
+    {
+    }
+};
+
+/**
+ * @brief Where device memory comes from, and the stream that orders it.
+ *
+ * Deliberately not a singleton and never consulted from global scope. RAPIDS
```
bdice, line 109: These comments are overly verbose. They reflect a point-in-time from other libraries. We’re retiring RAPIDS as a brand and removing RMM from most direct call paths in our libraries (to reduce central points of instability). Please critically evaluate what we want to say, rather than letting AI declare these as a permanent fact.

cpp/include/cucim/memory/device_resources.h
```diff
@@ -0,0 +1,211 @@
...
+ */
+class EXPORT_VISIBLE DeviceResources
+{
+public:
+    /// Raw `cudaMalloc`/`cudaFree` on the default stream.
+    DeviceResources();
+
+    /// Allocate from *resource*, ordering allocations on *stream*.
+    explicit DeviceResources(any_device_resource resource,
+                             ::cuda::stream_ref stream = ::cuda::stream_ref{ cudaStream_t{} });
```
bdice, line 139: More explicit:
```suggestion
                             ::cuda::stream_ref stream = ::cuda::stream_ref{ cudaStream_t{cudaStreamDefault} });
```

cpp/include/cucim/memory/device_resources.h
```diff
@@ -0,0 +1,211 @@
...
+
+    /**
+     * @brief Allocate *bytes* of device memory.
+     *
+     * @return pointer to device memory, or nullptr when *bytes* is 0.
+     * @throws std::bad_alloc if the resource cannot satisfy the request.
+     */
+    void* allocate(std::size_t bytes);
```
bdice, line 147: Please add alignment arguments to match the CCCL resource API. It is incredibly painful to wire that argument through later.

cpp/include/cucim/memory/device_resources.h
```diff
@@ -0,0 +1,211 @@
...
+
+    /**
+     * @brief Return memory obtained from allocate().
+     *
+     * *bytes* must be the size originally requested; pool resources need it to
```
bdice, line 152: This aside about pool resources probably isn’t necessary.

cpp/include/cucim/memory/device_resources.h
```diff
@@ -0,0 +1,211 @@
...
+
+    /**
+     * @brief Borrowed reference, for handing to APIs that take a resource.
+     *
+     * Non-const because a `resource_ref` allocates through the resource it
+     * refers to, so CCCL will only build one from a non-const resource.
+     */
+    device_resource_ref resource() noexcept;
+
+    ::cuda::stream_ref stream() const noexcept
```
bdice, line 165: Why does this need to hold a stream? It’s a much safer, performant, and flexible design if a container holds its resource and stream separately. That’s a better way because it lets you decouple the two, and pass an object between streams. The stream held by the container is only used for deallocation. Otherwise all methods should take a stream parameter. When an object is used on a stream, you update the stream it’s holding so that deallocation is always ordered-after the method that was called.

cpp/src/cache/image_cache_per_process.cpp
```diff
@@ -59,8 +60,10 @@ PerProcessImageCacheValue::~PerProcessImageCacheValue()
             cucim_free(data);
             break;
         case io::DeviceType::kCUDA: {
-            cudaError_t cuda_status;
-            CUDA_TRY(cudaFree(data));
+            // Returned to the resource that allocated it, not to cudaFree: a
```
bdice, line 63: These comments don’t add a lot. Remove?

cpp/src/memory/device_resources.cpp
```diff
@@ -0,0 +1,105 @@
...
+}
+
+DeviceResources::DeviceResources(any_device_resource resource, ::cuda::stream_ref stream)
+    : resource_(std::move(resource)), stream_(stream)
+{
+}
+
+void* DeviceResources::allocate(std::size_t bytes)
+{
+    if (bytes == 0)
```
bdice, line 76: I recommend not doing this early exit. Always call the base resource’s allocate method. If you put in a logger, it should log allocations of zero bytes. If you put in a resource that tracks statistics, it should track this. Etc.

cpp/src/memory/device_resources.cpp
```diff
@@ -0,0 +1,105 @@
...
+    {
+        return nullptr;
+    }
+    return resource_.allocate(stream_, bytes, kDefaultAlignment);
+}
+
+void DeviceResources::deallocate(void* ptr, std::size_t bytes) noexcept
+{
+    if (ptr == nullptr)
```
bdice, line 85: Remove, as above.

cpp/src/memory/device_resources.cpp
```diff
@@ -0,0 +1,105 @@
...
+    {
+        return;
+    }
+    // Deallocation runs from destructors, where letting an exception escape
```
bdice, line 89: Deallocation is guaranteed to be noexcept. This try/catch should be removed.

bdice (changes requested):
Partial review, hope this helps as a starting point.

jakirkham:
Retargeted from `release/26.10` to `main`. Planning on this for RAPIDS 26.12

jakirkham:
Closing and reopening to restart CI

PR #1148, opened 2026-09-07
Verify the loaded nvImageCodec version at runtime, and consolidate nvimgcodec.cmake

cdinea:
## Summary

Follow-up to @mkepa-nv's review feedback on the nvImageCodec SONAME candidates:

> sure, we can consolidate in another PR. In the next PR you can also consider adding runtime check that the correct version of nvimagecodec was loaded - I think it is good to load `libnvimgcodec.so.0` or just `libnvimgcodec.so` in case there are non others in the system, but I would verify in runtime via getProperties that the version is supported

Both parts are here: the runtime `getProperties` check, and the consolidation of the two `nvimgcodec.cmake` copies.

## Why the runtime check is the one that matters

Investigating this turned out to justify the suggestion more strongly than I expected. The configure-time header check I added in the previous PR is bypassed entirely in the default build:

- `WITH_DYNAMIC_NVIMGCODEC` defaults to **ON** (`cuslide2/CMakeLists.txt:216`), so nvImageCodec is `dlopen`'d at runtime and `deps::nvimgcodec` is **not linked at all**.
- The SONAME candidates end with `libnvimgcodec.so.0` and `libnvimgcodec.so`, which resolve to whichever major-0 build is installed.
- There are no compile-time version guards.

So the version found at configure time need not be the version loaded, and an unsupported library surfaces as a missing symbol or as wrong behaviour partway through a decode. Asking the library its own version is the only dependable check — exactly as suggested.

## What the check does

`NvImageCodecTiffParserManager`'s constructor now calls `nvimgcodecGetProperties()` before creating an instance, and rejects anything outside `>=0.9.0,<0.10.0`.

That constructor is a good fit for this: it is the single point every read path passes through exactly once (`CuImage` → `parser_open` → `TIFF` → `TiffFileParser` → `manager.instance()`), and it already reports failure by setting `status_message_` and returning, which `TiffFileParser` turns into an exception. So an unsupported install now fails at open with one clear message naming the version found and pointing at `LD_LIBRARY_PATH` and the `libnvimgcodec.so` symlinks, instead of failing later and less legibly.

Worth noting `nvimgcodecGetProperties` was already being resolved as a symbol (`nvimgcodec_tiff_parser.cpp:86`) purely for `dladdr()` path probing, and never actually called — so no new dynlink surface is involved.

## On the packed version encoding

`nvimgcodecProperties_t::version` is packed as `major * 1000 + minor * 100 + patch`, which gives the minor component only one digit. Two consequences, both handled:

- `MAKE_SEMANTIC_VERSION(0, 10, 0)` and `MAKE_SEMANTIC_VERSION(1, 0, 0)` **both equal 1000**. The range is therefore compared as packed integers rather than by unpacking major/minor — ordering is preserved, so the comparison stays correct, and both values are out of range either way.
- A diagnostic can render an out-of-range `0.10.0` as `"1.0.0"`. Commented at the formatting helper. The human-readable range in messages is a separate literal rather than derived from the constants, so it cannot be rendered misleadingly.

## Consolidating the two cmake files

`cpp/cmake/deps/nvimgcodec.cmake` turned out to be **dead code**, which makes the consolidation simpler than "merge two live files":

- Only `cuslide2/CMakeLists.txt:112` calls `superbuild_depend(nvimgcodec)`.
- cuslide2 sets `SUPERBUILD_ADDITIONAL_DEPS_DIRS` to its own `cmake/deps` (`CMakeLists.txt:31`), which is searched first.
- Its fallback `CMAKE_SUPERBUILD_DEPS_ROOT_DIR` points at **cuslide's** cmake directory, not `cpp/cmake`, so the top-level copy was unreachable even as a fallback.

So the cuslide2 copy is now the single source of truth. Before deleting the orphan I merged over the one capability the live copy lacked: resolving `$PREFIX` when `CONDA_BUILD` is set, since `CONDA_PREFIX` points at the build environment rather than the install prefix under conda-build.

I also gave the `find_package` path the upper-bound warning it was missing — it previously only had the `<0.9.0` fatal error and never called `_cucim_check_nvimgcodec_version()` — so one policy now applies however nvImageCodec is discovered. That was the duplication I flagged in the review thread; with one copy it is no longer duplicated.

This does leave the more general duplication across plugin `cmake/deps/` directories (`catch2`, `cli11`, `fmt`, `googletest`, `googlebenchmark` are copied between cuslide, cumed and the top level) untouched. nvimgcodec was the case under discussion; the rest is a larger cleanup.

## Testing

`tests/test_nvimgcodec_version.cpp` (new) pins the supported range. The range has to agree with the pin in `dependencies.yaml` and the one enforced on headers in `cmake/deps/nvimgcodec.cmake`, and nothing derives one from another, so `is_supported_nvimgcodec_version()` is exposed to make that testable. Cases:

- 0.7.0, 0.8.0 and 0.8.99 rejected — 0.8 being the case that prompted this, since it was still a SONAME candidate after the pin moved and `limit_images` was removed in 0.9.
- 0.9.0, 0.9.1 and 0.9.99 accepted.
- 0.10.0, 1.0.0 and 2.3.4 rejected.
- `NVIMGCODEC_VER` from the vendored headers is itself in range, which catches a header bump that would otherwise reject every install including the intended one.
- Ordering across the packed encoding, plus the 0.10.0/1.0.0 collision, pinned explicitly so the reason the comparison is written this way does not get lost.

Verification and its limits: the modified translation unit and the new test both compile clean with the build's own `-Werror -Wall -Wextra`. The boundary logic was additionally executed standalone against the real vendored macros and all ten cases pass. **A full cuslide2 build and the Catch2 suite were not run here (no GPU access in this environment), so CI is the real check.** The `getProperties` call itself has not been exercised against a live library — worth a look from someone with a GPU box, particularly the case where the loaded library is genuinely older.


gigony (commented):

**AI-assisted review.** No actionable correctness defect was identified in this review. There is no finding to anchor inline, so this is a summary-only review; no formal approval is implied.

Reviewed commit: `84cda22c703f1d1463dca9ffd868222e803ec5dc`. The PR description was read before the diff; the current head and inline locations were rechecked before publication.

**Validation and limitations:** The actual version helper bodies passed a host C++17 harness using the vendored version macros, a properties-API stub, and a formatting stand-in. Supported/unsupported versions, packed values 0 through 2500, and query failure were covered. This does not validate a real nvImageCodec binary ABI, dynamic-library load, full cuslide2/Catch2 build, or GPU decoding.

<details>

### Scope and method

Ran `gh pr checkout 1148`, read `gh pr view 1148`, then reviewed `gh pr diff 1148 --color never`. Reviewed all six changed files, including the deleted dependency module, and traced the runtime query, dynamic symbol wrapper, constructor failure path, version macros, dependency pins, and plugin build flow.

No corrective implementation diff is proposed because this review did not establish a concrete defect. The excerpts below document the reviewed behavior; they are not independently applicable patches.

### Runtime version check

### Relevant additions

`cpp/plugins/cucim.kit.cuslide2/src/cuslide/nvimgcodec/nvimgcodec_tiff_parser.cpp:79-80,98-101`:

```diff
--- a/cpp/plugins/cucim.kit.cuslide2/src/cuslide/nvimgcodec/nvimgcodec_tiff_parser.cpp
+++ b/cpp/plugins/cucim.kit.cuslide2/src/cuslide/nvimgcodec/nvimgcodec_tiff_parser.cpp
@@ -71,0 +79,2 @@
+static constexpr uint32_t kMinNvimgcodecVersion = MAKE_SEMANTIC_VERSION(0, 9, 0);
+static constexpr uint32_t kFirstUnsupportedNvimgcodecVersion = MAKE_SEMANTIC_VERSION(0, 10, 0);
@@ -71,0 +98,4 @@
+bool is_supported_nvimgcodec_version(uint32_t packed_version)
+{
+    return packed_version >= kMinNvimgcodecVersion && packed_version < kFirstUnsupportedNvimgcodecVersion;
+}
```

The runtime query initializes the API structure before calling the dynamically resolved function:

```diff
--- a/cpp/plugins/cucim.kit.cuslide2/src/cuslide/nvimgcodec/nvimgcodec_tiff_parser.cpp
+++ b/cpp/plugins/cucim.kit.cuslide2/src/cuslide/nvimgcodec/nvimgcodec_tiff_parser.cpp
@@ -71,0 +122,6 @@
+    nvimgcodecProperties_t properties{};
+    properties.struct_type = NVIMGCODEC_STRUCTURE_TYPE_PROPERTIES;
+    properties.struct_size = sizeof(nvimgcodecProperties_t);
+    properties.struct_next = nullptr;
+
+    const nvimgcodecStatus_t status = nvimgcodecGetProperties(&properties);
```

### Assessment

- The checked-in version macro is `major * 1000 + minor * 100 + patch`; the intended supported packed interval is `[900, 1000)`.
- The runtime policy matches the current `>=0.9.0,<0.10.0` dependency declarations in `dependencies.yaml` and the vendored 0.9.0 header.
- The 0.10.0/1.0.0 encoding collision is real, but both values are rejected. The explicitly documented diagnostic ambiguity is not a range-check bug for this supported interval.
- A failed properties query fails closed and reports the API status plus the supported range.
- The manager performs the query before instance/decoder creation (`nvimgcodec_tiff_parser.cpp:407-426`). A version failure leaves `initialized_` false and records the detailed message.
- The generated wrapper has a missing-symbol fallback returning `NVIMGCODEC_STATUS_IMPLEMENTATION_UNSUPPORTED`, so the new call does not unconditionally invoke a null function pointer (`nvimgcodec_stubs_generated.cc:10-19`).
- The message is not lost solely because diagnostic printing is behind `DEBUG`: `TiffFileParser` throws the manager status (`nvimgcodec_tiff_parser.cpp:578-584`), and the TIFF constructor rethrows it with the file context (`tiff.cpp:331-351`).

### CMake consolidation

The new conda-build branch uses the host/install prefix rather than the active build environment:

```diff
--- a/cpp/plugins/cucim.kit.cuslide2/cmake/deps/nvimgcodec.cmake
+++ b/cpp/plugins/cucim.kit.cuslide2/cmake/deps/nvimgcodec.cmake
@@ -85,1 +98,6 @@
-        if (DEFINED ENV{CONDA_PREFIX})
+        if (DEFINED ENV{CONDA_BUILD})
+            # Inside conda-build, CONDA_PREFIX points at the build environment
+            # rather than the install prefix, so use PREFIX instead.
+            _cucim_find_nvimgcodec_library(NVIMGCODEC_LIB_PATH "$ENV{PREFIX}/lib")
+            set(NVIMGCODEC_INCLUDE_PATH "$ENV{PREFIX}/include/")
+        elseif (DEFINED ENV{CONDA_PREFIX})
```

Checked the deletion against the repository's current build flow:

- The only `superbuild_depend(nvimgcodec)` call is in the cuslide2 project.
- cuslide2 configures its additional dependency directory before loading the helper, and the helper searches that directory first.
- The top-level project does not add the plugin as a subdirectory. The `run` script configures cuslide2 as a separate CMake project.
- The old top-level nvimgcodec dependency file is therefore not used in that flow. Its conda-build prefix handling was moved into the live module before removal.
- The newly added package-discovery upper-bound warning matches the existing header-discovery warning policy. The runtime check intentionally enforces a stricter rejection for unsupported loaded versions.

No regression was identified in these changes.

### Validation performed

Compiled the actual version-check helper bodies extracted from the reviewed Git object into a C++17 host harness with:

- The repository's real vendored `nvimgcodec_version.h` macros.
- A stub properties API that verifies the structure type, size, and null extension pointer, then returns a selected version or failure status.
- A minimal formatting stand-in; this does not validate the real fmt library or nvImageCodec binary ABI.

The harness passed:

1. Supported/unsupported version-query cases for packed values 700, 800, 899, 900, 901, 999, 1000, and 2304.
2. A complete packed-value sweep from 0 through 2500, accepting exactly `[900, 1000)`.
3. Acceptance of the vendored `NVIMGCODEC_VER`.
4. The documented 0.10.0/1.0.0 packed collision.
5. A non-success properties response, checking that the status and supported range appear in the error.
6. Required range and loader-path guidance in unsupported-version diagnostics.

The harness compiled with `clang++ -std=c++17 -Wall -Wextra -Werror` and ran successfully. It does not replace a full cuslide2 build or live-library test.

### Remaining integration coverage

- Run the built `cuslide2_tests` executable with the Catch2 filter `[test_nvimgcodec_version.cpp]`.
- Exercise the dynamic loader in separate processes against a supported library, an older library, an out-of-range newer library, and a missing/unavailable properties symbol. Separate processes matter because the manager and symbol pointers are cached.
- Verify rejection occurs before instance creation and that the public `CuImage` error includes the observed version/status and supported range.
- Test both dynamic-loading and direct-link build configurations on supported Linux/CUDA installations.
- Verify conda-build discovery with distinct `PREFIX` and `CONDA_PREFIX` values.
- Add a check comparing dependency declarations if automatic pin-drift detection is desired: the current hard-coded C++ cases establish today's range, but do not parse a future edit to `dependencies.yaml`.

No source files were modified, and no real GPU decoding, installed-library loading, or full Catch2 suite was executed on this macOS machine.

</details>

cpp/plugins/cucim.kit.cuslide2/tests/test_nvimgcodec_version.cpp
```diff
@@ -0,0 +1,77 @@
...
+
+// cuCIM checks the nvImageCodec version at run time because the build cannot:
+// the library is dlopen'd by default and the SONAMEs searched end with
+// libnvimgcodec.so.0 and libnvimgcodec.so, which resolve to whichever major-0
+// build is installed. These tests pin the range that check enforces.
+//
+// The range must stay in step with the pin in dependencies.yaml and with the
+// one cmake/deps/nvimgcodec.cmake enforces on headers. Nothing derives one from
+// the other, so this is what catches them drifting apart.
+TEST_CASE("supported nvImageCodec range matches the packaging pin", "[test_nvimgcodec_version.cpp]")
```
smatysik-nv, line 26: Non-blocking, for future consideration: this test pins the same boundaries as the C++ implementation, but it does not inspect `dependencies.yaml`, so it cannot detect packaging-pin drift by itself. For example, packaging could move to `>=0.10.0,<0.11.0` while this helper and test remain unchanged and continue passing. For this PR, I suggest renaming the test to describe the runtime policy (for example, “supported nvImageCodec runtime range is 0.9.x”) and avoiding the claim that it catches YAML drift. A separate CI consistency check or generated source of truth could enforce that later.

cpp/plugins/cucim.kit.cuslide2/src/cuslide/nvimgcodec/nvimgcodec_tiff_parser.cpp
```diff
@@ -69,6 +69,83 @@ namespace cuslide2::nvimgcodec
...
+
+// Same range as packed semantic-version integers, so they can be compared
+// directly against nvimgcodecProperties_t::version.
+static constexpr uint32_t kMinNvimgcodecVersion = MAKE_SEMANTIC_VERSION(0, 9, 0);
+static constexpr uint32_t kFirstUnsupportedNvimgcodecVersion = MAKE_SEMANTIC_VERSION(0, 10, 0);
+
+/**
+ * @brief Render a packed nvImageCodec version for a diagnostic message.
+ *
+ * The encoding gives the minor component a single digit
```
smatysik-nv, line 85: This ambiguity is an nvImageCodec bug rather than something cuCIM can resolve from the packed value. It should be fixed in nvImageCodec 0.10.0 by replacing `MAKE_SEMANTIC_VERSION` with `NVIMGCODEC_MAKE_SEMANTIC_VERSION` and changing the encoding to `major * 10000 + minor * 100 + patch`. This preserves `0.9.0 == 900` and `0.10.0 == 1000`, while making `1.0.0 == 10000`.

Until cuCIM adopts the 0.10 header, packed value `1000` remains ambiguous to its current formatter. Could this diagnostic therefore avoid claiming the loaded version is exactly `1.0.0`, and report the raw/ambiguous value instead? The range rejection itself is correct.

cdinea: thank you for the feedback @smatysik-nv - I think I addressed this in this commit  7c219ba: the diagnostic now reports the raw packed value and notes the ambiguity instead of identifying it as exactly 1.0.0. The range check remains unchanged, and I added regression coverage for ambiguous values, including 1000.

cpp/plugins/cucim.kit.cuslide2/tests/test_nvimgcodec_version.cpp
cdinea: thank you for the review @smatysik-nv - yes, the test verifies the runtime policy but cannot detect packaging-pin drift. Addressed in the previous 7c219ba commit : renamed it to “supported nvImageCodec runtime range is 0.9.x” and updated the comments to  ctlarify that checking consistency with the packaging pin requires a separate check

cdinea:
thank you for the review @gigony amd @smatysik-nv - I believe I addressed all the feedback.Please take a look and let me know if there are any other concerns

gigony (approved):
Thank you @cdinea for this change! It looks good to me.

PR #1149, opened 2026-09-07
Add ImageMetadata::store_string() and document why _GLIBCXX_USE_CXX11_ABI=0

cdinea:
## Summary

Addresses review feedback on the OME channel-name change, which asked two things:

> What is the reason for using `_GLIBCXX_USE_CXX11_ABI=0`?
>
> Before, I missed the fact that you are using pmr. If possible using string with pmr would be less tricky to read. But this can be done in a future PR

### Answering the question

`_GLIBCXX_USE_CXX11_ABI=0` dates to the initial commit and was set for **manylinux2014 wheels**: that platform tag requires a gcc4-compatible libstdc++ ABI (gcc configured with `--with-default-libstdcxx-abi=gcc4-compatible`), which forces the old ABI on CentOS 7. The reason was recorded in exactly one place — a comment inside `image_format.h` — while all six CMake files that actually set the flag carried a misleading `# TODO: create two library, one with CXX11 ABI and one without it.`

### Why the two asks are coupled

`std::pmr::string` is only defined for the new libstdc++ ABI, so this flag removes it. Confirmed directly:

```
$ g++ -std=c++17 -D_GLIBCXX_USE_CXX11_ABI=1 ...   # OK
$ g++ -std=c++17 -D_GLIBCXX_USE_CXX11_ABI=0 ...
error: 'string' is not a member of 'std::pmr'; did you mean 'std::string'?
```

That is also why `ImageMetadata` keeps two sets of member declarations under `#if _GLIBCXX_USE_CXX11_ABI`, and why the suggested form was already sitting commented out in `cuimage.cpp`:

```cpp
// std::pmr::string coord_sys_str{ image_metadata_->coord_sys ? image_metadata_->coord_sys : "", &resource };
```

### Why not just flip the flag

`cpp/CMakeLists.txt` exports the definition as an `INTERFACE` compile definition, so it lands in the installed `cucim-targets.cmake` and is part of the package's contract — flipping it is an ABI break for anything already linked against `libcucim`, not a local cleanup. Worth noting the original justification does look obsolete (the repo no longer references manylinux anywhere and `requires-python` is `>=3.11`), but that deserves its own deliberate change. This PR documents the situation instead of acting on it.

## Changes

- **`ImageMetadata::store_string()`** — copies a string into the metadata's own memory resource and returns a `std::string_view` of the copy. This is what the call sites actually wanted, and it is what `std::pmr::string` would have been used for. It works under either ABI setting, so it needs no revisiting if the flag is ever flipped.
- **Three call sites simplified** — cuslide2 channel names, plus the spacing-unit and `coord_sys` strings in `CuImage::read_region()`. The stale commented-out `std::pmr::string` attempts are removed.

  ```cpp
  // before
  char* buf = static_cast<char*>(resource.allocate(name.size() + 1, alignof(char)));
  std::memcpy(buf, name.c_str(), name.size() + 1);
  channel_names.emplace_back(buf, name.size());

  // after
  channel_names.push_back(out_metadata.store_string(name));
  ```

- **Comments at all six flag sites** replaced with the real reason, the concrete cost, and what changing it would involve, consolidated into one canonical explanation in the top-level `CMakeLists.txt`.
- **Tests** in `cpp/tests/test_metadata.cpp`.

One subtlety worth a reviewer's eye: `store_string()` writes the terminator itself rather than copying `size() + 1` bytes, because the input may be a view into a longer buffer whose following byte is not `'\0'` (e.g. `std::string_view{"DAPI/FITC", 4}`). The old channel-name code copied `size() + 1` and was only safe because its source was always a `std::string`.

## Testing

- The new `store_string()` cases need no GPU or slide files, unlike the rest of `test_metadata.cpp`.
- Logic verified standalone under `-fsanitize=address,undefined` with `_GLIBCXX_USE_CXX11_ABI=0`; all assertions pass clean, including the truncated-view and empty-string cases.
- `image_format.cpp`, `cuimage.cpp`, `cuslide.cpp`, and `test_metadata.cpp` all pass full semantic analysis (`-fsyntax-only` with each file's real compile-database flags).
- No behavior change intended: every touched site produced a null-terminated copy in the metadata resource before and does so now.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -91,7 +91,30 @@ option(CUCIM_SUPPORT_NVTX "Support NVTX" ON)
 # Setup CXX11 ABI
 # : Adds CXX11 ABI definition to the compiler command line for targets in the current directory,
 #   whether added before or after this command is invoked, and for the ones in sub-directories added after.
-add_definitions(-D_GLIBCXX_USE_CXX11_ABI=0) # TODO: create two library, one with CXX11 ABI and one without it.
+#
+# Why 0 (this is the canonical explanation; the other CMakeLists.txt files that
+# set the same definition point here):
+#
+#   This dates to building manylinux2014 wheels. That platform tag requires a
+#   gcc4-compatible libstdc++ ABI (gcc configured with
+#   '--with-default-libstdcxx-abi=gcc4-compatible', see
+#   https://gcc.gnu.org/onlinedocs/libstdc++/manual/configure.html), which forces
+#   _GLIBCXX_USE_CXX11_ABI=0 on CentOS 7.
+#
+# What it costs: parts of the standard library that are only defined for the new
+# ABI are unavailable, most visibly std::pmr::string. ImageMetadata therefore
+# keeps two sets of member declarations and stores strings as std::string_view
+# into its own memory resource; see cpp/include/cucim/io/format/image_format.h
+# and ImageMetadata::store_string().
+#
+# Changing it is not a local edit: cpp/CMakeLists.txt exports this definition as
+# an INTERFACE compile definition, so it is part of the installed package's
+# contract and flipping it breaks ABI for anything already linked against
+# libcucim. The manylinux2014 constraint itself no longer appears to apply (the
+# repository no longer references manylinux anywhere and requires-python is
+# >=3.11), so this is worth revisiting deliberately as its own change rather
+# than in passing.
+add_definitions(-D_GLIBCXX_USE_CXX11_ABI=0)
 
 ################################################################################
 # Define basic dependencies
--- a/cpp/tests/test_metadata.cpp
+++ b/cpp/tests/test_metadata.cpp
@@ -122,6 +122,56 @@ TEST_CASE("Verify metadata", "[test_metadata.cpp]")
     // REQUIRE(1 == 1);
 }
 
+TEST_CASE("ImageMetadata::store_string copies into metadata memory", "[test_metadata.cpp]")
+{
+    cucim::io::format::ImageMetadata metadata{};
+
+    SECTION("the copy is independent of the source and null-terminated")
+    {
+        std::string source{ "Cy5" };
+        std::string_view stored = metadata.store_string(source);
+
+        REQUIRE(stored == "Cy5");
+        REQUIRE(stored.data() != source.data());
+        // Consumers reach these through ImageMetadataDesc as plain `char*`.
+        REQUIRE(stored.data()[stored.size()] == '\0');
+
+        source[0] = 'X';
+        REQUIRE(stored == "Cy5");
+    }
+
+    SECTION("a view into a longer buffer is terminated at its own end")
+    {
+        // The reason store_string() writes the terminator itself instead of
+        // copying size() + 1 bytes from the source.
+        std::string_view middle_of{ "DAPI/FITC", 4 };
+        std::string_view stored = metadata.store_string(middle_of);
+
+        REQUIRE(stored == "DAPI");
+        REQUIRE(stored.size() == 4);
+        REQUIRE(stored.data()[4] == '\0');
+    }
+
+    SECTION("an empty value is still a valid null-terminated string")
+    {
+        std::string_view stored = metadata.store_string(std::string_view{});
+
+        REQUIRE(stored.empty());
+        REQUIRE(stored.data() != nullptr);
+        REQUIRE(stored.data()[0] == '\0');
+    }
+
+    SECTION("separate calls do not alias")
+    {
+        std::string_view first = metadata.store_string("R");
+        std::string_view second = metadata.store_string("G");
+
+        REQUIRE(first == "R");
+        REQUIRE(second == "G");
+        REQUIRE(first.data() != second.data());
+    }
+}
+
 TEST_CASE("Load test", "[test_metadata.cpp]")
 {
     cucim::CuImage img{ g_config.get_input_path("private/philips_tiff_000.tif") };
--- a/cpp/CMakeLists.txt
+++ b/cpp/CMakeLists.txt
@@ -136,7 +136,7 @@ target_compile_definitions(${CUCIM_PACKAGE_NAME}
         CUCIM_STATIC_GDS=$<BOOL:${CUCIM_STATIC_GDS}>
         CUCIM_SUPPORT_CUDA=$<BOOL:${CUCIM_SUPPORT_CUDA}>
         CUCIM_SUPPORT_NVTX=$<BOOL:${CUCIM_SUPPORT_NVTX}>
-        _GLIBCXX_USE_CXX11_ABI=0  # TODO: create two library, one with CXX11 ABI and one without it.
+        _GLIBCXX_USE_CXX11_ABI=0  # See the top-level CMakeLists.txt for why this is 0.
 )
 
 # Link libraries
--- a/cpp/include/cucim/io/format/image_format.h
+++ b/cpp/include/cucim/io/format/image_format.h
@@ -68,6 +68,22 @@ public:
     ImageMetadata();
     ~ImageMetadata();
     void* allocate(size_t size);
+
+    /**
+     * @brief Copy @p value into this metadata's memory and return a view of the copy.
+     *
+     * The setters below take std::string_view and the views are handed out through
+     * ImageMetadataDesc as plain `char*`, so the characters have to outlive whatever
+     * temporary produced them and have to be null-terminated. This copies once into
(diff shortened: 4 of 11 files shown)


cpp/tests/test_metadata.cpp
```diff
@@ -122,6 +122,56 @@ TEST_CASE("Verify metadata", "[test_metadata.cpp]")
...
+    SECTION("a view into a longer buffer is terminated at its own end")
+    {
+        // The reason store_string() writes the terminator itself instead of
+        // copying size() + 1 bytes from the source.
+        std::string_view middle_of{ "DAPI/FITC", 4 };
+        std::string_view stored = metadata.store_string(middle_of);
+
+        REQUIRE(stored == "DAPI");
+        REQUIRE(stored.size() == 4);
+        REQUIRE(stored.data()[4] == '\0');
```
gigony, line 152: ### [Optional] Use a source with no byte beyond the view

**AI-assisted review comment (optional suggestion 1/1).**

Optional: the truncated-view case checks the destination terminator, but its source literal has valid storage beyond the four-byte view. A four-byte array with no terminator would also catch a future `size() + 1` source read under ASan. The present implementation already passes this case; this is regression-test strengthening, not a correctness finding.

**Validation:** The actual store_string() body passed a host harness with a real monotonic_buffer_resource under AddressSanitizer and UndefinedBehaviorSanitizer, including empty, bounded/nonterminated, embedded-NUL, and arena-growth cases. The harness used macOS libc++, not Linux libstdc++'s dual ABI. Full cuCIM/Catch2, plugin, and slide integration tests were not run. The optional patch passed git apply --check without being applied.

<details>

**Location:** `cpp/tests/test_metadata.cpp:143-157`.

**Classification:** Non-blocking test-strengthening suggestion, not a bug in the current implementation.

The existing truncated-view test is useful, but its source is the string literal `"DAPI/FITC"`. Reading one extra byte would still be within that literal's storage. A four-byte nonterminated source, run under ASan, more directly guards against future reintroduction of a `size() + 1` source read.

### Suggested test patch

```diff
diff --git a/cpp/tests/test_metadata.cpp b/cpp/tests/test_metadata.cpp
--- a/cpp/tests/test_metadata.cpp
+++ b/cpp/tests/test_metadata.cpp
@@ -152,6 +152,16 @@
         REQUIRE(stored.data()[4] == '\0');
     }
 
+    SECTION("a bounded view does not require a source terminator")
+    {
+        const char source[] = { 'D', 'A', 'P', 'I' };
+        std::string_view stored = metadata.store_string(
+            std::string_view{ source, sizeof(source) });
+
+        REQUIRE(stored == "DAPI");
+        REQUIRE(stored.data()[stored.size()] == '\0');
+    }
+
     SECTION("an empty value is still a valid null-terminated string")
     {
         std::string_view stored = metadata.store_string(std::string_view{});
```

</details>

gigony (commented):

**AI-assisted review.** No actionable correctness defect was identified in this review. Sharing one explicitly optional, non-blocking test-strengthening suggestion inline.

Reviewed commit: `8831f7ef45c40acf9afc65567b6dc3b3f05caebd`. The PR description was read before the diff; the current head and inline locations were rechecked before publication.

**Validation and limitations:** The actual store_string() body passed a host harness with a real monotonic_buffer_resource under AddressSanitizer and UndefinedBehaviorSanitizer, including empty, bounded/nonterminated, embedded-NUL, and arena-growth cases. The harness used macOS libc++, not Linux libstdc++'s dual ABI. Full cuCIM/Catch2, plugin, and slide integration tests were not run. The optional patch passed git apply --check without being applied.

<details>

### Scope and method

Ran `gh pr checkout 1149`, read `gh pr view 1149`, then reviewed `gh pr diff 1149 --color never`. Reviewed all eleven changed files, including the three refactored call sites, metadata resource lifetime, descriptor setters, test registration, and all six CMake ABI-definition sites.

The description explicitly preserves the old ABI and defers any ABI migration. This review does not treat retaining `_GLIBCXX_USE_CXX11_ABI=0` as a defect.

Source locations refer to the reviewed head. PR diff excerpts are contextual, not standalone patches. The optional test patch is a proposal only.

### Implementation assessment

The new helper is in `cpp/src/io/format/image_format.cpp:27-38`:

```diff
--- a/cpp/src/io/format/image_format.cpp
+++ b/cpp/src/io/format/image_format.cpp
@@ -26,0 +27,12 @@
+std::string_view ImageMetadata::store_string(std::string_view value)
+{
+    char* buf = static_cast<char*>(res_.allocate(value.size() + 1, alignof(char)));
+    if (!value.empty())
+    {
+        std::memcpy(buf, value.data(), value.size());
+    }
+    // Terminate explicitly rather than copying value.size() + 1 bytes: `value` may
+    // be a view into a longer buffer, in which case the extra byte is not a '\0'.
+    buf[value.size()] = '\0';
+    return std::string_view{ buf, value.size() };
+}
```

The implementation correctly:

- Allocates from the same `ImageMetadata` object's monotonic resource that owns the resulting view's storage.
- Copies exactly the view length, without requiring the source's following byte to exist or be a terminator.
- Skips `memcpy` for an empty, null-data `string_view`, while still returning an allocated one-byte empty C string.
- Explicitly terminates the destination at `value.size()`.
- Preserves previous allocations when the monotonic resource obtains additional backing blocks.

The three refactored call sites preserve the relevant source lifetime and termination behavior:

| Call site | Reviewed location | Assessment |
| --- | --- | --- |
| cuslide2 channel names | `cpp/plugins/cucim.kit.cuslide2/src/cuslide/cuslide.cpp:284-289` | Copies each temporary `std::string` into `out_metadata` before the temporary vector is destroyed |
| Read-region spacing units | `cpp/src/cuimage.cpp:857-868` | Retains the null-pointer fallback and bounded source length |
| Read-region coordinate system | `cpp/src/cuimage.cpp:897-906` | Retains the null guard and bounded source length |

The existing descriptor setters may reuse already-terminated views. Views returned by this helper have a valid byte at `data()[size()]` and storage tied to the metadata object, so that reuse is safe at these changed sites.

Adding a nonvirtual member function does not alter the metadata descriptor layout or change the existing ABI macro. All six CMake sites still set the same value, 0.

### Remaining integration coverage

- Run the built `cucim_tests` executable with the specific Catch2 test name `ImageMetadata::store_string copies into metadata memory`; the broader `[test_metadata.cpp]` tag also includes slide-dependent tests.
- Validate the helper and affected translation units with the supported Linux/libstdc++ build flags.
- Exercise OME channel names and read-region spacing/coordinate-system metadata after temporary source strings have gone out of scope.
- Keep any future ABI migration separate and evaluate its packaging/binary compatibility explicitly; the Python minimum version alone is not sufficient evidence that an ABI change is safe.

No implementation files were changed during this review.

</details>

cpp/tests/test_metadata.cpp
cdinea: thank you for the review @gigony - I addressed the review in this commit https://github.com/rapidsai/cucim/pull/1149/commits/3b2c0d9601a4a83273731feec8595b28582bb04c

gigony (approved):
Thank you @cdinea for handling this.
It looks good to me!

CMakeLists.txt
```diff
@@ -91,7 +91,30 @@ option(CUCIM_SUPPORT_NVTX "Support NVTX" ON)
...
+# and ImageMetadata::store_string().
+#
+# Changing it is not a local edit: cpp/CMakeLists.txt exports this definition as
+# an INTERFACE compile definition, so it is part of the installed package's
+# contract and flipping it breaks ABI for anything already linked against
+# libcucim. The manylinux2014 constraint itself no longer appears to apply (the
+# repository no longer references manylinux anywhere and requires-python is
+# >=3.11), so this is worth revisiting deliberately as its own change rather
+# than in passing.
+add_definitions(-D_GLIBCXX_USE_CXX11_ABI=0)
```
jakirkham, line 115: Agree we should revisit this. Think it is reasonable to switch to the new C++11 string ABI. Have raised an issue to track this: https://github.com/rapidsai/cucim/issues/1163

Though also agree it should be in a dedicated PR that makes this change. Let's revisit in 26.12

jakirkham (approved):
Thanks Cristiana! 🙏

As a clarifying comment/change, think this is ok.

Though agree we should revisit this setting and switch to the newer C++11 string ABI. Let's discuss in 26.12

For now 26.10, this change looks good

jakirkham:
Looks like we are missing style fixes. Once those are pushed (and CI passes). This should be good to go

CMakeLists.txt
cdinea: thank you for the guidance @jakirkham

